PR #69, opened 2014-06-11
Open url by Ctrl + left click

kulti:


dglent:
This function 'ctrl + clic' to open a url, should be notified / mentioned, or added in the context menu on left click on the url. 
I was looking to report an issue and i discovered that i could open the url with ctrl + L/R clic here...

worblehat:
I definitely agree! I only discovered this feature by coincidentally reading this pull request, too.

PR #105, opened 2015-01-12
Use GNUInstallDirs in CMakeLists.txt to stop hardcoding paths

paulolieuthier:
Only tested with Qt5 on Linux.


paulolieuthier:
@jleclanche fixed now, translations are working.

PR #112, opened 2015-02-20
Fixed #93: Add 'Toggle Menu' and 'Preferences...' to context menu.

### LINKED ISSUE #93: Some more options should be available on right click if menu is hidden
Like "Preferences" and "Show menu", for example.

worblehat:


paulolieuthier:
Nice.

PR #118, opened 2015-02-20
Add fullscreen mode.

worblehat:
I really miss a fullscreen mode in QTerminal. When working in the terminal I sometimes like to disable all window decorations, menu bars, tab bars, etc. and make it fullscreen so that the black terminal window occupies the whole screen.

On the otherhand I don't really know, if there is a use case for the "Hide Window Borders" option. So is it maybe needless when we have a fullscreen mode?
What do you think?

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -60,6 +60,7 @@
 
 #define HIDE_WINDOW_BORDERS "Hide Window Borders"
 #define SHOW_TAB_BAR "Show Tab Bar"
+#define FULLSCREEN "Fullscreen"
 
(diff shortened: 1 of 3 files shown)


worblehat:
Sorry...I meant "Hide Window Borders".

jleclanche:
I personally think if this is there it should replace the hide window borders setting. thoughts @pvanek ?

pvanek:
let me check this first, please. I don't use stuff like this so I have no clue what "window borders" are in this case ;)

PR #121, opened 2015-02-23
Appdata file for QTerminal (resolves #60).

### LINKED ISSUE #60: Please create an AppData file for QTerminal
Please consider writing and installing an AppData file with the application description and some screenshots, else QTerminal looks really bad in the GNOME and KDE Software Centers. We'd love to showcase more applications, but without the extra data file we can't. See http://people.freedesktop.org/~hughsient/appdata/ for details; thanks!

Richard

worblehat:
Proposal for an appdata file. Does the description fit?

By now there are several applications on my system that already ship appdata. So maybe QTerminal should have it, too. Unfortunately I don't really know how to test it. There is a validator on http://people.freedesktop.org/~hughsient/appdata/ but it seems to be offline.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,6 +195,10 @@ install(FILES
     DESTINATION "${CMAKE_INSTALL_DATADIR}/applications"
 )
 
+install(FILES
+    qterminal.appdata.xml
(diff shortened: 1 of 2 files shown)


worblehat:
Forgot to mention: Did you take a look at the screenshots referenced in the AppData-file? Are they ok?

PR #122, opened 2015-02-23
Changed layout of font-setting to better fit to the other settings in th...

worblehat:
...e widget.

**Before**:

settings_before

**After**:

settings_after

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>688</width>
+    <width>713</width>
     <height>487</height>
    </rect>
   </property>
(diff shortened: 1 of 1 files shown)


jleclanche:
Can you swap the position of the change button and the display?

worblehat:
Done. That's indeed better.

settings_after_2

jleclanche:
Thanks, gtm

jleclanche:
Come to think of it, why not use a QFontComboBox?

pvanek:
my 2c: I'd say - keep using group box. It really helps with logic widget organization in this case.

QFontComboBox was used originally and it's really slow on initialisation. The prefs dialog appeared after some seconds - not immediately - using QFontComboBox.

jleclanche:
OK then

Issue #133, opened 2015-03-08
Setting 'Terminal Transparency' does not work.


worblehat:
'Application Transparency' makes the whole application transparent. But 'Terminal Transparency' has no effect. I tested this with a KDE-desktop.

Can anyone confirm this?

AcarBurak:
None is persistent in Fluxbox+Compton. Both transparencies work initially set in Preferences, but they revert back to 0 transparency when qterminal is relaunched. I need to re-enable them in Preferences every time.

worblehat:
I can confirm this. You used the latest version from the repository, don't you?
It is related to the latest renaming of the transparency settings. I will fix this tomorrow. Thanks!

Back to the original question: Making the terminal transparent without making the whole application transparent works for you?

AcarBurak:
Sorry to omit this. No, same as yours, as you've said in the first comment.

I use Arch AUR packages (qtermwidget-git and qterminal-git by jleclanche). 

Thank you.

agaida:
Confirmed, latest git with Qt 5.3.2/5.4.1 on debian sid

jleclanche:
dup of #138

agaida:
hmm - build the latest version from git - transparency still not work so i reopen the bug

PR #136, opened 2015-03-14
Make sure no settings are overwritten in migrateSettings().

worblehat:
Fixes a bug mentioned in #133. But does not fix the original bug of this issue!


worblehat:
Fixes #138

Issue #142, opened 2015-03-29
Single middle-click for closing tabs


fturco:
In Firefox you can close tabs by clicking them with the middle button of the mouse. You can also hide the X icon on each tab. It would be cool if with QTerminal you could do the same.

jleclanche:
I'm vetoing this out as bad UX. On firefox you can recover tabs, but on qterminal you can't and you might accidentally close important processes etc.

fturco:
Actually I accidentally closed important processes a number of times by left-clicking on the X icons, especially when tab names are very short. Using two different mouse buttons (left/middle) can help preventing this. I never click the middle mouse button when the left one is intended, or vice versa. But perhaps that's just me.

dimgel:
I'd very much like to see this feature too: (1) close tab by middle button; (2) ability to turn off "X" icons. Same troubles as flurco mentioned: left-clicking tab may accidentially hit "X". Middle-clicks work fine not only in browsers, but also in tint2 taskbar, for example.

yan12125:
> ability to turn off "X" icons

That sounds another useful feature. Could you open a new issue for that?

dimgel:
Sure: #300

pmattern:
The arguments in favour of closing this issue aren't remotely plausible.
While it's true that the new feature can cause incidental closing of tabs the same applies to the existing buttons ("X") where it may even be more likely as rightly pointed out above. Yet nobody called for removing those buttons so far.

Closing a feature request due to the subjective opinion of a single team member is rarely good practice.

All in all reopening to allow for some more discussion. IMO implementing this as an option would very well be an advantage.

yan12125:
Already implemented in https://github.com/lxqt/qterminal/pull/397

jlacros:
I can't even tell you how many tabs I have closed by accidentally middle-clicking the tab when I meant to middle-click in the window and paste text... it's absolutely infuriating. Did you implement a way to disable this?

yan12125:
> Did you implement a way to disable this?

Currently no. Please create a new issue about that, thanks!

tsujan:
> Currently no.

Terminal Settings → Behavior → Close tab on middle-click

yan12125:
Sorry forgot about it. That option is available since qterminal 0.16 (#677)

PR #144, opened 2015-04-04
Close tabs by middle click.

worblehat:
Implements the feature requested in #142.


pvanek:
this is very dangerous feature. I don't say it's useless, but maybe it should be configured (default off). @jleclanche your thoughts?

jleclanche:
I agree with you @pvanek, this deserves an option and off by default.

worblehat:
Is there a way to figure if there is a running foreground-process in a tab? We could add an additional confirmation dialog if there is an active process.

jleclanche:
Rejecting.

Issue #131, opened 2015-03-07
[req] stop QSettings storing window geometry in human not readable ByteArrays


LeonardKoenig:
Hello,

I've just read up about it. It theoretically is possible to, instead of using toByteArray() to save window geometry, one can "store both pos() and size() and to restore the geometry using QWidget::resize() and move() before calling show(), as demonstrated in the Application example." (http://doc.qt.io/qt-5/restoring-geometry.html)
The application example is this one: http://doc.qt.io/qt-5/qtwidgets-mainwindows-application-example.html

I prefer numbers over bytearrays when it comes to actually saving the data into a settings file :-)

Is that possible to do?

pvanek:
yes, it is possible to do.

But for me it has almost minimal priority as I don't see real profit from this change. Please do not take it too personally, but you should provide a patch or the change won't happen ;)

Just because I see it as a change for a change only.

LeonardKoenig:
Sure, I understand that it has minimal priority ;-)

Maybe I can further look into this later this year, but I don't have more experience with Qt than translations and what I just researched ;-)

pvanek:
well, the change is very easy. Just read QSettings class docs. But it would require settings migration in Properties::migrate_settings()

LeonardKoenig:
Sure 'nogh, but I'm just full in my final exams, so I don't have too much of time :/

worblehat:
+1

The font setting is also stored as binary array. I think it would definitely be an improvement to store it in human readable form, too. Maybe I'll take a look at it sometime.

jleclanche:
PR wanted.

skebanga:
+1

jleclanche:
@pvanek I don't think this needs settings migration code, it's very minor.

PR #147, opened 2015-04-12
split geometry into position - size, add an option to save each

pierrenopicnic:
When qterminal is closed, geometry is saved.
New instance of qterminal will restore geometry, and overlap existing one. so I have split geometry into position and size, and I add two options to save and restore both.

the purpuse is, when save position is unchecked, to let the WM places new windows

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -60,10 +60,11 @@
 
 #define HIDE_WINDOW_BORDERS "Hide Window Borders"
 #define SHOW_TAB_BAR "Show Tab Bar"
+#define FULLSCREEN "Fullscreen"
 
 /* Some defaults for QTerminal application */
 
-#define DEFAULT_WIDTH                  800
+#define DEFAULT_WIDTH                  1000
 #define DEFAULT_HEIGHT                 600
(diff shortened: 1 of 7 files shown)


worblehat:
I think it's a good idea to make saving/restoring of the window geometry configurable.

Storing the geometry in separate properties (size and position) instead of a single byte-array also resolves #131. The settings need to be migrated in `Properties::migrateSettings()` though. At least the old geometry setting should be removed from the configuration file.

Although I like the separation into different properties, I would suggest to have only one option for the user in the settings dialog (as in you first commit 8075d69). I don't see a real benefit in making size and position configurable separately and I think we should try to minimize the options in the settings dialog.

pierrenopicnic:
sorry for the fullscreen patch. I'll remove it.

I will have a look on migrateSettings too.

but I disagree with the third point. In my case, I use to open 3 or 4 terminals, and I like them to have the same size and be cleverly placed by the WM. so I need size to be remembered but not position!

If we link size and position in configuration, then I will have to add configurable default height and width, and it will be 3 options instead of 2...

PR #158, opened 2015-07-12
Fixes missing App Window icon

luis-pereira:
Qt Icon Engine doesn't serch in /usr/share/pixmaps. It should but it
doesn't. pixmaps is for legacy apps anyway. The hicolor theme is the right
place to put unthemed icons.

Closes qterminal/qterminal#143.


jleclanche:
gtm

Issue #163, opened 2015-07-24
appearance: separate transparency of terminal widget with full opaque main window


antis81:
The transparency values of the terminal widget and the main window can be set (both default to 0% == opaque).

Changing the value for the terminal's transparency has no effect at all -> :beetle: bug. Setting the main window's transparency works like expected, but I wish for a behaviour like in `lxterminal` (I think most terminal emulators), where just the terminal is transparent, while the app-window always is fully opaque.

This can be achieved similar to this example:
```c++
MainWindow::MainWindow(QWidget *parent)
    : QMainWindow(parent)
{
    // This makes the (widget area) background completely transparent, leaving the menu and tool bar untouched.
    this->setAttribute(Qt::WA_TranslucentBackground);
    this->setupUi(this);

    // the following flags might or might not be required (check them out):
    // terminal->setAttribute(Qt::WA_TranslucentBackground, false);
    // terminal->setAttribute(Qt::WA_OpaquePaintEvent);
    // terminal->setAutoFillBackground(true);
    terminal->setTerminalOpacity(.5);
}
```

Like expected, this only works with available compositing (I use `Compton` for X11).

Wrote a little test application for myself, playing with the values. To fake the terminal I used a simple QLabel with overriden `paintEvent`. Result looks like this:

![qterm-transparent-55](https://cloud.githubusercontent.com/assets/440517/8876269/ecf2b7a4-321e-11e5-98a7-6932b8c3a703.png)

... and fully tranparent:

![qterm-transparent-100](https://cloud.githubusercontent.com/assets/440517/8876281/f9086e58-321e-11e5-8086-f7f3627d445e.png)

pvanek:
the pull request has been merged. Can we close this one?

PR #169, opened 2015-08-15
make only terminal widget transparent, not the parent window

antis81:
Closes issue #163.

The attribute Qt::WA_translucentBackground applies only to top-level widgets (main windows) and makes every child widget transparent, unless AutoFillBackground is set or the child explicitly draws it's background.

This allows to control the terminal widget's transparency, while the window frame stays fully opaque. Unfortunately this is not valid vice versa. Meaning: If you change window transparency, it also affects the terminal transparency.

Anyway, this is far better than before.

Screenshot:

qterminal-semi-transparent

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -46,6 +46,8 @@ MainWindow::MainWindow(const QString& work_dir,
       m_dropLockButton(0),
       m_dropMode(dropMode)
 {
+    setAttribute(Qt::WA_TranslucentBackground);
+
     setupUi(this);
     Properties::Instance()->migrate_settings();
     Properties::Instance()->loadSettings();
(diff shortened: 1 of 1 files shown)


jleclanche:
This sounds good, but I don't use the transparency feature, so waiting on feedback.

antis81:
>This sounds good, but I don't use the transparency feature, so waiting on feedback.

It's already there - just doesn't work correct. This PR fixes it. :)

paulolieuthier:
This PR introduced a bug. The menu is now transparent for me even though I have transparency set to 0.

screen

antis81:
@paulolieuthier This obviously only affects some compositors or WM's. I use `openbox & compton` and it works all good. Didn't check without any compositor though. Which setup are you using?

Anyway, setting `mainMenu()->setAutoFillBackground(true);` should work. Without this, the background of the main window and any child widget (e.g. the menu) appears completely translucent (see `Qt::WA_TranslucentBackground`). This is independent of any opacity values and creates the effect, that only the window's frame is visible.

paulolieuthier:
I'm using KWin 5.4.2.

Issue #168, opened 2015-08-12
[!] Failed to "make install"! Please have a look, Thanks!


MuSamDu:
I'm trying to install your qterminal. I installed all the dependences, and successfully compiled it. But when i try to "make install" it, errors occured.
Outputs after "make install" begin to seem strange from here:
>
[ 54%] Building CXX object CMakeFiles/qterminal.dir/src/main.cpp.o
[ 57%] Building CXX object CMakeFiles/qterminal.dir/src/mainwindow.cpp.o
[ 59%] Building CXX object CMakeFiles/qterminal.dir/src/tabwidget.cpp.o
[ 61%] Building CXX object CMakeFiles/qterminal.dir/src/termwidget.cpp.o
/home/msd/Downloads/qterminal/src/termwidget.cpp:107:2: warning: #warning TODO/FIXME: disable the action when there is only one terminal [-Wcpp]
 #warning TODO/FIXME: disable the action when there is only one terminal
  ^
[ 64%] Building CXX object CMakeFiles/qterminal.dir/src/termwidgetholder.cpp.o
[ 66%] Building CXX object CMakeFiles/qterminal.dir/src/properties.cpp.o
[ 69%] Building CXX object CMakeFiles/qterminal.dir/src/propertiesdialog.cpp.o
[ 71%] Building CXX object CMakeFiles/qterminal.dir/src/bookmarkswidget.cpp.o
[ 73%] Building CXX object CMakeFiles/qterminal.dir/src/fontdialog.cpp.o
[ 76%] Building CXX object CMakeFiles/qterminal.dir/src/third-party/qxtglobalshortcut.cpp.o
[ 78%] Building CXX object CMakeFiles/qterminal.dir/src/third-party/qxtglobalshortcut_x11.cpp.o
/home/msd/Downloads/qterminal/src/third-party/qxtglobalshortcut_x11.cpp:37:45: fatal error: qpa/qplatformnativeinterface.h: No such file or directory
 #   include <qpa/qplatformnativeinterface.h>
                                             ^
compilation terminated.
CMakeFiles/qterminal.dir/build.make:407: recipe for target 'CMakeFiles/qterminal.dir/src/third-party/qxtglobalshortcut_x11.cpp.o' failed
make[2]: *** [CMakeFiles/qterminal.dir/src/third-party/qxtglobalshortcut_x11.cpp.o] Error 1
[...]

PR #170, opened 2015-08-17
solve private header dependency for X11 global shortcuts

antis81:
@pvanek Can you please review CMakeLists.txt and check on OSX? This change actually should be part of the unmaintained libQxt library.


antis81:
Fixes #168

antis81:
@pvanek **poke** please reply, so we can solve this. I'm not going to contribute to the unmaintained `libqxt`. Like I tried to say in other post, we need to get rid of those dependencies. Why? First, libQxt is unmaintained :arrow_right: becomes unstable over time. We should rather make this a part to Qt itself!

jleclanche:
It looks fine, I'm landing this.

antis81:
@jleclanche Thanks! Will reduce headaches to lot of developer's. :+1:

PR #171, opened 2015-08-21
Fix license issues within qterminal

agaida:


agaida:
@pvanek - are you ok with this changes? We need the fixed headers as we are eager to upload qterminal to debian.

jleclanche:
gtm

PR #172, opened 2015-08-22
Add chinese translation

losywee:
Add a Simplified Chinese translation for qterminal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,7 @@ set(QTERM_TS
     src/translations/qterminal_hu.ts
     src/translations/qterminal_ru.ts
     src/translations/qterminal_ja.ts
+    src/translations/qterminal_zh_CN.ts
 )
 
 if(USE_QT5)
--- a/src/translations/qterminal_zh_CN.ts
+++ b/src/translations/qterminal_zh_CN.ts
@@ -1,0 +1,658 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="ja">
+<context>
+    <name>BookmarksWidget</name>
+    <message>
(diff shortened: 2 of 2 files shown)


jleclanche:
I don't think that's how access keys are supposed to be done in cjk..

losywee:
Maybe you are correct, but it is not cjk to do. I am sorry that I can't understand you issue. I only want to translate it to Simplified Chinese translation.

pvanek:
maybe it makes sense. Alt+<char> is listed in ascii sequence as part of simplified chinese strings. Maybe it's standard in zh_CN, who knows ;)

u-ho-bros:
Simplified Chinese translation is not working .
Startup translation files [qterminal_zh_CN.qm] for is not loaded .
To fix the problem , you must modify the main.cpp.

--------------------------------------------------------------------------
[kuma@kuma-manjaro-vm ~]$ export LANG='zh_CH.UTF-8'
[kuma@kuma-manjaro-vm ~]$ qterminal 

(process:1418): Gtk-WARNING **: Locale not supported by C library.
        Using the fallback 'C' locale.
iconTheme "Compass"
TRANSLATIONS_DIR: Loading translation file "qterminal_zh.qm" from dir /usr/share/qterminal/translations
load success: false

losywee:
Thanks for your reply. Yes, I had known currently, maybe I am a string idiot, oha...

u-ho-bros:
Please look at the pull request # 175.
I like that can be displayed in Chinese, it attempted to fix.
I In the environment have menu can be displayed in Chinese.
manjaro_lxqt

losywee:
Yes, Thanks.

losywee:
To be honest, I translate it by Japanese translation. I am sorry for the careless pedestrian.

PR #194, opened 2015-12-22
Fix menubar transparency

### LINKED ISSUE #187: Menubar Transparency
Since you've set WA_TranslucentBackground for the whole window, please set the property "autoFillBackground" to true for the menubar (m_menuBar) in "qterminal.ui".

My reason is that, contrary to Fusion, some style engines don't see any reason to fill the menubar with the bg color (because they support window translucency) and with them, the menubar could be totally transparent.

f2404:
Fixes #187.

### DIFF
--- a/src/forms/qterminal.ui
+++ b/src/forms/qterminal.ui
@@ -63,6 +63,9 @@
      <height>26</height>
     </rect>
    </property>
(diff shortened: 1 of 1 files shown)


paulolieuthier:
GTM.

Issue #198, opened 2015-12-25
Make tab active when right-clicking on it


f2404:
Now, as the context (right-click) menu actions are applied to the tab that the mouse was pointed over, we need to make that tab active when the menu is shown.
Looks like we want this behavior to be optional.

pmattern:
Hm, are you sure about this one?

It can be very convenient to e. g. rename some tab while watching what's going on in another one. Also, as pointed out in #189, several other applications, among other Konsole, don't activate a tab upon opening its context menu either.

So IMO the behaviour introduced in 616c47b is just fine. If you should introduce ativating a tab by opening its context menu I'd really ask to make it an option and not something mandatory.

f2404:
I've recalled the for now PR, thanks.

Issue #189, opened 2015-12-08
tabs: context menu always referring to active tab


pmattern:
The context menu of tabs is always referring to the active tab, not to the one it was opened over. E. g. in this screenshot
![qterminal_tab-context-menu-reference](https://cloud.githubusercontent.com/assets/9860801/11670814/4ad5ba14-9e04-11e5-80cd-875cb6856e6f.png)
both actions of the context menu are applied to tab `Shell No. 1` besides the context menu was opened over tab `Shell No. 2`.

This behaviour is inconvenient, rather confusing and doesn't seem to be common practice either.
Konsole as well as other applications like PCManFM-Qt, Firefox, Otter Browser and QupZilla are applying context menus' actions to the tab the menu was opened over regardless of whether or not that tab happens to be the active one. Xfce terminal is making a tab the active one as soon as its context menu gets opened.
I think the former behaviour (Konsole, PCManFM-Qt and so on) is the most reasonable and should be implemented in QTerminal, too.

PR #199, opened 2015-12-28
Make the tab active when showing context menu

f2404:
Fix for #198


f2404:
One issue here: First time, setCurrentIndex() works fine. But if rename the tab, and then try renaming another one - then the tab focus switches back to the previous tab, and after that returns.
My Qt knowledge is not enough to figure out why this is happening. Will be glad if someone helps.

pmattern:
I'd really ask to have this feature optional at best, see my attempt to explain [in #189](https://github.com/lxde/qterminal/issues/198#issuecomment-167562932).

f2404:
Yeah, I was referring to the XFCE terminal behavior to make the tab active... But I think I can step back and recall this change, especially knowing that it's not working very well. Thanks!

PR #200, opened 2015-12-29
Delete MainWindow object on exit

f2404:


f2404:
Cleared up the destructor.

PR #203, opened 2016-01-02
Add Greek (el) translation

dglent:
Add Greek (el) translation qterminal_el.ts and translation of the desktop files.


dglent:
I will modify the commit because i didn't translate the latest git version

dglent:
ok, i pushed again (forced) and now the commit contains the latest git translations

PR #207, opened 2016-01-27
Make the rename current session shortcut work and allow to configure it

### LINKED ISSUE #188: tabs: shortcut to rename not configurable, default not functional
The shortcut to rename tabs cannot be configured, default setting `Alt+Shift+S` doesn't work.

Seen running QTermWidget lxde/qtermwidget@45ef247 and QTerminal lxde/qterminal@ee32e84 on Arch Linux i686 and x86_64, LXQt or KDE.
Shortcut `Alt+Shiftt+S` wasn't assigned globally as it was working as expected if assigned to e. g. `Next Tab` from Preferences - Shortcuts.

f2404:
Closes #188

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -58,6 +58,7 @@
 
 #define HIDE_WINDOW_BORDERS "Hide Window Borders"
 #define SHOW_TAB_BAR "Show Tab Bar"
+#define RENAME_SESSION "Rename Session"
 #define FULLSCREEN "Fullscreen"
 
 /* Some defaults for QTerminal application */
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -99,11 +99,6 @@ MainWindow::MainWindow(const QString& work_dir,
        the main window; otherwise, the initial prompt might
        get jumbled because of changes in internal geometry. */
     consoleTabulator->addNewTab(command);
-
-    // Add global rename Session shortcut
-    renameSession = new QAction(tr("Rename Session"), this);
-    renameSession->setShortcut(QKeySequence(tr(RENAME_SESSION_SHORTCUT)));
-    connect(renameSession, SIGNAL(triggered()), consoleTabulator, SLOT(renameSession()));
 }
 
 MainWindow::~MainWindow()
@@ -295,7 +290,15 @@ void MainWindow::setup_ActionsMenu_Actions()
     Properties::Instance()->actions[TOGGLE_MENU]->setShortcut(seq);
     connect(Properties::Instance()->actions[TOGGLE_MENU], SIGNAL(triggered()), this, SLOT(toggleMenu()));
     addAction(Properties::Instance()->actions[TOGGLE_MENU]);
-    // tis is correct - add action to main window - not to menu to keep toggle working
+    // this is correct - add action to main window - not to menu to keep toggle working
+
+    // Add global rename current session shortcut
+    Properties::Instance()->actions[RENAME_SESSION] = new QAction(tr("Rename session"), this);
+    seq = QKeySequence::fromString(settings.value(RENAME_SESSION, RENAME_SESSION_SHORTCUT).toString());
+    Properties::Instance()->actions[RENAME_SESSION]->setShortcut(seq);
+    connect(Properties::Instance()->actions[RENAME_SESSION], SIGNAL(triggered()), consoleTabulator, SLOT(renameCurrentSession()));
+    addAction(Properties::Instance()->actions[RENAME_SESSION]);
+    // this is correct - add action to main window - not to menu
 
     settings.endGroup();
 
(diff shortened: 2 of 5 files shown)


pmattern:
In f2404/qterminal@fcf8995d6e changing tab titles by shortcut `Alt+Shift+S` works as expected, so the commit seems GTM.

On a side note I was having a hard time cloning your fork as git kept poking me about credentials.
This happened sporadically when invoking git itself from the command line
```
git clone https://github.com/f2404/qterminal.git
```
but always when I tried to install the commit by a modified version of an Arch Linux AUR package ([qterminal-git](https://aur.archlinux.org/packages/qterminal-git/) here) which is the approach I normally always take to perform tests
```
  -> Cloning qterminal git repo...
Cloning into bare repository '/usr/local/peter/Quellen/qterminal'...
Username for 'https://github.com':
```
If you should happen to have an idea what's going on here I'd be glad to know.

f2404:
@pmattern Thanks for checking the PR!
I'm also using Arch Linux, and I'm experiencing no issues with either cloning the repo manually or building `qterminal-git` from my repo.
I wonder what could be happening to your system...

f2404:
BTW, speaking about Arch Linux packages: wouldn't it make sense if `qterminal-git` strictly required `qtermwidget-git`?

pmattern:
That credential thing seems pretty weird.
The fact that it's just me to see it suggests it's something wrong at my side, the fact that I cannot see it with any other of the bunch of repositories and packages I'm dealing with suggests there's something wrong with the repo.
But as the commit itself works as expected I really don't feel like spending time on trying to track this down, tbh.

Regarding the question whether VCS packages should depend on VCS packages as well I'm not that sure.
Generally speaking there's one huge drawback. Say software foo depends on software bar and VCS package `foo-git` can get compiled against release package `bar` as well as against VCS package `bar-git`. Then stating `bar` as dependency allows for using *both* packages `bar` and `bar-git` while stating `bar-git` allows for using this package *only* and will enforce the removal of an eventually installed package `bar`. This may impose restrictions on users who want to run other software which eventually needs bar, too, but can *not* deal with `bar-git`.
This probably isn't an issue at all with regards to QTerminal. But in package [xf86-video-qxl-git](https://aur.archlinux.org/packages/xf86-video-qxl-git/) I recently replaced dependency `spice-protocol-git` with `spice-protocol` as the former turned out to be inconvenient.
All in all I think VCS packages are primarily for testing. Hence users who compile them should probably know what they are doing anyway and in particular be aware of any potential drawbacks which result from stating release instead of VCS packages as dependency.
It's of course a different situation if a VCS packages really needs the VCS version of its dependency. But probably this isn't the situation you had in mind.

f2404:
For the credential thing - I guess the question should be addressed to the GitHub support team. Or, I can delete my repo and create it again and see what happens.

For the dependencies thing - I think it's fine for a VCS package to depend on either stable or VCS version of another package - unless that another package changes ABI or adds a functionality our package depends on in VCS.
In general, I'm thinking that while both packages are under development (like `qterminal` and `qtermwidget`), it's safer and more error-prone to make stable version depend on a stable version, and VCS version depend on a VCS version. You saying that anyone who compiles a VCS package should know what they are doing are also correct. But still, maybe it's just more straight-forward to take the VCS->VCS approach unless there are good reasons to depend on both versions.

f2404:
The latest commit makes the shortcut work and allows to configure it - and thus fully resolves #188

pmattern:
f2404/qterminal@2119cef4378 is indeed fixing both problems addressed in #188 right now.
Major UX improvement, IMO. So thanks a lot for implementing this!
You cold eventually add a line like `Closes #188` to the commit message which will have this issue closed automatically upon merging the commit.
But also without this GTM.

On a side note cloning your fork was working like a charm today...

f2404:
@pmattern Thanks! but unfortunately, the commit is in fact introducing a regression - when choosing "Rename session" from the context menu, the tab being renamed is the active tab, not the one the menu has been opened against (I've already fixed this issue).

Now I will have to determine if the action was called by a shortcut (and then rename the active tab), or from the menu.
Please don't merge it for now.

f2404:
The latest one fixes the regression.

pmattern:
Ups, f2404/qterminal@2119cef was indeed reverting 616c47b and thus re-introducing the problem discussed in #189, but f2404/qterminal@fdd9e9b seems to fix the problem again.

> Please don't merge it for now.

Except trivial stuff like fixed typos or maybe translations I'd normally never merge commits based upon a single vote only.

f2404:
Anyone?

f2404:
Ping @jleclanche @pmattern @pvanek

jleclanche:
This is for @pvanek to decide on

f2404:
Ping @jleclanche @pmattern @pvanek @agaida

jleclanche:
Landed in bd61ea83bfcf7ff10cae6b0e8b570fd82ad6333e

PR #208, opened 2016-01-28
Open a new tab instead of closing the app in dropdown mode

### LINKED ISSUE #154: qterminal -d quits after the last terminal has been closed
1. open qterminal in dropdown mode
2. in qterminal's bash type "exit" (or press ctrl-d)
=> qterminal process quits

Expected behaviour:
- There should always be at least one active terminal.
- If the last terminal has been exited it is replaced by a new one. (qterminal process can still be stopped by File->Exit)

f2404:
Fixes #154

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -757,7 +757,11 @@ void MainWindow::onCurrentTitleChanged(int index)
 void MainWindow::maybeClose()
 {
     if (m_dropMode)
+    {
+        int index = consoleTabulator->currentIndex();
         addNewTab();
+        consoleTabulator->removeTab(index);
+    }
     else
         close();
 }


jleclanche:
That... doesn't seem right at all.

f2404:
Why? I see it's implementing the request.

pvanek:
It seems strange to me. But I'm not dropdown user...

f2404:
@agaida Rebased.

agaida:
tested this and see to many glitches - the idea is good, but the code not (yet)
https://bugs.siduction.org/attachments/download/272/vokoscreen-2016-10-05_20-06-27.mkv

f2404:
I will try to check it tomorrow.

f2404:
@agaida Please check the latest commit.

agaida:
@f2404: Works for me

@palinek @pmattern @paulolieuthier - could you test too?

pmattern:
Closing a second (regular) window which had been opened from a QTerminal "window" in drop-mode before terminates the QTerminal instance altogether. Don't know whether this is expected behavior.

That aside f2404/qterminal@aa01554e seems to behave as asked for in #154. But I'm probably not the right person to judge that whole thing because I [still](https://github.com/lxde/qterminal/issues/154#issuecomment-177171689) think the behavior proposed here is just plain wrong.

pmattern:
Meanwhile I do think this new feature is a good thing as well.

@f2404
(Assuming you are not too annoyed by this PR celebrating its first birthday some days ago:)
What I wrote [above](https://github.com/lxde/qterminal/pull/208#issuecomment-261397026) (first paragraph) wasn't accurate. When another "regular" instance is launched by invoking `qterminal &` in a drop-down terminal and this additional instance is terminated, the instance in drop-down mode is not terminated as well. Rather, the key to make the drop-down instance visible seems to get disabled. Isn't this worth having another look at?
Any chance you can turn the new feature into an option? Could very well be enabled by default. As for the location within the settings I'd suggest an additional new topic at Behaviour (left pane) - Behaviour (section in right pane).
Sure enough the commits have to be rebased by now as well.

PR #210, opened 2016-02-28
Enable to set TERM by using new property setenvTERM

rago1975:
In current status, qterminal sets environment variable TERM to xterm. However, I sometimes want to use xterm-256color when I use Vim with color scheme (related issue #100).
My suggestion is addition a new property _setenvTERM_ which can be set by editing in configuration file (`~/.config/qterminal.org/qterminal.ini`).

When the commited program is executed , a line for property _setenvTERM_ is automatically added in `qterminal.ini`. 

If user want to change TERM, edit this line in the file and restart.
For example, in the case using xterm-256color, edit the line in `qterminal.ini` as following:
`setenvTERM=xterm-256color`

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -100,8 +100,6 @@ void parse_args(int argc, char* argv[], QString& workdir, QString & shell_comman
 
 int main(int argc, char *argv[])
 {
-    setenv("TERM", "xterm", 1); // TODO/FIXME: why?
-
     QApplication::setApplicationName("qterminal");
     QApplication::setApplicationVersion(STR_VERSION);
     QApplication::setOrganizationDomain("qterminal.org");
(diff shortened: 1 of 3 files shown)


rago1975:
I revised committed codes.

pvanek:
hmm, I don't see proper setenv handling. Should not go real setenv to main()?

also I believe this:

```
    //setenv "TERM". If setenvTERM is empty, set it to xterm.
    setenvTERM = m_settings->value("setenvTERM", QString()).toString();
    if (setenvTERM.isEmpty())
        setenv("TERM", "xterm", 1);
    else
        setenv("TERM", setenvTERM.toUtf8().data(), 1);   
```

can be done as:

```
    //setenv "TERM". If setenvTERM is empty, set it to xterm.
    setenvTERM = m_settings->value("setenvTERM", "xterm").toString();
    setenv("TERM", setenvTERM.toUtf8().data(), 1);
```

pvanek:
and also - there is no UI to set this value in preferences

Issue #212, opened 2016-04-02
qterminal segfaults when collapsing vertical subterminals


agaida:
@palinek @pvanek @paulolieuthier
Please see full bug at: http://bugs.siduction.org/issues/1853
 backtrace and a video how to reproduce are attached. Source is  31d21b7

palinek:
Me?
I'm neither using the qterminal nor did I ever change anything in its code base...

agaida:
sorry @palinek

f2404:
@agaida ?

agaida:
GTM

agaida:
Fixed with lxde/qterminal/67e2ba

PR #214, opened 2016-04-04
Fix segfault when collapsing subterminals

f2404:
Looks like it's just trying to access a deleted object... Fix for #212


luis-pereira:
GTM

PR #221, opened 2016-04-23
Update qterminal_it.ts

stefonarch:
Updating italian translation.


paulolieuthier:
You added a lot of `&amp;`s. Was that on purpose?

stefonarch:
I took over  all the keyboard accelerators from source text (actually a lot, for near everything...).
But actually I see now that none of them get's displayed in qterminal.
schermata03

jleclanche:
Closing, translations are now tracked separately.

PR #222, opened 2016-05-12
ts-files removal

agaida:
remove only, leave history untouched


agaida:
@paulolieuthier @palinek @luis-pereira - ok with the removal?

pvanek:
Why are TS files removed at all?

palinek:
The translations were moved into lxde/translations... Sure, the build system here (to pull translations) should have been updated together with the removal in the same PR. But unluckily it wasn't. I'll try to fix that as soon as I'll have time ...

PR #225, opened 2016-05-15
build: Use external translations

palinek:
Also use the lxqt_translate_ts from the liblxqt. But this
is only a build dependency, no lxqt component is needed in runtime.


agaida:
Tested and GTM

PR #228, opened 2016-05-23
Add keyboard cursor shape setting

### LINKED ISSUE #151: Cursor shape setting?
Looking at the code, it seems that the cursor shape in qterminal can be changed, but there does not seem to be a setting for it under "Preferences." Perhaps a menu could be added with the set of options included in a number of other terminal emulators?

f2404:
Closes #151 

Sorry for the `propertiesdialog.ui` diff - I edited the UI with Qt Designer and have no idea why it changed the file so much. Would be happy to receive an advice on reducing the change.


f2404:
@pvanek Thanks for merging all the stuff!

PR #234, opened 2016-05-26
Add location for custom color schemes

### LINKED ISSUE #141: Custom color schemes
It would be great if QTerminal supported adding custom color schemes besides those shipped by default. For example I'd like to use [base16](https://github.com/chriskempson/base16) colors.

f2404:
Depends on https://github.com/lxde/qtermwidget/pull/73
Closes #141


agaida:
With  lxde/qtermwidget#73 applied - tested and GTM

matkoniecz:
Thank you so much for implementing and merging of this! I hope to say that it is fine to say that I am really thankful for what you did here.

Without that I would be stuck with deciding which kind of unwanted effect I prefer (invisible dark blue? no contrast? glaring white background)

PR #235, opened 2016-05-31
Fix context menu actions being applied to the window opened last

### LINKED ISSUE #126: pasting clipboard by context menu takes place in window opened last
If more than one window is open pasting the clipboard by selecting "Paste Clipboard" from QTerminal's context menu always takes place in the window that got opened last no matter in which window the context menu is invoked.
Pasting clipboard by keyboard shortcut isn't affected.

Latest Git versions of both QTermWidget and QTerminal compiled against Qt 5.3.2 or 5.4.0 on Arch Linux x86_64.

f2404:
The problem is that the static actions stored within the `Properties` instance object get overwritten when a new terminal window is created which leads to them being applied to that last window.

Fixes #126

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -94,6 +94,7 @@ MainWindow::MainWindow(const QString& work_dir,
     setup_FileMenu_Actions();
     setup_ActionsMenu_Actions();
     setup_ViewMenu_Actions();
+    setup_ContextMenu_Actions();
     setupCustomDirs();
 
     /* The tab should be added after all changes are made to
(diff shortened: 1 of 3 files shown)


f2404:
This solution has a drawback - a separate context menu object containing all actions gets created for each tab opened...
Maybe store only action's parent widget (pointer to the `mainwindow` object) instead of the whole menu.

f2404:
The menu object could be stored within `TabWidget` object which; there is a single object for entire window even if contains multiple tabs. An interface for getting the menu would have to be created.
Opinions?

f2404:
Ok, I've revised my solution to store the context menu actions in mainwindow object, together with other menu objects.
Right click will be handled this way:
```c++
void TermWidgetImpl::customContextMenuCall(const QPoint & pos)
{
    const MainWindow *main = static_cast<MainWindow *>(parent()->parent()->parent()->parent()->parent()->parent()->parent());
    main->getContextMenu()->exec(mapToGlobal(pos));
}
```
How could these `parent()` calls be avoided? Pass the `MainWindow` pointer all the way down to `TermWidgetImpl`, through `TabWidget`, `TermWidgetHolder`, `TermWidget`? Any other ideas?

r-a-v-a-s:
What about [window()](http://doc.qt.io/qt-4.8/qwidget.html#window) or signals & slots?

f2404:
@r-a-v-a-s `window()` seems to do the trick, thanks!

f2404:
Please review the latest commit.

f2404:
You know, seeing how these PRs are treated is making me sad and frustrated... I can understand that the people here have their jobs and don't have much spare time, that's all right, but there are discussions and activities going on in other projects but not in qterminal and qtermwidget.
Is that because these projects are not needed and have no active maintainer?
Anyway, some people obviously have write access to the repos and could be merging the PRs from time to time, I think.
Thanks.

PR #244, opened 2016-06-09
Disable menu actions that are not applicable currently

f2404:
E.g., the switching tabs actions should be disabled (greyed out) if there is a single tab only.
Similarly, switching between the subterminals is impossible when there are no subterminals.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -20,6 +20,7 @@
 #include <QDesktopWidget>
 #include <QToolButton>
 #include <QMessageBox>
+#include <functional>
 
 #include "mainwindow.h"
 #include "tabwidget.h"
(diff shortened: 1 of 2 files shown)


f2404:
I'm using `MainWindow` member functions to determine if a certain action should be enabled or disabled: `hasMultipleTabs()` for tabs-related actions, `hasMultipleSubterminals` for subterminal-related ones.
The `std::function` objects contain pointers to that member functions and packed into the `data` fields of the actions, in form of `QVariant` object.
The function called right before the menu is shown (`aboutToShowActionsMenu()`) checks all actions and calls the function from the `data` field if it contains one.

Suggestions and improvements are welcome as usual.

agaida:
@f2404 - lgtm, can you please resolve the conflicts?

f2404:
@agaida Rebased.

agaida:
GTM

alexhenrie:
QTerminal does not compile after this change:

```
src/termwidget.cpp: In member function ‘void TermWidgetImpl::propertiesChanged()’:
src/termwidget.cpp:94:71: error: ‘setTerminalBackgroundImage’ was not declared in this scope
```

yan12125:
@alexhenrie You need to upgrade qtermwidget to the latest git master. ```setTerminalBackgroundImage``` was added in https://github.com/lxde/qtermwidget/pull/95

PR #251, opened 2016-07-17
Handle title changing ASCII escape sequences

yan12125:
PS. This relies on https://github.com/lxde/qtermwidget/pull/83

I follow the idea proposed by @palinek at https://github.com/lxde/qterminal/issues/52#issuecomment-227962047. That is, handle OSC commands and push changes to tabs and the main window. Suprisingly, OSC parsing functionality is already implemented in qtermwidget. What I do is just connecting signals and slots.

It's my first time writing Qt codes. Sorry if these changes looks naive.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -745,7 +745,7 @@ void MainWindow::addNewTab()
 
 void MainWindow::windowTitleChanged()
 {
-    QString userTitle = consoleTabulator->title();
+    const QString& userTitle = consoleTabulator->title();
     if (!userTitle.isNull())
     {
         setWindowTitle(userTitle);
@@ -754,7 +754,7 @@ void MainWindow::windowTitleChanged()
 
 void MainWindow::tabChanged()
 {
-    QString userTitle = consoleTabulator->title();
+    const QString& userTitle = consoleTabulator->title();
     if (!userTitle.isNull())
     {
         setWindowTitle(userTitle);
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -84,8 +84,8 @@ MainWindow::MainWindow(const QString& work_dir,
 
     consoleTabulator->setAutoFillBackground(true);
     connect(consoleTabulator, SIGNAL(closeTabNotification()), SLOT(close()));
-    connect(consoleTabulator, SIGNAL(titleChanged(QString, QString)),
-            this, SLOT(windowTitleChanged(QString, QString)));
+    connect(consoleTabulator, SIGNAL(titleChanged()), this, SLOT(windowTitleChanged()));
+    connect(consoleTabulator, SIGNAL(tabChanged()), this, SLOT(tabChanged()));
     consoleTabulator->setWorkDirectory(work_dir);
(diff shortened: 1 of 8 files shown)


yan12125:
More TODO:

1. When a new tab is opened and the no title setting command is received, the window title should be ```QTerminal```
2. Handle subterminal titles

palinek:
@jleclanche why did you assign this to me? I'm neither using nor developing qterminal

jleclanche:
@palinek You proposed the idea :) I don't know who else can review it, my C++ is too limited. @luis-pereira?

src/mainwindow.cpp
```diff
@@ -740,3 +742,21 @@ void MainWindow::addNewTab()
...
+}
+
+void MainWindow::tabChanged()
+{
+    const QString& userTitle = consoleTabulator->title();
+    if (!userTitle.isNull())
+    {
+        setWindowTitle(userTitle);
+    }
+}
```
f2404: What's the point of adding two slots doing the same thing?

yan12125: Seems I misunderstood the meaning of signals and slots. Will change them.

yan12125:
Found one more bug: if a tab is changed via mouse, the window title is not touched. Only tab switching via hotkeys works.

palinek:
I was a bit lost in the whole implementation (I'm seeing the qtermwidget/qterminal code for the first time), so I had to dig into it from the bottom to the top.... and this resulted into the new implementation -> #252

yan12125:
Sorry for bothering ... what happens here goes out of my expectation. My comments above are just reminders for myself, as I don't have much time on fixing them these two days.

yan12125:
#252 looks much better than mine. Let's move to there.

Justinzobel:
I'm running 0.14.1 but the Tab and Window title are still named "Shell No. 1".

yan12125:
I guess you've got some ideas from IRC? If there is still something wrong, make sure `Preferences` -> `Appearance` -> `Change window title based on current terminal` is checked. With that you shold be able to change the tab and window title with
```
echo -n "\e]2;some window title\a"
```

tsujan:
@yan12125 
I need this:
```bash
PROMPT_COMMAND='echo -ne "\e]2;some window title\a"'
```
Simple `echo` doesn't work.

In Konsole, this works here:
```bash
echo -ne "\033]30;New Name\007"
```
But it doesn't work in Qterminal. It should work in gnome-terminal too, I guess, but I don't have it.

So, it can be said that Qterminal's behavior isn't like other terminal emulators. Haven't looked into the code though.

tsujan:
Of course, this also works for me with QTerminal:
```bash
PROMPT_COMMAND= ;echo -ne "\e]2;some window title\a"
```
My point is that, for some reason unknown to me, playing with `PROMPT_COMMAND` isn't needed with Konsole or terminal emulators.

yan12125:
Hmm naybe related to this https://github.com/lxqt/qtermwidget/issues/124.

tsujan:
I can't say it's a bug but it would be nicer if QTerminal's title changed without setting `PROMPT_COMMAND`. I never use this feature but will read the code if I find the time.

Issue #52, opened 2014-04-14
Tab names should be dynamic


GinoM:
Names on Tabs should be dynamic and reflect running application, or at least support some customization via variables

Example: 
in Konsole
%d : %n 
allows for CurrentDir: CurrentRunnignProgram

pmattern:
+1

After having a closer look at Konsole, Xfce4 Terminal Emulator and xterm I'd suggest to go with the latter, which has been proposed recently in #111 as well.
Xfce4 Terminal basically behaves the same as xterm but has some albeit minor flaws regarding su. Konsole seems to feature several kinds of display none of which seems really helpful to me:
First there's '\<directory\> : \<process\>' which gets shown permanently in tabs except during SSH sessions and in windows' title bars by default. 'directory' is stating only the last one of the current path, e. g. "postfix" if either in /etc/postfix or /var/spool/postfix. Regarding su sessions ('su \<user\>') it sticks with the one su had been invoked in but doesn't display the one the current user is actually in. Also, no information about the current user is provided in su sessions.
During SSH sessions tabs' display turns into a permanent '(\<remote user\>) \<host\>' which doesn't provide information about the current path on the remote system either.
The last one is '\<user\>@\<hostname\>:\<path\>'. This gets displayed in windows' title bars if "Show window title on the title bar" is set and is basically the same as in xterm. But unlike xterm Konsole doesn't handle modifications of the window title as e. g. made by vim.

ryanreich:
+1 This is very useful in Gentoo, whose package manager sets the terminal title while merging.  Gentoo merges can be quite long; it's nice to know what package is currently in operation.

jleclanche:
PRs welcome :)

ryanreich:
If I'd ever used qt before, I'd be all over it.  Still considering it :)
ᐧ

jleclanche:
Don't hesitate to ask for mentorship on the lxde-list mailing list or the #lxde channel on Freenode.

palinek:
> and the window title as well... I just have to find how to do it

You just need to sniff for the [particular escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code). 

ESC ] -> Starts a control string for the operating system to use. OSC sequences are similar to CSI sequences, but are not limited to integer arguments. In general, these control sequences are terminated by ST.[13]:8.3.89 In xterm, they may also be terminated by BEL.[14] For example, in xterm, the window title can be set by `OSC 0;this is the window title BEL.`

[...]

PR #252, opened 2016-07-19
Add support for automatic title & icon modification

palinek:
Another implementation for  #52

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -636,6 +636,8 @@ void MainWindow::propertiesChanged()
         qobject_cast<BookmarksWidget*>(m_bookmarksDock->widget())->setup();
     }
 
+    onCurrentTitleChanged(consoleTabulator->currentIndex());
+
     Properties::Instance()->saveSettings();
     realign();
 }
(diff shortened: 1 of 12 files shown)


yan12125:
An idea after trying this branch: when a new tab is opened and no title-setting command is issued, the tab text is an empty string. I guess keep the old behavior "Shell no. XX" is better.

src/mainwindow.cpp
```diff
@@ -740,3 +738,16 @@ void MainWindow::addNewTab()
...
+{
+    QString title;
+    QIcon icon;
+    if (-1 != index)
+    {
+        title = consoleTabulator->tabText(index);
+        icon = consoleTabulator->tabIcon(index);
+    }
+    setWindowTitle(title.isEmpty() ? QStringLiteral("QTerminal") : title);
+    setWindowIcon(icon.isNull() ? QIcon::fromTheme("utilities-terminal") : icon);
```
yan12125: Maybe this line should be removed. For example if I run ```mpv xxx.mp4``` from QTerminal, the window icon becomes the same as mpv's. It brings inconvenience in LXQT, as I use icons in the Alt-Tab window to tell different applications.

palinek: So you propose, not to change the qterminal's main icon based on icon of selected terminal? Sounds reasonable... but IMO reasonable is it the other way round. Maybe we should make it configurable. What do you mean?

yan12125: > So you propose, not to change the qterminal's main icon based on icon of selected terminal?

Yes. And a configurable feature sounds a better option than disabling it.

palinek:
> An idea after trying this branch: when a new tab is opened and no title-setting command is issued, the tab text is an empty string. I guess keep the old behavior "Shell no. XX" is better.

Please, try with the @aff2562 (needs also update of the lxde/qtermwidget#84)

yan12125:
Thanks. Now looks all good.

palinek:
@yan12125 Now also the configuration for changing window title/icon should be done.

yan12125:
Configuration options regarding window title and icon work fine. Many thanks!

pmattern:
Works like a charm here.

GTM.

PR #255, opened 2016-08-02
Replace numbered terminals with directional navigation

pas2k:
This change replaces the former "next/previous" subterminal navigation with spatial left/right/top/bottom navigation based on split geometry, similar to Terminator.

This doesn't require additional bookkeeping and can be made optional, but I don't see any reason to use the older scheme over this one.

Important note: the change modifies the default control scheme to accommodate the changes, binding Ctrl+PgUp/PgDown for tab control, and using Shift+Up/Down for newly introduced "Top subterminal" and "Bottom subterminal" actions.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -37,8 +37,10 @@
 #define SPLIT_VERTICAL "Split Terminal Vertically"
 
 #define SUB_COLLAPSE "Collapse Subterminal"
-#define SUB_NEXT "Next Subterminal"
-#define SUB_PREV "Previous Subterminal"
+#define SUB_LEFT "Left Subterminal"
+#define SUB_RIGHT "Right Subterminal"
+#define SUB_TOP "Top Subterminal"
+#define SUB_BOTTOM "Bottom Subterminal"
 
 #define MOVE_LEFT "Move Tab Left"
 #define MOVE_RIGHT "Move Tab Right"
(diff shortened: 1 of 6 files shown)


pmattern:
Looks like a reasonable suggestion to me and works as expected when compiled against lxde/qtermwidget@42a2d78 on Arch Linux here. No obvious downsides.

If we decide to merge this the commits would have to be rebased. Probably there would be some conflict with the changes introduced in #275 / 0e64b978 in the meantime.

pas2k:
@pmattern Done and smoke-tested.

pmattern:
Running pas2k/qterminal@b436586 compiled against lxde/qtermwidget@248f29e on a new user profile, that is after deleting `$XDG_CONFIG_HOME/qterminal.org`, shortcuts `Shift+{Up,Down,Left,Right}` are no longer working here.
Problem didn't exist running under the conditions stated [above](https://github.com/lxde/qterminal/pull/255#issuecomment-282433770) so I guess it has to do with #275 / 0e64b978.

pas2k:
@pmattern Strangely enough, `Shift+{Up,Down,Left,Right}` are working for me just fine with the deleted `$XDG_CONFIG_HOME/qterminal.org` (the rest of the settings are lost, as expected).

What's even more strange, `Ctrl+PgDown|Ctrl+Tab` and `Ctrl+PgUp|Ctrl+Shift+Tab` didn't work for me right away. And the strangest thing about this is that I cannot reproduce this effect, even when removing `$XDG_CONFIG_HOME/qterminal.org` once again.

My versions are the same -- https://github.com/lxde/qterminal/pull/255/commits/b436586a919541a6d3db51c794749b2bf3dba0e8 and https://github.com/lxde/qtermwidget/commit/248f29e9ffeb8037f8edc179ac504e74915c89fd

And notice that if you've used Prefs->Shortcuts dialog to change the bindings, those are often not applied right away due to another bug -- #256.

pas2k:
Rebased against master. With all the work done recently on actions subsystem, the shortcut issues should hopefully be gone.

agaida:
pretty cool and impressive - GTM

Issue #238, opened 2016-06-03
Storing terminal properties in a singleton instance


f2404:
... which is common for all windows might not be a good idea: apart from possible [segfault](https://github.com/lxde/qterminal/issues/126) after the last opened window is closed, is also makes the settings look incorrectly in the UI after they'd been changed in another window which was closed afterwards.

Example:
- open qterminal window,
- open another window, change the color scheme - it will be changed for the 2nd window, the 1st window's scheme won't change,
- close the 2nd window,
- go to the 1st window's preferences - the color scheme will show the one from the closed window, not the actual one.

I can imagine that there are more that one example of such improper behavior.

r-a-v-a-s:
I support the removal of any or all singletons... ;-]

`Properties::Instance()->actions[CLEAR_TERMINAL]`

Is there any reason the action map can't be a member of the main window?
Writing `Properties::Instance()->` seems to make things much less convenient...
might as well just use QObject's `findChild` if we are going to write that much... >_<

f2404:
Another problem with the Properties singleton is that it's actually not destroyed, and the memory leaks.

agaida:
@yan12125 - is that still an issue?

tsujan:
This was fixed in https://github.com/lxqt/qterminal/commit/78755a23f00e87bd8f126e56492a76127c21f71b

PR #256, opened 2016-08-08
Refactor dangling actions, delete windows on close

pas2k:
These changes (probably partially) fix https://github.com/lxde/qterminal/issues/156
As a side effect, it changes the behavior where newly set shortcuts from Properties didn't apply to all windows.

Current version of QTerminal never deletes windows that are open via File->New Window, leading to out-of-control memory growth on prolonged use.
This patch allows use of setAttribute(Qt::WA_DeleteOnClose), or manual window deletion without harmful side effects.

This requires major refactoring of actions, since the previous implementation creates a lot of dangling pointers attached to Properties->actions on parent MainWindow deletion.

Since the changeset is unfortunately rather large, here's a breakdown of changes:
- Each MainWindow now has its own set of QActions, since those are destroyed along with their parent.
- There's a registry for MainWindows in singleton QTerminalApp, subclassed from QApplication
- Reusable actions are now only added dynamically, not via MOC, to allow for unified deletion and re-creation on setting changes.
- MainWindow children now use corresponding actions from parent MainWindow, allowing for their and actions' safe removal.
- To avoid code duplication, Properties and PropertiesDialog now use actions from the first MainWindow they find.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ set(QTERM_SRC
 )
 
 set(QTERM_MOC_SRC
+    src/qterminalapp.h
     src/mainwindow.h
     src/tabwidget.h
     src/termwidget.h
--- a/src/forms/qterminal.ui
+++ b/src/forms/qterminal.ui
@@ -111,21 +111,6 @@
     <string>About &amp;Qt...</string>
    </property>
   </action>
-  <action name="actProperties">
-   <property name="text">
-    <string>&amp;Preferences...</string>
-   </property>
-  </action>
-  <action name="actQuit">
-   <property name="icon">
-    <iconset theme="application-exit">
-     <normaloff/>
-    </iconset>
-   </property>
-   <property name="text">
-    <string>&amp;Quit</string>
-   </property>
-  </action>
  </widget>
  <customwidgets>
   <customwidget>
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -19,11 +19,13 @@
 #include <QApplication>
 #include <QtGlobal>
 
+#include <assert.h>
 #include <stdio.h>
 #include <getopt.h>
 #include <stdlib.h>
 
-#include  "mainwindow.h"
+#include "mainwindow.h"
+#include "qterminalapp.h"
 
 #define out
 
@@ -38,6 +40,8 @@ const struct option long_options[] = {
     {"profile", 1, NULL, 'p'},
     {NULL,      0, NULL,  0}
 };
+
+QTerminalApp * QTerminalApp::m_instance = NULL;
 
 void print_usage_and_exit(int code)
 {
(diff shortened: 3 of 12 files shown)


jleclanche:
lgtm at a glance. Someone for a second look please?

pas2k:
This commit already fixes #126, so https://github.com/lxde/qterminal/commit/abc8a91cf422c17b11ca989151759e9d07757d1c is unnecessary.

The current approach to action management is unmaintainable, both code and memory-wise. Almost every new UI feature or code hack for the bugs related to action management (e.g. aforementioned #126 or shortcut application in Properties described in the original pull request), will only make the issue worse.

The problem was also mentioned in #238.

agaida:
@pas2k - could you please rebase?

pas2k:
@agaida -- here you go. Since I had to actually port conflicting new features, it's important to keep a list of changes. Here's a squashed version, if necessary: [https://github.com/pas2k/qterminal/tree/window_leak_fix_squashed](/pas2k/qterminal/tree/window_leak_fix_squashed)

agaida:
@pas2k i would like to merge it, but the patch don't apply clean.

pas2k:
@agaida  Hmm. I see no conflicts. Did you try the "window_leak_fix_squashed" version I linked above? It's basically a rebased and squashed version of all the changes from window_leak_fix branch against the current QTerminal -- this should work even if window_leak_fix cannot be merged for some reason.

agaida:
no, i tried to appy the pull request as patch - does't look nice

pas2k:
I'm sorry, I don't quite get it. You mean, the changeset contains too many commits, so the history doesn't look nice? All right, I'll fix it tomorrow.

agaida:
@pas2k - no - checked out latest git and get the pr-diff - should be applicable with quilt - but it wasn't. Maybe i mess something up - will try it again tomorrow

pas2k:
Looked at the patch -- and indeed, some of the earlier commits are re-applied for some reason, maybe due to my cherry-picking for ports.

Here is a patch that will apply cleanly, containing all the same changes: 8785fc8127d55962bccb1f06a5323810f922b824

Here's how to apply it:
`curl https://github.com/pas2k/qterminal/commit/8785fc8127d55962bccb1f06a5323810f922b824.patch | patch -p1`

pas2k:
Resolved the conflicts with the current master. The patch is cleaner now since part of the refactoring is already implemented in 22e4e8ed962fe470879a6c36c27876beeb546deb.

jleclanche:
Thanks! Sorry for the delays.

@pas2k Are you interested in becoming a maintainer of qterminal?

pas2k:
@jleclanche No problem. As for becoming a maintainer, sure. I care a lot about qterminal since it became my primary workhorse, and would love to see it thrive. Not sure this qualifies me, but I certainly have the technical expertise and will try my best.

PR #273, opened 2016-10-25
Add preferences for background images

yan12125:
Depends on https://github.com/lxde/qtermwidget/pull/95, implements https://github.com/lxde/qtermwidget/issues/88


agaida:
GTM

PR #275, opened 2016-10-28
Add common shortcuts for switching tabs

alexhenrie:
Applications with tab interfaces, such as Firefox, commonly accept the shortcuts Ctrl+(Shift+)Tab and Ctrl+(PgUp|PgDown) to switch tabs. These shortcuts are especially convenient because they only require only one hand.

This patchset cleans up the shortcut setup code, adds the ability to have multiple shortcuts for an action, and adds the common shortcuts for switching tabs.

I chose to use the pipe character (|) to separate shortcuts in the config file and preferences dialog, in part because it is a common delimiter, and in part because it is very unlikely to be wanted as a shortcut.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -137,155 +137,101 @@ void MainWindow::setDropShortcut(QKeySequence dropShortCut)
     }
 }
 
-void MainWindow::setup_ActionsMenu_Actions()
+void MainWindow::setup_Action(const char *name, QAction *action, const char *defaultShortcut, const QObject *receiver,
+                              const char *slot, QMenu *menu, const QVariant &data)
 {
     QSettings settings;
     settings.beginGroup("Shortcuts");
 
     QKeySequence seq;
+
+    Properties::Instance()->actions[name] = action;
+    seq = QKeySequence::fromString(settings.value(name, defaultShortcut).toString());
+    Properties::Instance()->actions[name]->setShortcut(seq);
+
+    if (receiver)
+    {
+        connect(Properties::Instance()->actions[name], SIGNAL(triggered(bool)), receiver, slot);
+        addAction(Properties::Instance()->actions[name]);
+    }
+
+    if (menu)
+        menu->addAction(Properties::Instance()->actions[name]);
+
+    if (!data.isNull())
+        Properties::Instance()->actions[name]->setData(data);
+}
+
+void MainWindow::setup_ActionsMenu_Actions()
+{
     QVariant data;
 
     const checkfn checkTabs = &MainWindow::hasMultipleTabs;
     const checkfn checkSubterminals = &MainWindow::hasMultipleSubterminals;
 
-    Properties::Instance()->actions[CLEAR_TERMINAL] = new QAction(QIcon::fromTheme("edit-clear"), tr("&Clear Current Tab"), this);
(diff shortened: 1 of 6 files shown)


yan12125:
Count me as another fan of browser-like key bindings!

By the way, actions for switching tabs are TAB_PREV_SHORTCUT and TAB_NEXT_SHORTCUT. MOVE_LEFT_SHORTCUT and MOVE_RIGHT_SHORTCUT maps to Ctrl-Shift-PgUp and Ctrl-Shift-PgDown in Firefox.

alexhenrie:
Ahh you're right, I put the new shortcuts on the wrong action and didn't notice it with my quick testing. It's fixed now; merging this pull request will get the correct code.

yan12125:
Hope this one get reviewed and merged soon :)

src/mainwindow.cpp
```diff
@@ -330,56 +265,28 @@ void MainWindow::setup_FileMenu_Actions()
                            consoleTabulator, SLOT(preset4Terminals()));
     menu_File->addMenu(presetsMenu);
 
-    Properties::Instance()->actions[CLOSE_TAB] = new QAction(QIcon::fromTheme("list-remove"), tr("&Close Tab"), this);
-    seq = QKeySequence::fromString( settings.value(CLOSE_TAB, CLOSE_TAB_SHORTCUT).toString() );
-    Properties::Instance()->actions[CLOSE_TAB]->setShortcut(seq);
-    connect(Properties::Instance()->actions[CLOSE_TAB], SIGNAL(triggered()), consoleTabulator, SLOT(removeCurrentTab()));
-    menu_File->addAction(Properties::Instance()->actions[CLOSE_TAB]);
-    addAction(Properties::Instance()->actions[CLOSE_TAB]);
+    setup_Action(CLOSE_TAB, new QAction(QIcon::fromTheme("list-remove"), tr("&New Tab"), this),
```
alexhenrie: Good catch, fixed now.

pmattern:
Not sure whether the usage in the pipe *within the settings GUI* really makes for a good UX. On the other hand I couldn't tell a better way to handle an arbitrary, non-fixed number of shortcuts per setting.

That aside the commits are working as expected here, sure represent an important improvement and are GTM.

alexhenrie:
@pmattern I agree that the use of the pipe character in the GUI is not wonderful, but it's no worse than the current situation of having the user type the names of key combinations manually. Ideally, instead of typing something like "Ctrl+Shift+T" into the GUI, the GUI would have a button to capture the next pressed key combination and save it. Until we have something like that, I think typing pipe characters into the GUI is acceptable.

pmattern:
> Ideally, instead of typing something like "Ctrl+Shift+T" into the GUI, the GUI would have a button to capture the next pressed key combination and save it.

Agreed.

> Until we have something like that, I think typing pipe characters into the GUI is acceptable.

Sure, as said above.

yan12125:
lxqt-globalkeys has a shortcut selector: https://github.com/lxde/lxqt-globalkeys/blob/master/ui/shortcut_selector.cpp. Maybe it's useful for qterminal, too.

yan12125:
Hi @alexhenrie could you rebase? Things are changed after #244

agaida:
GTM - @yan12125  could you have a look again?

Issue #152, opened 2015-04-29
Add open link to context menu on links 


ianorlin:
In qterminal I would like the ability if I have a link like in irssi or other terminal based irc client man page to in the context menu be able to open the link with the context menu instead of having to hit control + click which is difficult for me. qterminal: An ability to copy the link could be nice as well. 

I am using version 
  Installed: 0.6.0-8
  Candidate: 0.6.0-8
  Version table:
 **\* 0.6.0-8 0
        500 http://ftp.uni-stuttgart.de/siduction/extra/ unstable/main amd64 Packages
        100 /var/lib/dpkg/status
on siduction. qterminal:
  Installed: 0.6.0-8
  Candidate: 0.6.0-8
  Version table:
 **\* 0.6.0-8 0
        500 http://ftp.uni-stuttgart.de/siduction/extra/ unstable/main amd64 Packages
        100 /var/lib/dpkg/status

LightGuard:
CTRL+click is nice, but :+1: to having it on the right button context menu as well.

yan12125:
Submitted pull request #276. Here's what I got:
![url-context-menu](https://cloud.githubusercontent.com/assets/1937689/19837312/84a8f676-9ef2-11e6-90c5-a5b0c3bebb1f.png)

yan12125:
This is implemented now

PR #276, opened 2016-10-30
Add context menu actions for URLs and E-Mail addresses

yan12125:
Implements #152

Depens on https://github.com/lxde/qtermwidget/pull/97

Context menus are now generated dynamically. According to valgrind, there are no memory leaks in the line `QMenu* contextMenu = new QMenu(this);`. I guess Qt is handling things well.


agaida:
Tested and GTM

PR #306, opened 2017-03-15
Properly clean up QApp

pas2k:
Following the comments from @palinek, added a proper call to the destructor for `QApplication` singleton. Not doing so might lead to the cryptic SEGFAULTS at the exit.


agaida:
GTM

PR #308, opened 2017-03-17
Support custom QSS styles

pas2k:
This implements the ability to style QTerminal with QSS by placing style.qss into the configuration directory.

Here's an example of what QSS might do:
qterminal-styles


pvanek:
awesome (personally I prefer system themes, but it's really nice)

agaida:
GTM

PR #309, opened 2017-03-17
New features: trim \n from pasted strings, confirm multiline pastes

pas2k:
This fixes few of my biggest pet peeves when using any terminal.

It implements 2 features:
1. Trimming trailing \n from pasted strings. 
2. Confirming multiline pastes, with a preview (in case those are still multiline after optional trimming). Usual Qt shortcuts (Enter, Esc, Space) work for the dialog, so it doesn't pull you out of the zone or from the keyboard.

Both features are optional, disabled by default, exposed via UI, and separate.

Rationale is, as follows:
1. Trimming trailing newlines makes triple left click to select a line actually useful in most apps (including, but not limited to terminal itself and browser). Even when editing texts, I found pasted newlines generally useless.
2. In addition to defense vs [paste attacks](http://thejh.net/misc/website-terminal-copy-paste), this allows you to not break your session by accidentally selecting something huge and pasting it. Same goes for confusing X `PRIMARY` buffer with X `CLIPBOARD` buffer.

qterminal-multiline-paste

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -423,6 +423,20 @@
             </property>
            </widget>
           </item>
+          <item row="3" column="0" colspan="3">
+           <widget class="QCheckBox" name="confirmMultilinePasteCheckBox">
+            <property name="text">
+             <string>Confirm multiline paste</string>
+            </property>
+           </widget>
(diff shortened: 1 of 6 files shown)


agaida:
GTM

PR #310, opened 2017-03-17
Restore filter actions

yan12125:
Accidentally removed in https://github.com/lxde/qterminal/pull/256. Those were originally added in #276. In this PR I also fixed a bug - an extra separator if there are no extra filter actions.


pas2k:
Oops -- my fault. Probably slipped during conflict resolution in one of the rebases.

yan12125:
No worry! Rebasing is never simple :-)

agaida:
GTM

PR #313, opened 2017-03-20
Fix duplicated items, rework MainWindow memory-management

### LINKED ISSUE #312: Duplicated menu entries after properties changed
This seems a bug caused by #256. After changing some properties (for example uses a smaller font size or vice versa), there are duplicate actions in the menu.
duplicated_actions

I guess ```rebuildActions``` does not clear previous actions completely. @pas2k Could you have a look?

pas2k:
These changes fix the memory management inside MainWindow related to setting changes.

QActions and QMenus allocated with `new`, were bound to the window itself, so those weren't cleaned up when replaced. The issue can be reproduced by opening Settings and repeatedly pressing Apply while monitoring QTerminal memory consumption.

The leaks aren't visible on Valgrind since the leaked objects are properly deleted when the window closes, so Massif was used for debugging.

The fix is twofold:
- Instead of binding all constructed QObjects to the window itself, a proxy object `settingOwner` is used. It's destroyed and re-created on setting changes, so all child objects follow.
- The nested menus aren't re-created on Setting changes. The `settingOwner` trick would have reduced the changeset, reworking nested menus dynamically result in `DBusMenuExporterDBus` warnings for some reason. 

This also fixes #312

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -50,6 +50,14 @@ MainWindow::MainWindow(TerminalConfig &cfg,
                        Qt::WindowFlags f)
     : QMainWindow(parent,f),
       DBusAddressable("/windows"),
+      tabPosition(NULL),
+      scrollBarPosition(NULL),
+      keyboardCursorShape(NULL),
+      tabPosMenu(NULL),
+      scrollPosMenu(NULL),
+      keyboardCursorShapeMenu(NULL),
+      settingOwner(NULL),
+      presetsMenu(NULL),
       m_config(cfg),
       m_dropLockButton(0),
       m_dropMode(dropMode)
@@ -118,16 +126,15 @@ MainWindow::MainWindow(TerminalConfig &cfg,
 
 void MainWindow::rebuildActions()
 {
-    QMap< QString, QAction * > oldActions(actions);
+    // Delete all setting-related QObjects
+    delete settingOwner;
+    settingOwner = new QWidget(this);
+    settingOwner->setGeometry(0,0,0,0);
 
+    // Then create them again
     setup_FileMenu_Actions();
     setup_ActionsMenu_Actions();
     setup_ViewMenu_Actions();
-
-    for (const auto *a : const_cast<const QMap<QString, QAction*>&> (oldActions))
-    {
-        delete a;
-    }
 }
 
 MainWindow::~MainWindow()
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -111,16 +119,14 @@ MainWindow::MainWindow(const QString& work_dir,
 
 void MainWindow::rebuildActions()
 {
-    QMap< QString, QAction * > oldActions(actions);
+    // Delete all setting-related QObjects
+    delete settingOwner;
+    settingOwner = new QWidget(this);
```
yan12125: Is ```QWidget``` necessary? I guess ```QObject``` is enough.

src/mainwindow.h
```diff
@@ -48,6 +48,10 @@ class MainWindow : public QMainWindow, private Ui::mainWindow
     QActionGroup *tabPosition, *scrollBarPosition, *keyboardCursorShape;
     QMenu *tabPosMenu, *scrollPosMenu, *keyboardCursorShapeMenu;
 
+    QWidget *settingOwner;
```
yan12125: Could you add some comments for this object? It's not intuitive from the name.

src/mainwindow.cpp
pas2k: If menus are to be re-parented, then yes, QWidget is necessary. But since menus are not re-parented due to DBusMenuExporterDBus warnings, it's not that necessary. Can be useful in future for other setting-related QWidgets, so I'd say it's justified.

agaida:
@pas2k - would you be so kind and rebase - GTM when rebased.

pas2k:
@agaida Done -- and ready to rebase again once some other commits are merged.

agaida:
hmm - the patch applies, the package build and install fine - there is only one problem, i can't click on File and Action any longer reliable - both menu entries react not reliable or not on mouseclicks. all other top level menu entries do. The keyboard shortcuts work just fine.

agaida:
@pas2k - sorry, could you rebase again, really not my day

pas2k:
@agaida No problem, done. As for File/Action, I'm not seeing any difference in functionality or responsiveness compared to master. I'm on Ubuntu 17.04, recently updated from 16.10 (where this was also fine, I use qterminal on daily basis), qt is stock.

agaida:
good, will test it again - i was a little bit surprised too

agaida:
@pas2k - ok, it is weird - but i made a little video: https://youtu.be/9b22pfd-xZs

agaida:
ok, next information - the menu behave ok after i opened preferences and hit ok or apply - cancel is not sufficient.

pas2k:
@agaida Thanks for the video! I was able to reproduce the issue on non-ubuntu VM with no menubar integration. Turns out the *invisible parent QWidget* was blocking the menu. Doh! Fixed by setting its geometry to 0 (to avoid possibly hiding children with hide()). Should be fine now.

agaida:
Cool, thanks.

agaida:
GTM

PR #322, opened 2017-05-01
Update main.cpp : High-DPI Pixmaps

beanpole135:
Enable the High-DPI pixmap option.
This fixes the pixellation issues on 4K monitors.


agaida:
After a good documentation reading: GTM

PR #325, opened 2017-05-03
DBus DropMode

pas2k:
Since Terminator completely broke on recent Ubuntu, I decided to switch to QTerminal drop mode as well as the primary terminal app.

Because drop mode Global Shortcut never worked for me on stock Unity, I've implemented a workaround via the existing DBus interface using Process->toggleDropdown(), which should be invoked via WM hotkey handler.

The PR also implements an ability to distinguish between "normal" and "drop mode" processes using Process->isDropMode()

In other changes, I've sneaked in a simple check for unchanged window geometry to prevent unnecessary redraws I've experienced on every visibility toggle.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -289,5 +289,26 @@ QDBusObjectPath QTerminalApp::getActiveWindow()
     return qobject_cast<MainWindow*>(aw)->getDbusPath();
 }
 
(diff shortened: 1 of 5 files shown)


agaida:
Cool thing - i see no downsides yet and using dbus i the right idea. GTM

PR #327, opened 2017-05-06
Change subterminal shortcuts to avoid breaking qtermwidget scrolling

### LINKED ISSUE #326: Shortcut conflicts: termwidget scrolling & subterminal switching
Since lxde/qtermwidget@934107ae7de8dbabd7556c0b7d38b6f17b0e2187, Shift + Up/Down is used to scroll the termwidget. Shift + Up scrolls the term widget up by one line, and Shift + Down does the opposite. In application screens (tmux, vim, etc.), Shift + Up/Down/Left/Right are also used to send specific ASCII escape sequences.

In 5fa724c566b2b46b4295325346813abb0aa052e3, Shift + arrow keys are used to switch between subterminals. As this has a higher proirity, scrolling functions does not work anymore. Maybe we can have a different set of default shortcuts for switching subterminals? I don't use subterminals in daily work, maybe @pas2k has some ideas?

yan12125:
Fix #326

Alt + Direction keys are borrowed from Terminator. It may conflict with TUI applications. If so users can change shortcuts from the preference. The point of this PR is to remove self-conflicts from QTerminal defaults.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -74,10 +74,10 @@
 
 #define TAB_PREV_SHORTCUT	       "Ctrl+PgUp|Ctrl+Shift+Tab"
 #define TAB_NEXT_SHORTCUT	       "Ctrl+PgDown|Ctrl+Tab"
-#define SUB_BOTTOM_SHORTCUT	       "Shift+Down"
-#define SUB_TOP_SHORTCUT	       "Shift+Up"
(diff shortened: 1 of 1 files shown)


agaida:
Good idea - GTM

yan12125:
Thanks!

PR #330, opened 2017-05-08
Fix toggle menu action.

### LINKED ISSUE #329: Toggle Menu does the same thing as Find
When I press Ctrl+Shift+M it performs the same action as Find (Ctrl+Shift+F). I would expect it to toggle the visibility of the menu bar.

EDIT:
The same thing happens if I right click and select "Tooge Menu"

tmccombs:
Toggle menu with toggle menu action rather than calling `find()`.

Fixes #329


yan12125:
Thanks for finding the problem and the fix!

PR #331, opened 2017-05-08
Make disabled actions consistent at all times

### LINKED ISSUE #328: Shortcuts not captured if a disabled menu item is shown
Steps to reproduce:

1. Run QTerminal. Currently there's only one tab.
2. Click on ```Actions``` menu. As there's only one tab, ```Next Tab``` and ```Previous Tab``` are disabled.
3. Create a new tab by ```File``` -> ```New Tab```
4. Try to switch tabs by hotkeys. I use ```Ctrl + PgDown``` and ```Ctrl + PgUp``` here

Now if I hit Ctrl + V and then Ctrl + PgDown, ```^[[5;5~``` is printed on the shell, which indicates that Ctrl + PgDown is captured by qwermwidget rather than qterminal.

If I hit the Actions menu again after creating a new tab, Ctrl + PgDown and Ctrl + PgUp works as usual.

Subterminals have a similar issue. After creating a new subterminal, I have to click on the Actions menu again to make subterminal-related shortcuts work.

I have two machines. One uses Qt 5.9 beta 3 and another has Qt 5.8. Both of them use Arch Linux. The bug is reproducible on both systems, but @pas2k said it's not reproducible on Ubuntu. I'm not sure whether it's a Qt bug or not. Maybe Ubuntu has some local patches for Qt that fixes it.

pas2k:
Fixes #328

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -116,7 +116,7 @@ MainWindow::MainWindow(TerminalConfig &cfg,
     setupCustomDirs();
 
     connect(consoleTabulator, &TabWidget::currentTitleChanged, this, &MainWindow::onCurrentTitleChanged);
-    connect(menu_Actions, SIGNAL(aboutToShow()), this, SLOT(aboutToShowActionsMenu()));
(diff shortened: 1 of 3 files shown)


agaida:
GTM

PR #338, opened 2017-07-03
Drops Qt5Core_VERSION_STRING

luis-pereira:
Use Qt5Core_VERSION. Qt5Core_VERSION_STRING is a compatibility variable and
it was removed in Qt 5.9 release.


agaida:
GTM

PR #350, opened 2017-08-21
Fix action inconsistency when switching tabs

pas2k:
Fixes an issue where navigation between panes doesn't work when switching between tabs.

Steps to reproduce the issue this PR fixes:
- Open QTerminal
- Split pane vertically using keyboard shortcuts.
- Ensure that navigation using keyboard shortcuts works
- Create a tab using keyboard shortcuts (assuming new tab has a single pane)
- Switch back to original tab using keyboard shortcuts
- Notice that navigation between panes no longer works


agaida:
GTM

PR #359, opened 2017-10-04
Made font in settings file human readable

### LINKED ISSUE #218: qterminal font settings in qterminal.ini
font settings in qterminal.ini should be human readable and changeable, the current state is far form good:

```
 colorScheme=Linux
 emulation=default
-font=@Variant(\0\0\0@\0\0\0\x12\0M\0o\0n\0o\0s\0p\0\x61\0\x63\0\x65@(\0\0\0\0\0\0\xff\xff\xff\xff\x2\x1\0\x32\x10)
+font=@Variant(\0\0\0@\0\0\0\x12\0N\0o\0t\0o\0 \0M\0o\0n\0o@&\0\0\0\0\0\0\xff\xff\xff\xff\x2\x1\0\x32\x10)
```

YourFin:
Closes #218

### DIFF
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -72,6 +72,8 @@ void Properties::loadSettings()
 
     font = QFont(qvariant_cast<QString>(m_settings->value("fontFamily", defaultFont().family())),
                  qvariant_cast<int>(m_settings->value("fontSize", defaultFont().pointSize())));
+    //Legacy font setting
+    font = qvariant_cast<QFont>(m_settings->value("font", font));
 
     mainWindowSize = m_settings->value("MainWindow/size").toSize();
     mainWindowPosition = m_settings->value("MainWindow/pos").toPoint();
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks for the fix! Could you add transition codes so that existing users with ```font``` in their configuration files can get the new keys without reconfiguration?

YourFin:
That should take care of it. In the process I noticed that the `defaultFont` method adds a style hint for `TypeWriter`; `Monospace` [looks more appropriate](http://doc.qt.io/qt-4.8/qfont.html), but I can't claim to be an expert on the topic.

YourFin:
@yan12125 is there anything else I should do to get this merged?

YourFin:
Sorry closed by accident

yan12125:
Sorry for being late - I have no access to my developing machine until tonight.

Here's an issue: If I migrated from an old configuration file, new font settings won't be saved. Steps to reproduce:

1. Build current git master of qterminal
2. Change the font with the preferences dialog. For example I use "DejaVu Sans Mono 18"
3. Build with this patch
4. Change the font to something else. For example to "DejaVu Sans Mono 19"
5. Restart the terminal. The font used is still "DejaVu Sans Mono 18"

Now qterminal.ini has both ```font``` and ```fontFamily```/```fontSize```, and the latter two are ignored. I guess deleting the ```font``` key during migration can fix it.

YourFin:
Yeah, sorry I didn't realize that was an issue. It would be relatively easy to solve by just having the save settings function clobber the `font` setting, but I don't know how to do that off the top of my head. I'll look into it later today.

YourFin:
Ok, it should be fixed now. That was easier than I thought it would be. Thanks for your time!

yan12125:
Thanks! It works fine!

PR #363, opened 2017-10-19
Add maximum tab width and proper text eliding.

nathan-osman:
One of the things I noticed with qterminal is that the width of each tab changes based on the length of the text. This is fine if the current directory is `/tmp` but not if the current directory is `/usr/share/qtcreator/templates/wizards/classes/cpp`.

The tab becomes very wide and causes scrollbars to appear, making it difficult to navigate to other tabs with the mouse.

This pull request attempts to address that by limiting the maximum width of the tab to 300 pixels. Text that is longer than the tab width will be elided (an ellipsis shown at the end). The `TabBar` and `TabStyle` classes are necessary to correctly implement this behavior since the `QTabBar::setElideMode()` method doesn't work [due to a bug](https://bugreports.qt.io/browse/QTBUG-48994).

Here's a screenshot showing before and after:

titlebar

Note that the full path is still shown in the titlebar.

### DIFF
--- a/src/tabwidget.h
+++ b/src/tabwidget.h
@@ -28,6 +28,7 @@
     #include "dbusaddressable.h"
 #endif
 
+#include "tabbar.h"
 #include "terminalconfig.h"
 #include "properties.h"
 
@@ -111,6 +112,8 @@ private:
     int tabNumerator;
     /* re-order naming of the tabs then removeCurrentTab() */
     void renameTabsAfterRemove();
+
+    TabBar *mTabBar;
 };
 
 #endif
--- a/src/tabwidget.h
+++ b/src/tabwidget.h
@@ -28,10 +28,10 @@
     #include "dbusaddressable.h"
 #endif
 
-#include "tabbar.h"
 #include "terminalconfig.h"
 #include "properties.h"
 
+class TabBar;
 class TermWidgetHolder;
 class QAction;
 class QActionGroup;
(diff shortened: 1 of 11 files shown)


yan12125:
This is useful! How about make the width configurable?

nathan-osman:
Done:

configure

yan12125:
Some ideas after testing:

1. Enable it by default sounds fine. A larger default value should be better. On high-resolution screens, 300px is really small.

2. Here's a bug: with this patch icons on the tab bar are gone. Before:

image

After:
image

I'm using KDE's Breeze theme engine. Not sure if that makes a difference, though.

3. Maybe ElideLeft or ElideMiddle is better. In many shell configurations, the terminal title is set to ```user@hostname:/path```. The first few characters are always the same, so showing characters from the end help identifying tabs.

src/tabwidget.h
```diff
@@ -28,6 +28,7 @@
     #include "dbusaddressable.h"
 #endif
 
+#include "tabbar.h"
```
yan12125: A minor point: as there are only pointers, forward declaration can be used. (i.e., ```class TabBar;```) This reduces a little compiling time.

nathan-osman:
I was able to fix everything but the icons. How does qterminal normally display icons? No matter what I do, I can't seem to get any icons to display in the tabs, even without my patch.

nathan-osman:
Ah, never mind. I figured it out and fixed the icon display.

agaida:
@nathan-osman could you please rebase?

nathan-osman:
I can do that as soon as I get home from work tomorrow.

agaida:
Cool, thanks

nathan-osman:
Rebase complete.

PR #368, opened 2017-10-31
Support secondary font

blahgeek:
Context:

I use vim in terminal to write codes. I like qterminal because it supports font ligatures so that codes in vim looks great. (I use [Fira Code](https://github.com/tonsky/FiraCode)). However there's a little problem: font ligatures are designed for codes, but not other terminal applications. E.g., the output of `htop` would looks strange (see below, `||` is displayed as ligature so not aligned):

So basiclly I want to use different font for different terminal applications. Here I want to use "Fira Code" for vim, "Fira Mono" for everything else. This PR does exactly this.

Every time terminal title changes, it's matched against the regex pattern from user settings. If matched, the "secondary font" would be used, otherwise the default font.

----

I understand that this feature is not something that everyone would use (too specific to myself), so there's little chance that this PR would be merged. But I really want to know your thoughts about this :P

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>864</width>
-    <height>728</height>
+    <height>920</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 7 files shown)


yan12125:
Here's another idea - implementing [font-changing escape codes (OSC 50)](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Operating-System-Commands) in qtermwidget.

If OSC 50 is done, vim autocmds (VimEnter/VimLeave, [VimSuspend/VimResume](https://github.com/neovim/neovim/issues/3648)) can be used to change fonts [via writefile()](https://github.com/neovim/neovim/issues/7584)

yan12125:
Implementing konsoleprofile is also fine. I just prefer a simpler solution.

yan12125:
Yep - I'd prefer solutions with more flexibilities like Konsole's profile or [the font changing command of XTerm](https://stackoverflow.com/a/13422481/3786245).

PR #373, opened 2017-11-27
Add RTL support checkbox

### LINKED ISSUE #98: RTL text is displayed reversed
Any RTL texts (such As Arabic and Hebrew) are reversed, The first letter on the left and the last on the right, Which is wrong.
qterminalrtl0
The first is QTerminal (Displayed wrong), The second is Konsole (Displayed right).

SafaAlfulaij:
This adds a checkbox in the settings (appearance section) for enabling bidi support.
Must be done (and depend on) after lxde/qtermwidget#155 land.

Fixes #98

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -344,6 +344,13 @@
          </property>
         </widget>
        </item>
+       <item row="12" column="0" colspan="2">
+        <widget class="QCheckBox" name="enabledBidiSupportCheckBox">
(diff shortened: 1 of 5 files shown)


yan12125:
Thanks for this! Should bidi be enabled for correct RTL text rendering? If so, I think it's good to have it on by default.

Ref: https://bugs.kde.org/show_bug.cgi?id=198169 (Add bidirectional text support in default profile)

PR #382, opened 2017-12-30
Fixes tabAt() when tabBar is not at north position

jsm222:
When renaming a tab with the context menu when the tabBar is not
at the north position, the tabAt was returning -1 due to wrongly
calculated postition for tabAt(). This is now fixed.
See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=224718
and https://bugreports.qt.io/browse/QTBUG-16536


agaida:
GTM

PR #386, opened 2018-01-10
Fix terminal menubar

tsujan:
Setting `autoFillBackground` to `true`, to prevent a fully transparent menubar with styles that have translucency, was a bad idea (of mine) because it presupposed a flat background, while styles like Oxygen, Kvantum and QtCurve can have gradient for background. Instead, setting `WA_NoSystemBackground` to `false` fixes both issues and allows terminal translucency with compositing.

### DIFF
--- a/src/forms/qterminal.ui
+++ b/src/forms/qterminal.ui
@@ -13,9 +13,6 @@
   <property name="windowTitle">
    <string>MainWindow</string>
   </property>
-  <property name="autoFillBackground">
-   <bool>true</bool>
-  </property>
   <widget class="QWidget" name="centralwidget">
    <layout class="QGridLayout">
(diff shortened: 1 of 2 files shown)


tsujan:
For the sake of certainty, I checked Konsole's code and saw it did exactly the same thing :) 

Konsole → MainWindow.cpp → MainWindow::MainWindow():
```cpp
    if (useTransparency()) {
        // It is useful to have translucent terminal area
        setAttribute(Qt::WA_TranslucentBackground, true);
        // But it is mostly annoying to have translucent menubar and tabbar
        setAttribute(Qt::WA_NoSystemBackground, false);
    }
```

agaida:
GTM

Issue #304, opened 2017-03-02
SIGSEGV after closing all tabs and then create a new tab


yan12125:
Steps to reproduce:

1. Launch QTerminal
2. Close all terminals with Ctrl+D
3. When asked really leaving QTerminal or not, hit "No"
4. Open a new tab

gdb backtrace:
```
Thread 1 "qterminal" received signal SIGSEGV, Segmentation fault.
0x00007ffff7b7cd55 in QTermWidget::workingDirectory (this=0x77b040) at /home/yen/Projects/qtermwidget/lib/qtermwidget.cpp:385
385         if (!m_impl->m_session)
(gdb) bt
#0  0x00007ffff7b7cd55 in QTermWidget::workingDirectory (this=0x77b040) at /home/yen/Projects/qtermwidget/lib/qtermwidget.cpp:385
#1  0x0000000000445912 in TabWidget::addNewTab (this=0x733850, shell_program=...) at /home/yen/Projects/qterminal/src/tabwidget.cpp:76
#2  0x000000000043f9ef in MainWindow::addNewTab (this=0x718ca0) at /home/yen/Projects/qterminal/src/mainwindow.cpp:672
#3  0x000000000046bda7 in MainWindow::qt_static_metacall (_o=0x718ca0, _c=QMetaObject::InvokeMetaMethod, _id=15, _a=0x7fffffffd030)
    at /home/yen/Projects/qterminal/src/moc_mainwindow.cpp:151
#4  0x00007ffff5df3d49 in QMetaObject::activate(QObject*, int, int, void**) () from /usr/lib/libQt5Core.so.5
#5  0x00007ffff6cecc92 in QAction::triggered(bool) () from /usr/lib/libQt5Widgets.so.5
#6  0x00007ffff6cef3f0 in QAction::activate(QAction::ActionEvent) () from /usr/lib/libQt5Widgets.so.5
#7  0x00007ffff6cefd6c in QAction::event(QEvent*) () from /usr/lib/libQt5Widgets.so.5
#8  0x00007ffff6cf334c in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /usr/lib/libQt5Widgets.so.5
#9  0x00007ffff6cfab61 in QApplication::notify(QObject*, QEvent*) () from /usr/lib/libQt5Widgets.so.5
#10 0x00007ffff5dc8440 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /usr/lib/libQt5Core.so.5
[...]

PR #388, opened 2018-01-13
Fixes an error in the d3fa804 fix of #304

jsm222:
I by mistake also removed tabs, when one cancels the close event. This is now corrected.


agaida:
GTM

Issue #356, opened 2017-09-20
Make active tab somehow darker/lighter than the non active tabs (make active noticeable)


je-vv:
I'm currently using Arch, and I'm using qterminal under fluxbox + compton + qt5c + arc-gtk-theme, and one annoying thing is that it's hard to make a difference by looking at the tabs which is the active one.

Granted using the "windows" widget style makes the "closing tab button" red on active, but the difference should be noticed regardless of the style, and perhaps darker/lighter is an easy approach, :)

##### System Information

* Distribution & Version:  Arch GNU/Linux (up to date)
* Kernel:   4.12.13
* Qt Version:  5.9.1
* libqtxdg Version:  2.0.0 
* qtermwidget Version:  0.7.1
* lxqt-build-tools Version:  0.3.2 (not installed)
* Package version:  0.7.1

yan12125:
Could you paste a screenshot like this? Not quite sure what qterminal looks like under your configuration.
![default](https://user-images.githubusercontent.com/1937689/30631452-d326b3e6-9e16-11e7-91a8-e6ed02edbef6.png)

je-vv:
![image](https://user-images.githubusercontent.com/10799432/30632032-e0d545c0-9da3-11e7-8a61-42906c0b82bf.png)

je-vv:
Just in case, the .config/qterminal.org/qterminal.ini file:

[General]
AskOnExit=false
BookmarksFile=/home/general/.config/qterminal.org/qterminal_bookmarks.xml
BookmarksVisible=true
Borderless=false
ChangeWindowIcon=true
ChangeWindowTitle=true
HideTabBarWithOneTab=false
HistoryLimited=true
HistoryLimitedTo=1024
KeyboardCursorShape=0
MenuVisible=false
MotionAfterPaste=0
SavePosOnExit=true
SaveSizeOnExit=true
ScrollbarPosition=2
TabBarless=false
TabsPosition=0
TerminalBackgroundImage=
TerminalTransparency=25
TerminalsPreset=0
UseBookmarks=false
UseCWD=false
colorScheme=WhiteOnBlack
emulation=default
font=@Variant(\0\0\0@\0\0\0\x1a\0x\0o\0s\0\x34\0 \0T\0\x65\0r\0m\0i\0n\0u\0s@ \0\0\0\0\0\0\xff\xff\xff\xff\x2\x1\0\x32\x10)
guiStyle=qt5ct-style
highlightCurrentTerminal=false
[...]

PR #389, opened 2018-01-15
Make active tab text bold to make it more visible

kdave:
This partially addresses issue #356. The active tab is looking almost
the same as an inactive one. This probably depends on system theme
settings or maybe is platform dependent.

The styling can be changed by either external stylesheet or an "in-line"
via the QWidget::styleSheet property.

This patch adds a simple in-line styleshee to change the active tab text
to bold. This is IMHO sufficient to recognize the active tab immediatelly,
further refinements are possible (eg. configurable color etc.)


agaida:
Looks good, i like it a lot. GTM

PR #390, opened 2018-01-15
Fix a typo

luis-pereira:
Introduced by commit 23fb6f84fdb3.
Not caught due to so many warnings.... one more, one less... nobody notices
it.


agaida:
every typo less counts :D - GTM

luis-pereira:
Yes. We should get rid of so many warnings.
I'm gone try to cure my "terrible" flu.

PR #392, opened 2018-01-26
Allow to change tab title color

kdave:
Add a context menu action that opens a color picker dialog. The selected
color will be used for the tab title, background is left intact.

The icon for the menu item is 'color-management', that seems to be
present in at least breeze and oxygen themes.

### DIFF
--- a/src/tabwidget.cpp
+++ b/src/tabwidget.cpp
@@ -18,6 +18,7 @@
 
 #include <QTabBar>
 #include <QInputDialog>
+#include <QColorDialog>
 #include <QMouseEvent>
 #include <QMenu>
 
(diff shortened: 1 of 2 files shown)


src/tabwidget.cpp
```diff
@@ -58,6 +59,7 @@ TabWidget::TabWidget(QWidget* parent) : QTabWidget(parent), tabNumerator(0), mTa
     connect(this, SIGNAL(tabCloseRequested(int)), this, SLOT(removeTab(int)));
     connect(tabBar(), SIGNAL(tabMoved(int,int)), this, SLOT(updateTabIndices()));
     connect(this, SIGNAL(tabRenameRequested(int)), this, SLOT(renameSession(int)));
+    connect(this, SIGNAL(tabTitleColorChangeRequested(int)), this, SLOT(setTitleColor(int)));
```
yan12125: As you're writing new codes, could you use the new Qt5 style of ```connect()```?

kdave: Sure, done. Branch updated.

agaida: S

agaida:
convert the 'old' connects with the next PR

PR #394, opened 2018-01-31
Convert connect to the new syntax

kdave:
Convert connect calls that use the string-based syntax to the new one based on member pointer. Most of the conversions were straightforward, but there are some that are not:
* overloaded slots need to be explicitly cast, using static_cast, it looks ugly, the syntax used is from the qt wiki
* lambdas were converted to member functions
* inherited singals/slots must be typed to the base class
* signal and slot parameters must match, in simple cases dummy params were added, otherwise the static_cast would be needed
* implicit parameter values do not work anymore, but were used in some cases to keep the changes minimal where functions had no params and do not care

There's one conversion not done yet. It's in mainwindow.cpp, MainWindow::setup_Action. The slot name is passed as string. I did not succeed to convert that to the member pointer syntax with. The code could be refactored to avoid passing the slot as argument, but I'd rather get an ack before I do that.

### DIFF
--- a/src/fontdialog.cpp
+++ b/src/fontdialog.cpp
@@ -32,12 +32,12 @@ FontDialog::FontDialog(const QFont &f)
 
     sizeSpinBox->setValue(f.pointSize());
 
-    setFontSample();
+    setFontSample(f);
 
-    connect(fontComboBox, SIGNAL(currentFontChanged(QFont)),
-            this, SLOT(setFontSample()));
-    connect(sizeSpinBox, SIGNAL(valueChanged(int)),
-            this, SLOT(setFontSample()));
+    connect(fontComboBox, &QFontComboBox::currentFontChanged,
+            this, &FontDialog::setFontSample);
+    connect(sizeSpinBox, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged),
+            this, &FontDialog::setFontSize);
 }
 
 QFont FontDialog::getFont()
@@ -47,9 +47,16 @@ QFont FontDialog::getFont()
     return f;
 }
 
-void FontDialog::setFontSample()
+void FontDialog::setFontSample(const QFont &f)
 {
-    QFont f = getFont();
+    previewLabel->setFont(f);
+    QString sample("%1 %2 pt");
+    previewLabel->setText(sample.arg(f.family()).arg(f.pointSize()));
+}
+
+void FontDialog::setFontSize()
+{
+    const QFont &f = getFont();
     previewLabel->setFont(f);
     QString sample("%1 %2 pt");
     previewLabel->setText(sample.arg(f.family()).arg(f.pointSize()));
--- a/src/fontdialog.h
+++ b/src/fontdialog.h
@@ -32,7 +32,8 @@ public:
     QFont getFont();
 
 private slots:
-    void setFontSample();
+    void setFontSample(const QFont &f);
+    void setFontSize();
 
 };
 
--- a/src/tabwidget.cpp
+++ b/src/tabwidget.cpp
@@ -56,9 +56,9 @@ TabWidget::TabWidget(QWidget* parent) : QTabWidget(parent), tabNumerator(0), mTa
 
     tabBar()->installEventFilter(this);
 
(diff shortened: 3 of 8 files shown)


agaida:
hi - not really sure who can review it - so i added @palinek @luis-pereira and @tsujan - if i'm wrong with, sorry :)

src/fontdialog.cpp
```diff
@@ -47,9 +47,16 @@ QFont FontDialog::getFont()
...
-void FontDialog::setFontSample()
+void FontDialog::setFontSample(const QFont &f)
 {
-    QFont f = getFont();
+    previewLabel->setFont(f);
+    QString sample("%1 %2 pt");
+    previewLabel->setText(sample.arg(f.family()).arg(f.pointSize()));
+}
+
+void FontDialog::setFontSize(int i)
```
tsujan: `i` isn't used:
```c++
void FontDialog::setFontSize(int/* i*/)
```

src/fontdialog.h
```diff
@@ -32,7 +32,8 @@ class FontDialog : public QDialog, public Ui::FontDialog
     QFont getFont();
 
 private slots:
-    void setFontSample();
+    void setFontSample(const QFont &f);
+    void setFontSize(int i);
```
tsujan: `i` isn't used:
```c++
void setFontSize(int);
```

src/termwidget.cpp
```diff
@@ -296,10 +296,10 @@ TermWidget::TermWidget(TerminalConfig &cfg, QWidget * parent)
 
     propertiesChanged();
 
-    connect(m_term, SIGNAL(finished()), this, SIGNAL(finished()));
-    connect(m_term, SIGNAL(termGetFocus()), this, SLOT(term_termGetFocus()));
-    connect(m_term, SIGNAL(termLostFocus()), this, SLOT(term_termLostFocus()));
-    connect(m_term, &QTermWidget::titleChanged, this, [this] { emit termTitleChanged(m_term->title(), m_term->icon()); });
```
tsujan: Why removing this (and adding `TermWidget::term_termTitleChanged()`)?

tsujan:
> lambdas were converted to member functions

Why?

src/tabwidget.cpp
```diff
@@ -161,8 +161,9 @@ void TabWidget::zoomReset()
     terminalHolder()->currentTerminal()->impl()->zoomReset();
 }
 
-void TabWidget::updateTabIndices()
+void TabWidget::updateTabIndices(int, int)
```
palinek: Adding not used parameters to slot just because the signal's signature is not needed. This:
```diff
diff --git a/src/tabwidget.cpp b/src/tabwidget.cpp
index 0935747..2f0ca15 100644
--- a/src/tabwidget.cpp
+++ b/src/tabwidget.cpp
@@ -161,7 +161,7 @@ void TabWidget::zoomReset()
     terminalHolder()->currentTerminal()->impl()->zoomReset();
 }
 
-void TabWidget::updateTabIndices(int, int)
+void TabWidget::updateTabIndices()
 {
     /* The parameters are not used, needed only to pair with the signal */
     for(int i = 0; i < count(); i++)
diff --git a/src/tabwidget.h b/src/tabwidget.h
index 15720f1..930d4ab 100644
--- a/src/tabwidget.h
+++ b/src/tabwidget.h
@@ -107,7 +107,7 @@ protected:
      */
     bool eventFilter(QObject *obj, QEvent *event);
 protected slots:
-    void updateTabIndices(int from = 0, int to = 0);
+    void updateTabIndices();
     void onTermTitleChanged(QString title, QString icon);
 
 private:
```
gets compiled correctly.

kdave:
> >  lambdas were converted to member functions

> Why?

That one was likely wrong, the lambdas cannot be used if we'd want to do disconnect(). I mixed that up from another patch. Lambdas work with the new syntax fine, sorry for confusion.

tsujan:
> sorry for confusion.

No need to be sorry; I just asked.

src/termwidget.cpp
kdave: Yeah, that's not necessary.

kdave:
Branch updated:
- unused parameters removed
- lambda as slot is untouched

agaida:
GTM

PR #396, opened 2018-02-01
Fix tabbar

tsujan:
Fixes https://github.com/lxde/qterminal/issues/395

The following issues are fixed:

(1) `Properties::Instance()->showCloseTabButton` shouldn't be used with `setTabsClosable()` in the c-tor of `TabWidget` because its value is not defined yet. This fixes the erratic behavior of tab close buttons.

(2) The tab pages shouldn't be deleted at `MainWindow::closeEvent()` -- except for one case -- because `TabWidget::removeTab()` deletes them; the last page is also deleted because it's a child of `TabWidget`.

(3) To make the slected tab text bold, we should give enough space to the tab by making the tabbar font bold and then, making the non-selected tab text normal.

### DIFF
--- a/src/tabstyle.cpp
+++ b/src/tabstyle.cpp
@@ -63,7 +63,7 @@ void TabStyle::drawControl(ControlElement element,
         // Draw the final text
         drawItemText(painter,
                      rect,
-                     Qt::AlignVCenter,
+                     Qt::AlignCenter,
                      tabOption->palette,
                      tabOption->state & State_Enabled,
                      elidedText,
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -549,7 +549,6 @@ void MainWindow::closeEvent(QCloseEvent *ev)
         Properties::Instance()->saveSettings();
         for (int i = consoleTabulator->count(); i > 0; --i) {
             consoleTabulator->removeTab(i - 1);
-            delete consoleTabulator->widget(i-1); //Also delete widget
         }
         ev->accept();
         return;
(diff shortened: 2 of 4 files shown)


tsujan:
Oh, I forgot to explain this:

When `Properties::Instance()->showCloseTabButton` was used with `setTabsClosable()` at the c-tor, since the the former value wasn't defined yet, the tab close buttons had all *double* connections to `tabCloseRequested()` ( see `QTabWidget::setTabsClosable` in the Qt source: `(tabsClosable() == closeable)` was never `true`); hence the erratic behavior.

agaida:
GTM

src/tabstyle.cpp
```diff
@@ -63,7 +63,7 @@ void TabStyle::drawControl(ControlElement element,
         // Draw the final text
         drawItemText(painter,
                      rect,
-                     Qt::AlignVCenter,
+                     Qt::AlignCenter,
```
yan12125: Sorry for being late. Is this intentional? Now texts are aligned horizontally center, which looks strange to me, especially when there are icons.
default

I like the original left-aligned texts more.

tsujan: Yes, it's intentional, because there's an extra space for the text of inactive tabs. Horizontal symmetry is more elegant than than the lack of it; why does it look strange to you? Let me guess: because of Fusion and Breeze ;)

yan12125: I guess I want to get consistency between my terminal and my browser. Here's how Firefox draws a tab:
default
Whether it's more elegant or not, I'm accustomed to it.

tsujan: Terminal and browser?! Qt app and Firefox?

Change it to whatever seems better to you if others agree -- I won't disagree because this app isn't mine.

tsujan: Just to show that such things aren't absolute, this is my browser:

browser

yan12125: Oh, you just reminds me that it's possible to load custom QSS #308. I'll check that out.

tsujan: Wait a minute! Konsole has centered tab texts too. This screenshot is taken with Breeze but it's the same with Fusion:

konsole

It seems the centered tab text is a terminal tradition I respected unaware ;)

PR #399, opened 2018-02-01
A real shortcut editor

tsujan:
Fixes https://github.com/lxde/qterminal/issues/398.

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -21,6 +21,7 @@
 #include <QDebug>
 #include <QStyleFactory>
 #include <QFileDialog>
+#include <QKeySequenceEdit>
 
 #include "propertiesdialog.h"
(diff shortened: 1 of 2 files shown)


kdave:
Works fine. I don't know if removing the shortcut is some standardized way, after some experimenting I figured out that pressing Shift while editing would remove the shortcuts one by one. This could be noted somewhere in the panel.

tsujan:
> I don't know if removing the shortcut is some standardized way

`Alt` removes the whole shortcut.

tsujan:
BTW,

(1) Things like shortcut removal are included in `QKeySequenceEdit`.

(2) There's yet another instance of shortcut input in the Dropdown page, for which I'll add a `QKeySequenceEdit` later -- in another PR.

agaida:
GTM

PR #401, opened 2018-02-09
Support modifying $TERM and use xterm-256color by default

yan12125:
Fixes https://github.com/lxde/qterminal/issues/100
Closes https://github.com/lxde/qtermwidget/issues/12

Also, Properties is application-wise, not window-wise, so I moved it to main.cpp.

A screenshot:
default

I feel the UI design native. Feedbacks are welcomed.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>864</width>
-    <height>728</height>
+    <width>952</width>
+    <height>947</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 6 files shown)


yan12125:
Found an issue - htop's colors are messed with TERM=xterm-256color. xfce4-terminal (vte3) and konsole work fine.

agaida:
@yan12125 i think we should dig a little bit deeper, it _might_ be that we messed this up in the past few weeks, noticed that too three days ago, still investigating

agaida:
@yan12125 - work fine so far when qterminal is restarted - can we set the environment of the open terminals  when the value is changed? Otherwise we should mention the needed restart.

yan12125:
Seems there's no way to change $TERM for existing tabs. Changing $TERM for new tabs in existing windows should be possible. It requires non-trivial work and I may not be able to finish that soon - going to have a one-week vacation :)

agaida:
no problem - will add a new issue - GTM

Issue #409, opened 2018-04-25
Vertical tabs are not usable


aramazano:
Upgraded to qterminal 0.8.1~36-g1a4452a-1 and the tabs have become very large (plus statically sized as before). 

Each tab is like half the width of fullscreen size. Only 2 tabs can fit in the tab bar, 3rd tab overflows it. It gets worse when the tab bar is located on right or left of the window: Tabs keep their enormous widths, but they also get very deep, and neighboring tab titles get superimposed on one another when scrolling through tab-bar downwards.

Before the upgrade (qterminal_0.8.0-4) statically sized tabs were already an inconvenience (opening 5 tabs were enough to overflow the tab bar). Now with gigantic tab sizes the terminal has become barely usable.

I would like to use this opportunity to ask please also make the tabs dynamically sized. With statically sized tabs, it becomes inconvenient to use more than 5 tabs here. (Now it is dropped to 2). I know it is possible scroll through tabs, but it is an inconvenient hassle.

Thank you and regards.

##### System Information
* Distribution & Version:  LXQt experimental-snapshots on top of Debian Testing
* Kernel: 4.9.0-4-amd64
* Qt Version: 5.10.1
* libqtxdg Version: libqt5xdg3 / 3.1.1~13-gb5ddbd4-1
* qtermwidget Version: 0.8.1~24-g4cc05f9-1
* lxqt-build-tools Version: Not installed
* Package version: 0.8.1~36-g1a4452a-1 (About dialog: 0.8.0)

aramazano:
My bad. I did look into the settings but missed that one somehow. Nevertheless, I would suggest making this option unchecked by default.

BTW the other, "opportunistic" request (re dynamic/squeezing tab sizes) still pending. :)

aramazano:
Hmm there is still this issue;

> It gets worse when the tab bar is located on right or left of the window: Tabs keep their enormous widths, but they also get very deep, and neighboring tab titles get superimposed on one another when scrolling through tab-bar downwards.

There are 3 problems with "vertical" tabs.

1) Normally, the width of vertical tab bar should always be slim (like the height of horizoltal bar) and "Limit tab width" should affect its vertical width (depth). Actually it works as;

a) the -vertical- width (depth) of vertical tabs are always *fixed* (apparently equal to the length of horizontal tabs as if "limit tab width" unchecked), and

b) "Limit tab width" affects the horizontal width of vertical tab bar, while it should have affected the depth (vertical width).

[...]

PR #418, opened 2018-05-15
Fix tab texts, especially for vertical tabs

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/409

The custom style of the code was not only wrong but redundant. It was wrong -- like many other hard-coded styles -- because it ignored some aspects of `QStyle`. It was redundant because text eliding could be done in an easier way and style engines know how to draw icons, if any.

With this patch, not only vertical tabs are drawn correctly but also all tabs show the standard behavior: if the width limit is not set, tabs are shrinked to avoid tab scroll buttons as far as possible (previously, a single tab with a long text could make a useless scroll button show up).


agaida:
GTM

PR #422, opened 2018-05-19
Bump version to 0.9.0

agaida:
Bumped minimum widget version to 0.9.0 too


agaida:
no, not so important - i was only a bit irritated and had to look into the code where it is used :)

Issue #450, opened 2018-07-20
QTerminal doesn't remember the maximization state


tsimonq2:
When maximizing QTerminal, closing it, and reopening it, the maximization state is saved, whereas in e.g. PCManFM-Qt, it is saved.

yan12125:
Which window manager do you use? I had the same problem with openbox, and it works fine after I switched to xfwm4.

tsimonq2:
Openbox, but the WM is irrelevant here.

tsujan:
@yan12125 I didn't have time too look into the code but @tsimonq2 may be right. Some window managers may make the window maximized when its size is that of a maximized window. So, QTerminal may have a bug.

yan12125:
Oh, that's a good approach!

PR #451, opened 2018-07-20
Remember the maximization state of the window.

tsimonq2:
#450

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -196,6 +196,8 @@ MainWindow *QTerminalApp::newWindow(bool dropMode, TerminalConfig &cfg)
     else
     {
         window = new MainWindow(cfg, dropMode);
+        if (Properties::Instance()->windowMaximized)
+            window->setWindowState(window->windowState() | Qt::WindowMaximized);
         window->show();
     }
     return window;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -197,7 +197,7 @@ MainWindow *QTerminalApp::newWindow(bool dropMode, TerminalConfig &cfg)
     {
         window = new MainWindow(cfg, dropMode);
         if (Properties::Instance()->windowMaximized)
-            window->setWindowState(window->windowState() | Qt::WindowMaximized);
+            window->setWindowState(Qt::WindowMaximized);
         window->show();
     }
     return window;
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -542,6 +542,7 @@ void MainWindow::closeEvent(QCloseEvent *ev)
             if (Properties::Instance()->saveSizeOnExit) {
             	Properties::Instance()->mainWindowSize = size();
             }
+            Properties::Instance()->windowMaximized = isMaximized();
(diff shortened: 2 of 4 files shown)


tsimonq2:
Note, this is somewhat inspired by [the PR adding this feature](https://github.com/lxqt/pcmanfm-qt/pull/124) in PCManFM-Qt.

tsimonq2:
Tested this on Lubuntu Cosmic with Openbox, LXQt 0.13.0, and Qt 5.9.5. This feature works as intended.

src/main.cpp
```diff
@@ -196,6 +196,8 @@ MainWindow *QTerminalApp::newWindow(bool dropMode, TerminalConfig &cfg)
     else
     {
         window = new MainWindow(cfg, dropMode);
+        if (Properties::Instance()->windowMaximized)
+            window->setWindowState(window->windowState() | Qt::WindowMaximized);
         window->show();
```
tsujan: `setWindowState` takes `Qt::WindowState` as its argument while you've given it `Qt::WindowStates`. Ref. http://doc.qt.io/qt-5/qwindow.html

tsimonq2: So, to make sure I get this right, the proper code looks like this?
```
window->setWindowStates(window->windowState() | Qt::WindowMaximized);
```

tsimonq2: To answer my own question, that would make sense, so I'll change it.

If I'm understanding this right, `Qt::WindowStates` can still take only one argument, so it wouldn't hurt either way to be safe here,

tsujan: My above comment may sound pedantic but future versions of Qt might not accept it. Please test `setWindowState(Qt::WindowMaximized);` it should be enough.

tsimonq2: Except, that's FTBFS...
```
src/main.cpp:200:21: error: ‘class MainWindow’ has no member named ‘setWindowStates’; did you mean ‘setWindowState’?
             window->setWindowStates(window->windowState() | Qt::WindowMaximized);
                     ^~~~~~~~~~~~~~~
                     setWindowState
```

tsimonq2: ACK

tsujan: My suggestion was `setWindowState(Qt::WindowMaximized)`.

tsimonq2:
How's that look, @tsujan?

tsujan:
It should be OK. Please test it too!

Quote from http://doc.qt.io/qt-5/qt.html#WindowState-enum: "The WindowStates type is a typedef for QFlags<WindowState>. It stores an OR combination of WindowState values."

tsujan:
I should change it in pcmanfm-qt too.

tsimonq2:
Just tested it with the same environment listed above, and it works fine.

Since this is a fairly simple patch, I'll merge it.

tsujan:
GTM. Merge tit

tsimonq2:
Thanks @tsujan! :smiley:

tsujan:
You're welcome. My review was just about a tiny change.

PR #453, opened 2018-07-24
Update Czech localization

p-bo:


qterminal_drop.desktop
```diff
@@ -55,6 +56,7 @@ GenericName[zh_TW]=下拉式終端機
 
 Comment=A drop-down terminal emulator.
 Comment[ca]=Un emulador de terminal desplegable.
+Comment[ca]=Vysouvací emulátor terminálu.
```
yan12125: Should be `Comment[cs]`?

p-bo: Yes, definitely – forgot to edit, sorry for that…

PR #454, opened 2018-07-24
Update Czech localization

p-bo:


qterminal.desktop
```diff
@@ -3,9 +3,11 @@ Name=QTerminal
 Type=Application
 GenericName=Terminal emulator
 GenericName[es]=Emulador de terminal
+GenericName[cs]=Emulátor terminálu
```
yan12125: Please sort them alphabetically (`cs` before `es`, for example)

PR #458, opened 2018-08-02
Don't use automatic string conversions

luis-pereira:
* Disables automatic conversions from 8-bit strings (char *) to unicode
  QStrings.
* Disables automatic conversion from QString to 8-bit strings (char *).
* Disables automatic conversions from QByteArray to const char * or const
  void *.
* Disables automatic conversions from QString (or char *) to QUrl.
* Use QStringBuilder for more efficient string creation.

It make us aware of string and encoding conversions.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -83,23 +83,23 @@ void parse_args(int argc, char* argv[], QString& workdir, QString & shell_comman
                 print_usage_and_exit(0);
                 break;
             case 'w':
-                workdir = QString(optarg);
+                workdir = QString::fromLocal8Bit(optarg);
                 break;
             case 'e':
-                shell_command = QString(optarg);
+                shell_command = QString::fromLocal8Bit(optarg);
                 // #15 "Raw" -e params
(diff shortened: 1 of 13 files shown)


FlingeR:
@luis-pereira 
Working on OpenBSD.

src/main.cpp
```diff
@@ -158,14 +158,14 @@ int main(int argc, char *argv[])
...
+        QIcon::setThemeName(QStringLiteral("QTerminal"));
 
     // translations
-    QString fname = QString("qterminal_%1.qm").arg(QLocale::system().name().left(5));
+    QString fname = QString::fromLatin1("qterminal_%1.qm").arg(QLocale::system().name().left(5));
     QTranslator translator;
 #ifdef TRANSLATIONS_DIR
     qDebug() << "TRANSLATIONS_DIR: Loading translation file" << fname << "from dir" << TRANSLATIONS_DIR;
-    qDebug() << "load success:" << translator.load(fname, TRANSLATIONS_DIR, "_");
+    qDebug() << "load success:" << translator.load(fname, QString::fromUtf8(TRANSLATIONS_DIR), QStringLiteral("+"));
```
agaida: @tsujan if i remember en keyboardlayout right it was a typo - missed shift

PR #461, opened 2018-08-13
Added tab switching

zjes:
Another way change the active tab by cycling through the available ones. Use switch tabs history.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ set(QTERM_SRC
     src/bookmarkswidget.cpp
     src/fontdialog.cpp
     src/dbusaddressable.cpp
+    src/tab-switcher.cpp
 )
 
 set(QTERM_MOC_SRC
@@ -76,6 +77,7 @@ set(QTERM_MOC_SRC
     src/propertiesdialog.h
     src/bookmarkswidget.h
     src/fontdialog.h
+    src/tab-switcher.h
 )
 
 
--- a/src/config.h
+++ b/src/config.h
@@ -28,6 +28,8 @@
 #define QUIT "Quit"
 #define PREFERENCES "Preferences..."
 
+#define TAB_PREV_HISTORY "Previous Tab in History"
+#define TAB_NEXT_HISTORY "Next Tab in History"
 #define TAB_NEXT "Next Tab"
 #define TAB_PREV "Previous Tab"
 
@@ -72,8 +74,10 @@
 // ACTIONS
 #define CLEAR_TERMINAL_SHORTCUT        "Ctrl+Shift+X"
 
-#define TAB_PREV_SHORTCUT	       "Ctrl+PgUp|Ctrl+Shift+Tab"
-#define TAB_NEXT_SHORTCUT	       "Ctrl+PgDown|Ctrl+Tab"
+#define TAB_PREV_HISTORY_SHORTCUT  "Ctrl+Tab"
+#define TAB_NEXT_HISTORY_SHORTCUT  "Ctrl+Shift+Tab"
+#define TAB_PREV_SHORTCUT	       "Ctrl+PgUp"
+#define TAB_NEXT_SHORTCUT	       "Ctrl+PgDown"
 #define SUB_BOTTOM_SHORTCUT	       "Alt+Down"
 #define SUB_TOP_SHORTCUT	       "Alt+Up"
 #define SUB_LEFT_SHORTCUT	       "Alt+Left"
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -185,6 +185,7 @@ void MainWindow::setup_Action(const char *name, QAction *action, const char *def
     for (const QString &sequenceString : sequences)
         shortcuts.append(QKeySequence::fromString(sequenceString));
     actions[QLatin1String(name)]->setShortcuts(shortcuts);
+    actions[QLatin1String(name)]->setShortcutContext(Qt::WidgetWithChildrenShortcut);
 
     if (receiver)
     {
@@ -220,6 +221,12 @@ void MainWindow::setup_ActionsMenu_Actions()
 
     setup_Action(TAB_PREV, new QAction(QIcon::fromTheme(QStringLiteral("go-previous")), tr("&Previous Tab"), settingOwner),
                  TAB_PREV_SHORTCUT, consoleTabulator, SLOT(switchToLeft()), menu_Actions, data);
+
+    setup_Action(TAB_PREV_HISTORY, new QAction(tr("&Previous Tab in History"), settingOwner),
+                 TAB_PREV_HISTORY_SHORTCUT, consoleTabulator, SLOT(switchToPrev()), menu_Actions, data);
+
+    setup_Action(TAB_NEXT_HISTORY, new QAction(tr("&Next Tab in History"), settingOwner),
+                 TAB_NEXT_HISTORY_SHORTCUT, consoleTabulator, SLOT(switchToNext()), menu_Actions, data);
 
     setup_Action(MOVE_LEFT, new QAction(tr("Move Tab &Left"), settingOwner),
                  MOVE_LEFT_SHORTCUT, consoleTabulator, SLOT(moveLeft()), menu_Actions, data);
--- a/src/tab-switcher.cpp
+++ b/src/tab-switcher.cpp
@@ -1,0 +1,174 @@
+#include <QApplication>
+#include <QKeyEvent>
+#include <QTimer>
+#include <QStyledItemDelegate>
+
+#include "tab-switcher.h"
+#include "tabwidget.h"
+
+// -----------------------------------------------------------------------------------------------------------
+
+enum class AppRole {
+    Display = Qt::DisplayRole,
+    Index   = Qt::UserRole +1
+};
+
+// -----------------------------------------------------------------------------------------------------------
+
+AppModel::AppModel(QObject* parent, TabWidget* tabs):
+    QAbstractListModel(parent)
+{
+    for(QWidget* w: tabs->history()) {
+        int index = w->property("tab_index").toInt();
+        m_list.append({tabs->tabText(index), index});
+    }
+}
+
+int AppModel::rowCount(const QModelIndex& /*parent*/) const
+{
+    return m_list.size();
+}
+
+QVariant AppModel::data(const QModelIndex &index, int role) const
+{
+    if (!index.isValid() || index.row() >= m_list.size())
+        return QVariant();
+
+    switch(static_cast<AppRole>(role)) {
+    case AppRole::Display:
+        return m_list[index.row()].name;
+    case AppRole::Index:
+        return m_list[index.row()].index;
+    }
+
+    return {};
+}
+
+// -----------------------------------------------------------------------------------------------------------
+
+class AppItemDelegate: public QStyledItemDelegate
+{
+protected:
+    void paint(QPainter * painter, const QStyleOptionViewItem & option, const QModelIndex & index) const override
+    {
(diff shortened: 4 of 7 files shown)


tsujan:
Apart from the conflict your branch has with master, this is a good idea and your implementation works with `Ctrl+Tab`. But, at least, there is one bug: its behavior is wrong when the user presses `Ctrl+Tab` to show the switch list and then Up/Down arrows to select a tab from the list.

zjes:
Ups... fixed it :)
After review all commits will be squashed, rebased to current master etc. (sure, if this patch will be useful :))

tsujan:
Yes, that's fixed now.

Two other problems:

(1) `Ctrl+Shift+Tab` seems to work as expected but `Ctrl+Tab` doesn't; it just selects 2 tabs.

(2) The drawing is wrong; the selected item can have dark text on dark background, depending on the style. Either everything should be left to the active widget style (preferable) or the drawing should be done completely in the code, considering all possible parameters (hard and a waste of time).

tsujan:
> sure, if this patch will be usefu

IMO, it's useful; just its issues should be fixed.

zjes:
About styling. No problems. I just never think about it :(

tsujan:
The style problem should be fixed by something like this:
```c++
    void paint(QPainter * painter, const QStyleOptionViewItem & option, const QModelIndex & index) const override
    {
        QStyle *style = option.widget ? option.widget->style() : QApplication::style();
        style->drawPrimitive(QStyle::PE_PanelItemViewItem, &option, painter, option.widget);

        QString text = index.model()->data(index, static_cast<int>(AppRole::Display)).toString();
        QStyleOptionViewItem opt = option;
        initStyleOption(&opt, index);
        opt.text = text;
        style->drawControl(QStyle::CE_ItemViewItem, &opt, painter, option.widget);
    }
```
Please note: I didn't test it ;)

tsujan:
I added the missing eliding condition to the above code but, IMHO, it's better to set *the terminal width* as the max. text length, instead of using the number of characters.

zjes:
> instead of using the number of characters.

Possible you right here, but I just tried to avoid problems with dpi etc... It's good if you have settings of this width.

> Ctrl+Tab doesn't

Strange, it works for me, will check it letter, maybe something forgot to commit.

tsujan:
> but I just tried to avoid problems with dpi

Theoretically speaking (I haven't tested it though), we get the terminal width before showing the list and elide the text accordingly; so no problem should arise with HDPI -- if I've understood your concern correctly.

> Strange, it works for me

I'll also test it again later.

zjes:
Style was fixed, also was squashed and rebased...
Ctrl+Tab works for me. :(

tsujan:
(1) The style isn't correct yet. All Qt styles elide the text automatically when needed (and if someone makes a style that doesn't, it'll have a bug). Please consider something like this:
```c++
class AppItemDelegate: public QStyledItemDelegate
{
public:
    AppItemDelegate(int frameWidth = 0, QWidget* parent = nullptr) :
        QStyledItemDelegate(parent),
        mParent(parent),
        mFrameWidth(frameWidth) {}
    ~AppItemDelegate() {}

protected:
    void paint(QPainter * painter, const QStyleOptionViewItem & option, const QModelIndex & index) const override
    {
        QStyle *style = option.widget ? option.widget->style() : QApplication::style();

        QString text = index.model()->data(index, static_cast<int>(AppRole::Display)).toString();

        QStyleOptionViewItem opt = option;
        initStyleOption(&opt, index);
        opt.text = text;
        style->drawControl(QStyle::CE_ItemViewItem, &opt, painter, option.widget);
    }

    QSize sizeHint(const QStyleOptionViewItem &option, const QModelIndex &index) const override
    {
        QStyleOptionViewItem opt = option;
        initStyleOption(&opt, index);
        opt.decorationSize = QSize(0, 0);
        opt.text = index.model()->data(index, static_cast<int>(AppRole::Display)).toString();;
        const QWidget* widget = option.widget;
        QStyle* style = widget ? widget->style() : QApplication::style();
        QSize CTSize = style->sizeFromContents(QStyle::CT_ItemViewItem, &opt, QSize(), widget);
        return QSize(mParent ? qMin(mParent->width() - 2 * mFrameWidth, CTSize.width()) : CTSize.width(),
                     CTSize.height());
    }

private:
    QWidget* mParent;
    int mFrameWidth;
};

// -----------------------------------------------------------------------------------------------------------

TabSwitcher::TabSwitcher(TabWidget* tabs):
    QListView(tabs),
    m_tabs(tabs)
{
    setWindowFlags(Qt::Widget | Qt::Popup | Qt::WindowStaysOnTopHint);
    setItemDelegate(new AppItemDelegate(frameWidth(), tabs));
    setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);

    m_timer = new QTimer(this);
    m_timer->setInterval(100);
    m_timer->setSingleShot(true);

    connect(m_timer, &QTimer::timeout, this, &TabSwitcher::timer);
}

TabSwitcher::~TabSwitcher()
{}

void TabSwitcher::showSwitcher()
{
    setModel(new AppModel(this, m_tabs));

    if (!model()->rowCount())
        return;

    QStyleOptionViewItem option;

    int w = 0;
    int h = 0;
    int maxApp = 15;

    for(int i = 0; i < model()->rowCount(); ++i){
        QSize siz = itemDelegate()->sizeHint(option, model()->index(i, 0));
        w = qMax(w, siz.width());
        h += siz.height();
        if (i > maxApp)
            break;
    }

    w += 2 * frameWidth();
    h += 2 * frameWidth();
    resize(w, h);
    QPoint pos = m_tabs->mapToGlobal(m_tabs->geometry().topLeft());
    move(pos.x()+m_tabs->geometry().width()/2 - w / 2, pos.y()+m_tabs->geometry().height()/2 - h / 2);
    show();
}
```
With a code like this, the maximum width is exactly the tab-widget width (which can be changed by the user). The frame width should also be taken into account because a frame is drawn by default in all widget styles unless the frame is removed explicitly in the code (but there's no need to that).

The following screenshot shows a too long path with Fuison -- it's elided by Fusion too:

fusion


(2) `Alt+Tab` doesn't work in the sense that when I open ≥ 3 tabs and press `Alt+Tab`, it covers only two tabs, while `Ctrl+Shift+Tab` covers all ≥ 3 tabs.

zjes:
Alt+Tab? :)
And, maybe suspicion, have you any other key binding to Ctrl+Tab?

tsujan:
> Alt+Tab? :)

No, `Ctrl+Tab` (corrected it in the comment) -- it seems I was too drowsy ;)

> have you any other key binding to Ctrl+Tab?

No, I don't! If I find the time, I'll look into your code to know why that happens.

tsujan:
@yan12125 `Ctrl+Tab` aside, this PR needs fixes, as mentioned and suggested above.

zjes:
@yan12125 

> That would be confusing to existing users.
Ctrl+Tab is std shortcut for window cycling, almost everywhere, this is was a almost main reason why I made this changeset few years ago :)

@tsujan Ups. I forgot to commit fix of the styling. Will try to do it today. :(

tsujan:
@zjes Good! If the problems are fixed and the PR is approved, I'll do some fixes here and there after it's merged. As said above, IMO, this is a good addition.

BTW, I have no problem with `Ctrl+Tab` -- provided that it really works, of course. After all, the shortcuts can be customized.

zjes:
@tsujan  Style was fixed by your suggestion. I still cannot reproduce issue with forward switching.

zjes:
While switching you keep pressed modifier key (Ctrl)?

tsujan:
> Have you keeping pressed modifier key (Ctrl) while switching?

To see the bug, do this again and again with 3 tabs:

(1) Press `Ctrl` and keep it pressed;
(2) Press `Tab` once and release it immediately after that (you'll see the switcher but its middle item is selected); and
(3) Then, release `Ctrl`.

If you do the same thing with `Ctrl+Shift+Tab`, there will be no problem.

yan12125:
> Ctrl+Tab is std shortcut for window cycling, almost everywhere

Any examples? Besides QTerminal, I know Firefox only, where tabs are switched by the apparent order (the leftmost to the rightmost) instead of the last time used.

tsujan:
@yan12125 If, previously, `Ctrl+Tab` was for tab switching, that isn't changed; an interesting and useful switcher is shown in addition. That's why I like @zjes's work.

tsujan:
Two other examples: Chromium and Falkon.

zjes:
@tsujan Try do not release modifier key. Just repeat step 2. This shortcut have little bit different behaviour, means that he is trigger on modifier release event. 

@yan12125 Additionaly if you familiar with any development tools such as QtCreator, Eclipse, MSVC etc, all of them have such cycling. 

This is very easy: Alt-Tab to switch between applictions, Ctrl-Tab inside application with the same behaviour. Means press Ctrl, after that press one or many Tab to switch to application/tab (press +shift cycling to opposite direction). If you done, just release all keys.

tsujan:
> @tsujan Try do not release modifier key. 

Sorry but this isn't the way a program is developed. We can't tell users not to do so and so; instead, we should fix our codes.

tsujan:
I know what you say but I'm not sure it's acceptable with Qterminal. Yes, Qt Creator, for example, behaves in the same way but not Firefox, Chromium or Falkon. If we want to preserve `Ctrl+Tab` for tab switching (as I promised @yan12125), we might need to change this behavior. In this sense, I agree with @yan12125 that this would be confusing -- @yan12125 may had seen what I see only now.

I'll wait for other LXQt members to tell about their opinions. Please don't change your code until we hear from them.

tsujan:
Oh, I forgot to say that the behavior of `Alt+Tab` (if this idea came from it) depends on what the user chooses; at least in KDE, it's either "recently used" or "stacking order". I remember correctly, it was the second in Compiz.

zjes:
This is up to you :)
For me such "recently used" more useful and natural then "stacking order". What about to add this option to qterminal setting? If it on, then on Ctrl+Tab use this switcher, otherwise keep current behaviour?

yan12125:
> I know what you say but I'm not sure it's acceptable with Qterminal. Yes, Qt Creator, for example, behaves in the same way but not Firefox, Chromium or Falkon. If we want to preserve Ctrl+Tab for tab switching (as I promised @yan12125), we might need to change this behavior. In this sense, I agree with @yan12125 that this would be confusing -- @yan12125 may had seen what I see only now.

That's my point. Sorry for not explaining things clearly!

And using options would be great.

zjes:
What about to add a option first? :)

tsujan:
@yan12125 That was my fault. I don't use Qt Creator on a regular basis but @zjes was right: this is the behavior of Qt Creator's switcher. IMHO, it'll be a nice addition to QTerminal.

Actually, the idea of reactivating the last active tab seemed so good to me that I'm adding it to FeatherPad but without any switcher and with `F1` as the default shortcut.

tsujan:
> What about to add a option first? :)

I agree -- if you have time to do it here.

zjes:
Will do, possible today evening. :)
Not good to merge without this option.

zjes:
A problem, big problem. I don't know how to name this option. :)
[x] Use "recently used" tabs switcher
or something else?
Do you have any suggestions?

tsujan:
Something like this, IMO:

* [x] Tabs switcher selects last active tab

Or its opposite:

* [ ] Tabs switcher goes through all tabs

zjes:
"Tabs switcher selects last active tab" sounds good, but allow navigate in history too. :)

tsujan:
> but this option allows navigate in history too

I think users will find its other properties themselves. Its main property is related to `Ctrl+Tab`; isn't it? `Tab` can call Backtab (Shift+Tab) to mind.

PR #464, opened 2018-08-16
Refactored Limit/limit -> Fixed/fixed tabwitdh

### LINKED ISSUE #411: Interpretation of "Limit tab width:" setting
Currently the aforementioned setting actually works as "fixed tab width". For instance, when set to 500px, then even if I rename the session as "i" (minimal horizontal space needed) the tab is still 500 pixels in length.

Instead, I would suggest a behavior like this:

There is no enable/disable option (always enabled), but just a limit setting (always effective).

##### With top/bottom horizontal tab bar:

It places a LOWER limit to dynamically sized tabs *unless* there is free space in the tab bar. I.e. if there is free space on the tab bar, then the tabs occupy whatever space they need. If there is space shortage in the tab bar, then it dynamically squeezes the tabs until the limit set with this setting is reached. So, all the tabs are guaranteed to fit in the tab bar as long as space and this setting permits. Tab bar overflowing (scrolling) only occurs after hitting this lower limit. IOW, it behaves the other way around: For dynamically sized tabs, instead of setting the upper bound, it sets the lower bound.

##### With left/right vertical tab bar (or list view):

Re: https://github.com/lxqt/qterminal/issues/410

It places an UPPER(*) limit to dynamically sized tabs.

(*) In order to have a stable (not varying) terminal size, this can alternatively be a FIXED limit for vertical tab bar. Or, an option can be introduced to specify whether to make it fixed or adaptive.

E.g.

[*] Adaptive/Fixed tab width

Limit tab width to: [___ px]

Note that these two settings are independent of each other. That is, checking or unchecking "Adaptive width" doesn't disable "Limit tab width" setting.

PS: I didn't like the idea of introducing another option (Adaptive width). Mentioned here for the sake of completeness. I would rather interpret "Limit tab width" simply as fixed size for vertical tab bar.

Thanks and regards.

##### System Information
* Package version: 0.8.1~36-g1a4452a-1

agaida:
fixes #411

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -380,14 +380,14 @@
         </widget>
        </item>
        <item row="9" column="0">
-        <widget class="QCheckBox" name="limitTabWidthCheckBox">
+        <widget class="QCheckBox" name="fixedTabWidthCheckBox">
          <property name="text">
-          <string>Limit tab width:</string>
+          <string>Fixed tab width:</string>
          </property>
         </widget>
        </item>
        <item row="9" column="1">
-        <widget class="QSpinBox" name="limitTabWidthSpinBox">
+        <widget class="QSpinBox" name="fixedTabWidthSpinBox">
          <property name="enabled">
           <bool>false</bool>
          </property>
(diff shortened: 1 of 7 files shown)


agaida:
Second PR, when accepted - updating the translation files

PR #475, opened 2018-09-02
Implement margins

Admicos:
Depends on lxqt/QTermWidget#205

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -84,7 +84,104 @@
      </property>
      <widget class="QWidget" name="appearancePage">
       <layout class="QGridLayout" name="gridLayout_3">
-       <item row="18" column="1">
+       <item row="8" column="0">
+        <widget class="QCheckBox" name="hideTabBarCheckBox">
+         <property name="text">
+          <string>Hide tab bar with only one tab</string>
+         </property>
+        </widget>
+       </item>
+       <item row="2" column="1">
+        <widget class="QComboBox" name="colorSchemaCombo"/>
+       </item>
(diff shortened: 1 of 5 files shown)


src/properties.h
```diff
@@ -59,6 +59,7 @@ class Properties
 
         Sessions sessions;
 
+        int margin;
```
yan12125: Better to name this `terminalMargin`.

src/forms/propertiesdialog.ui
```diff
@@ -409,6 +409,23 @@
          </property>
         </widget>
        </item>
+       <item row="17" column="0">
+        <widget class="QLabel" name="label_15">
+         <property name="text">
+          <string>Terminal margin</string>
+         </property>
+         <property name="buddy">
+          <cstring>marginSpinBox</cstring>
```
yan12125: Ditto. `terminalMarginSpinBox` is better.

src/forms/propertiesdialog.ui
```diff
@@ -84,104 +84,7 @@
      </property>
      <widget class="QWidget" name="appearancePage">
       <layout class="QGridLayout" name="gridLayout_3">
-       <item row="8" column="0">
```
yan12125: Could you keep unrelated controls untouched? I know Qt Designer often messed up UI files, so you may need to do that manually.

Admicos:
All of those should be fixed now. Moved the terminal margin setting to the bottom to keep my edits minimal (not that familiar with Qt stuff, that might be unneccesary)

yan12125:
Thanks for efforts and the quick response! Merged.

PR #495, opened 2018-10-14
Improved cmake_minimum_required

agaida:
Removed CMAKE_BUILD_TYPE
Qt5DBus is required even in OSX


yan12125:
> Qt5DBus is required even in OSX

Then I think `if(APPLE)` should be removed

PR #518, opened 2018-11-29
CHANGELOG: Minor improvements

yan12125:
Ref: https://github.com/lxqt/lxqt/issues/1632


agaida:
@yan12125 - that was one reason for the changelog branches - just push needed changes next time

PR #519, opened 2018-12-10
Update from Weblate.

HelixBot:


src/translations/qterminal_gl.ts
```diff
@@ -555,7 +555,7 @@
     <message>
         <location filename="../forms/propertiesdialog.ui" line="577"/>
         <source>xterm-256color</source>
-        <translation></translation>
+        <translation>xterm en 256 cores</translation>
```
yan12125: @mbouzada This is a special term and should not be translated.

Notes to myself: we should prevent such fields from appearing in `*.ts` files.

agaida:
@yan12125 - we should do this first hand, because every translation update will bring this back if we don't do so.

yan12125:
> we should do this first hand, because every translation update will bring this back if we don't do so.

Yep and here's my attempt: #520

PR #520, opened 2018-12-10
Don't translate xterm and xterm-256color

yan12125:
These are special terms and should not be translated like https://github.com/lxqt/qterminal/pull/519.

I think `*.ts` files should be updated and uploaded to WebLate? Not sure how to do that.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -586,12 +586,12 @@
             </property>
             <item>
              <property name="text">
-              <string>xterm</string>
+              <string notr="true">xterm</string>
              </property>
             </item>
             <item>
              <property name="text">
-              <string>xterm-256color</string>
+              <string notr="true">xterm-256color</string>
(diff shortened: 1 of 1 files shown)


tsujan:
GTM

agaida:
GTM - @yan12125 - didn't read it right, was busy with our lovely discourse forum that don't liked a restart of the VM - ruby and version update needed - i read the file as the basic *ts-file, but `notr` looks good - we will see it today when i prepare the translation updates.

agaida:
https://github.com/lxqt/qterminal/pull/521/files#diff-2ef53241b215c3ffbbc2fe0e1a5ffd80L557

cool, seems to work fine

agaida:
lxqt/qterminal/pull/522/commits/7144c0c6af67285c3bd12de0fa85aca4762b18fa#diff-76eac8adb43f0c0b9107727f2b0029abL557 - better

PR #530, opened 2018-12-30
Update from Weblate.

HelixBot:


src/translations/qterminal_tr.ts
```diff
@@ -93,7 +93,7 @@
     <message>
         <location filename="../mainwindow.cpp" line="247"/>
         <source>&amp;Collapse Subterminal</source>
-        <translation>&amp;Subterminal Kapa</translation>
+        <translation></translation>
```
yan12125: @ayhanyalcinsoy why remove it?

agaida:
@yan12125 - most likely a problem with deleting the old string and not put in the new one  -  i guess that such comments are better placed in weblate.

yan12125:
> i guess that such comments are better placed in weblate.

Thanks for the comment! Didn't know weblate has such a feature.

PR #532, opened 2018-12-31
Fix text string

Atalanttore:


yan12125:
There's another "Keyboard Cursor Shape" in the `View` menu.

And I would say simply "Cursor shape" is fine.

yan12125:
@shlyakpavel We are on the same line :)

agaida:
@yan12125 @shlyakpavel  - better? Btw, @yan12125 - you should be able to change PRs directly

yan12125:
Thanks for the update! First I was hesitated without a comment from the original author, then I forgot about this PR lol

Issue #533, opened 2019-01-05
Revive qterminal on macOS


yan12125:
##### Expected Behavior
qterminal works fine on macOS

##### Current Behavior
It doesn't even build

##### Possible Solution
Land macOS-specific patches from https://gitlab.com/yan12125/macports-overlay/commit/f266c2d0a8d6813a25e0675a6ba5c2b4b0276086

##### Steps to Reproduce (for bugs)
1. Try to build qtermwidget and qterminal on macOS Mojave

##### Context
A requirement of https://github.com/lxqt/qtermwidget/issues/240

##### System Information
* Distribution & Version: macOS Mojave 10.14.2
* Kernel: xnu-4903.231.4~2/RELEASE_X86_64
* Qt Version: 5.12.0
* qtermwidget Version: git-master
* lxqt-build-tools Version: 0.5.0
* Package version: git-master

apjanke:
I've got qterminal building on macOS 10.13.6 against Qt 5.12, using Homebrew.

https://github.com/octave-app/homebrew-octave-app-bases/blob/bcb4c5a2b7eb6158bbbcee52a575cf9ed09a5bec/Formula/qterminal.rb
https://github.com/octave-app/homebrew-octave-app-bases/blob/bcb4c5a2b7eb6158bbbcee52a575cf9ed09a5bec/Formula/qtermwidget.rb
https://github.com/octave-app/homebrew-octave-app-bases/blob/bcb4c5a2b7eb6158bbbcee52a575cf9ed09a5bec/Formula/lxqt-build-tools.rb

```
brew tap octave-app/octave-app-bases
brew install qterminal
```

It's not working, though. I can launch it and the terminal comes up, but it doesn't display any text in the terminal or respond to keystrokes.

```
$ qterminal
Properties constructor called
TRANSLATIONS_DIR: Loading translation file "qterminal_en_US.qm" from dir /usr/local/Cellar/qterminal/0.9.0/share/qterminal/translations
load success: false
Canot open file "/Users/janke/.config/qterminal.org/qterminal_bookmarks.xml"
Trying to load translation file from dir "/usr/local/share"
Trying to load translation file from dir "/usr/share"
Trying to load translation file from dir "/usr/local/Cellar/qtermwidget/0.9.0/share/qtermwidget5/translations"
Translations found in "/usr/local/Cellar/qtermwidget/0.9.0/share/qtermwidget5/translations"
Can't open a pseudo teletype
[...]

PR #560, opened 2019-03-05
CMakeLists: Do not use Qt5X11Extras on Mac

apjanke:
Blocks #533.

Currently, when building against the common Homebrewed Qt on macOS, the `cmake` step fails with the following error.

```
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:28 (find_package):
  By not providing "FindQt5X11Extras.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "Qt5X11Extras", but CMake did not find one.

  Could not find a package configuration file provided by "Qt5X11Extras"
  (requested version 5.7.1) with any of the following names:

    Qt5X11ExtrasConfig.cmake
    qt5x11extras-config.cmake

  Add the installation prefix of "Qt5X11Extras" to CMAKE_PREFIX_PATH or set
  "Qt5X11Extras_DIR" to a directory containing one of the above files.  If
  "Qt5X11Extras" provides a separate development package or SDK, be sure it
  has been installed.


-- Configuring incomplete, errors occurred!
See also "/Users/janke/local/repos/qterminal-BUILD/CMakeFiles/CMakeOutput.log".
```

The common Homebrewed Qt installation does not include Qt5X11Extras because it does not include any X11 support. X11 is a legacy technology on Mac, and it should not be assumed to exist there.

This PR removes the Qt5X11Extras dependency on Mac. With this change, QTerminal will build on macOS again.


yan12125:
Thank you!

PR #563, opened 2019-03-05
Fix default font on macOS

apjanke:
There is no "Monospace" font on macOS. Attempting to use it will cause you to fall back to a variable-width font instead, which is not appropriate for terminals.

This PR changes `config.h` to use "Monaco" on macOS.

Because fonts are saved in the QTerminal settings file and persist between runs, you may need to `rm -rf ~/.config/qterminal.org` to get this to take effect.

Follows up https://github.com/lxqt/qterminal/pull/562.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -69,7 +69,12 @@
 
 #define DEFAULT_WIDTH                  800
 #define DEFAULT_HEIGHT                 600
+#ifdef Q_OS_MACOS
+// Qt does not support fontconfig on macOS, so we need to use a "real" font name.
+#define DEFAULT_FONT                   "Menlo"
+#else
 #define DEFAULT_FONT                   "Monospace"
+#endif
 
 // ACTIONS
 #define CLEAR_TERMINAL_SHORTCUT        "Ctrl+Shift+X"


yan12125:
I wonder if Qt on macOS uses fontconfig or not. If so "Monospace" should work.

By the way, in my memory QTerminal uses a monospace font with my patch on GitLab. I'm using Qt from MacPorts, though.

apjanke:
> I wonder if Qt on macOS uses fontconfig or not. If so "Monospace" should work.

I don't think so. At least, it's not guaranteed. The Homebrew Qt does not have a dependency on fontconfig, and I don't see them interacting. Octave uses fontconfig, but it declares that dependency directly, and sets up fontconfig as part of its own installation process. So I don't think you can rely on fontconfig being present unless you do additional work/take additional dependencies to ensure so.

I don't know how MacPorts's Qt is set up. But I think since different users use different installations of Qt (there's even a Qt binary distribution, though I think its use in practice by end users is rare), QTerminal should support both cases.

yan12125:
It's a bad news that Qt does not work well with fontconfig on macOS - that affects portability of Qt and forces developers to write platform-specific code. I'll check why monospace does not work on macOS even with fontconfig and later ask Homebrew developers to consider enabling it.

For now, it's fine to add this workaround. Is Monaco a built-in font on macOS?

Also, please add a comment for why this workaround is necessary.

apjanke:
> Is Monaco a built-in font on macOS?

Yes. Monaco is built-in on macOS. Menlo is another built-in alternative commonly used for terminals.

> Also, please add a comment for why this workaround is necessary.

Do you want the comment in the source code or the Git commit message?

apjanke:
According to https://bugreports.qt.io/browse/QTBUG-42839 and 
https://github.com/Homebrew/homebrew-core/issues/8502, Qt does not support fontconfig on macOS. It doesn’t look like a Homebrew-specific choice.

So this workaround might be here to stay if we want to support Mac. Unfortunate.

Maybe we should post an enhancement request on Qt’s Issue tracker.

apjanke:
I amended the commit to include a comment in the source code about why the workaround is needed.

Also switched from Monaco to Menlo as the default font: Menlo is also a built-in font on macOS, and it has been the default Terminal and Xcode font since OS X 10.6.

yan12125:
Alright. Thank you very much for investigating the fontconfig issue!

apjanke:
`qterminal` is still issuing a warning about using a variable-width font even though it's starting up with a fixed-width default font now.

```
$ qterminal                                                                                                                          master
Cannot connect to the D-Bus session bus.
[...]
Using a variable-width font in the terminal.  This may cause performance degradation and display/alignment errors.
[...]
```

Looks like there's one more place it's hardcoded to `"Monospace"` over in QTermWidget: https://github.com/lxqt/qtermwidget/blob/e292763e6d23fd92ac30cad9ac107709b8c47c71/lib/qtermwidget.cpp#L324. That should probably be given the same workaround as here, so QTermWidgets embedded in other apps still get a fixed-width default font, and to avoid that spurious warning here.

Issue #574, opened 2019-03-15
terminal settings window is not scrollable


librewish:
##### Expected Behavior

qterminal setting dialog/window/dialog should be scrollable
##### Current Behavior

when my font size is  11 and on 15.6 inch laptop screen with 1366x768 resolution
qterminal setting bottom part is not visible as it goes below the screen and no way to see it as scrolling is not implemented.
##### Possible Solution

implement scrolling in qterminal setting window/dialog
##### Steps to Reproduce (for bugs)

1. lxqt with kwin as window manager
2. 15.6 inch laptop screen 1366x768 resolution.
3. top panel.
4. noto sans 11 font.
5.qterminal >>> preferences dialog bottom part is not visible like applying setting

##### System Information

* Distribution & Version: manjaro lxqt-kwin
* Kernel: 4.19
* Qt Version: 5.12.1
* qtermwidget Version:0.14.1
* lxqt-build-tools Version:
* Package version: 0.14.1

tsujan:
@librewish If you compile from source, please try this branch: https://github.com/lxqt/qterminal/tree/compact_settings

You could also test it without installation like:
```
mkdir build && cd build
cmake ..
make
./qterminal
```
The reason why I asked is that your screen is small.

librewish:
yes it is scrollable and compact.
it fits my screen.

tsujan:
Thanks for testing! I wanted to know whether the dialog would fit in the screen with your resolution and font size. Good to know that the answer is "yes" :)

mpmc:
So glad this issue is here, I had exactly the same problem! Can't wait for the fix to be in master!

yan12125:
Fixed thanks to @tsujan!

PR #575, opened 2019-03-15
Scrollable and compact Settings dialog

tsujan:
Closes https://github.com/lxqt/qterminal/issues/574

If the diff is seen inside a visual diff tool (like KDiff3), it'll be obvious that only a `QScrollArea` is added; the rest is just about indentation.

Also, it's tried to limit the dialog size to available desktop geometry. If the screen is too small, the user could resize the dialog by pressing `Alt` because the scroll area can be made smaller.

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -21,6 +21,8 @@
 #include <QDebug>
 #include <QStyleFactory>
 #include <QFileDialog>
+#include <QScreen>
+#include <QWindow>
 
 #include "propertiesdialog.h"
 #include "properties.h"
@@ -188,7 +190,17 @@ PropertiesDialog::PropertiesDialog(QWidget *parent)
     trimPastedTrailingNewlinesCheckBox->setChecked(Properties::Instance()->trimPastedTrailingNewlines);
     confirmMultilinePasteCheckBox->setChecked(Properties::Instance()->confirmMultilinePaste);
 
-    resize(sizeHint()); // show it compact but not too much
+    // show it compact inside available desktop geometry
+    QSize ag;
+    if (parent != nullptr)
+    {
+        if (QWindow *win = parent->windowHandle())
+        {
+            if (QScreen *sc = win->screen())
+                ag = sc->availableVirtualGeometry().size();
+        }
+    }
+    resize(sizeHint().boundedTo(ag));
 }
 
 
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>952</width>
-    <height>947</height>
+    <width>700</width>
+    <height>700</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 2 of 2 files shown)


tsujan:
BTW, I didn't consider the title-bar height here because it couldn't be found in a reliable way for all window managers. I could have made room for 50 px but, IMO, it's unimportant.

src/propertiesdialog.cpp
```diff
@@ -188,7 +190,17 @@ PropertiesDialog::PropertiesDialog(QWidget *parent)
...
+    QSize ag;
+    if (parent != nullptr)
+    {
+        if (QWindow *win = parent->windowHandle())
+        {
+            if (QScreen *sc = win->screen())
+                ag = sc->availableVirtualGeometry().size();
+        }
+    }
+    resize(sizeHint().boundedTo(ag));
```
yan12125: What will happen if ag can't be retrieved? (e.g., `parent`, `win` or `sc` is `nullptr`)

tsujan: > What will happen if ag can't be retrieved?

Nothing! Because the window cannot be made smaller than its size hint.

Anyhow, `ag` can always be retrieved. Here, the above conditions make sense only logically, not practically.

tsujan: The following explanation is only for the sake of completeness:

Conditions like the above ones are necessary for windows/widgets that can appear on logging (like Panel) in or may change size on logging out (like desktop). It's fascinating how unusual the screen may be in those moments ;)

That's not the case with modal dialogs because their parents are already "formed" completely.

tsujan:
> X11, Wayland and Quartz...

So thorough! I didn't know that QTerminal worked fine on Wayland, let alone Quartz.

BTW, I meant minimum size hint by "size hint" (my bad): a window cannot be made smaller than its min. size.

yan12125:
Thanks to all awesome qterm\* contributors ;-)

tsujan:
@yan12125 That wouldn't be possible without your active participation in developing qterminal. Whenever I update it, I see some interesting fixes and additions. As I see it, you and other active devs rescued qterminal.

yan12125:
Thank you! BTW, you're also one of active devs :)

yan12125:
Looks like agaida is too busy to have a look. Given your expertise in Qt, I think it's fine to just merge it .

tsujan:
> Looks like agaida is too busy to have a look.

One man and too many PRs. With qterminal, your approval would be enough.

yan12125:
Challenge accepted!

tsujan:
> Challenge accepted!

:)

Bigger challenge:
And if you find time, some other projects too. For example, LXQt config (you added support for touchpad and mouse to it).

yan12125:
Good reminder :)

agaida:
@yan12125 and @tsujan - not the one man problem right now - but i cant work so fast as i want to. Getting upstream patches into debian before the upcoming release need patience and time.

tsujan:
> Getting upstream patches into debian before the upcoming release need patience and time.

Which means one man and so many jobs.

There's no rush here — we don't want to have a new release of LXQt so soon.

PR #580, opened 2019-03-26
Implement an action that allows you to send the history to your editor

Andersbakken:
This relies on https://github.com/lxqt/qtermwidget/pull/265 landing in
qtermwidget to work.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -320,6 +320,9 @@ void MainWindow::setup_ActionsMenu_Actions()
     setup_Action(FIND, new QAction(QIcon::fromTheme(QStringLiteral("edit-find")), tr("&Find..."), settingOwner),
                  FIND_SHORTCUT, this, SLOT(find()), menu_Actions);
 
+    setup_Action(HANDLE_HISTORY, new QAction(QIcon::fromTheme(QStringLiteral("handle-history")), tr("Handle history..."), settingOwner),
+                 NULL, this, SLOT(handleHistory()), menu_Actions);
+
 #if 0
     act = new QAction(this);
     act->setSeparator(true);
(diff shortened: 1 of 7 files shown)


src/mainwindow.cpp
```diff
@@ -697,6 +700,37 @@ void MainWindow::find()
     consoleTabulator->terminalHolder()->currentTerminal()->impl()->toggleShowSearchBar();
 }
 
+void MainWindow::sendHistoryToEditor()
+{
+    const QString fn = QStandardPaths::writableLocation(QStandardPaths::TempLocation) + QLatin1String("/qterminal.history.") + QString::number(QCoreApplication::applicationPid());
```
yan12125: The first issue - /tmp is often world-readable, while terminal outputs may contain sensitive information. I'd suggest not to save the history here.

src/mainwindow.cpp
```diff
@@ -697,6 +700,37 @@ void MainWindow::find()
...
+        return;
+    }
+    TermWidgetImpl *impl = consoleTabulator->terminalHolder()->currentTerminal()->impl();
+    impl->saveHistory(&file);
+    file.close();
+    Properties::EditorType type = Properties::Instance()->editorType;
+    QString editor = Properties::Instance()->editor;
+    if (type == Properties::Editor_Auto) {
+        type = Properties::Editor_GUI;
+        for (const char *ed : { "vim", "vi", "less", "more", "nano", "pico", "view", "ed" }) {
```
yan12125: I prefer to avoid another collection of terminal editors in qterminal. This should be done by query the system (e.g., the Terminal= key in XDG standards).

src/mainwindow.cpp
```diff
@@ -697,6 +700,37 @@ void MainWindow::find()
...
+            if (editor == QLatin1String(ed)) {
+                type = Properties::Editor_Terminal;
+                break;
+            }
+        }
+    }
+    if (type == Properties::Editor_Terminal) {
+        impl->sendText(QLatin1String("\1#\n")); // comment out current line first
+        impl->sendText(Properties::Instance()->editor + QLatin1Char(' ') + fn + QLatin1Char('\n'));
```
yan12125: Currently it's unlikely that filenames contain spaces or quotes. However, if we place the history file to somewhere else, the possibility will be higher. Such cases should be handled.

src/mainwindow.cpp
```diff
@@ -697,6 +700,37 @@ void MainWindow::find()
...
+    } else {
+        if (!QProcess::startDetached(Properties::Instance()->editor, QStringList() << fn)) {
```
yan12125: EDITOR/VISUAL can contain spaces. For example `EDITOR='emacs -nw'`.

src/mainwindow.cpp
yan12125: Also, the current process may not be a shell (e.g., `qterminal -e python`). I think it's possible to query the process owning the current PTY - lots of platform-dependent code needed, though.

src/mainwindow.cpp
```diff
@@ -752,6 +755,31 @@ void MainWindow::find()
...
+    if (!file.open(QIODevice::WriteOnly)) {
+        qDebug() << "Failed to open" << file.fileName() << "for writing";
+        return;
+    }
+    TermWidgetImpl *impl = consoleTabulator->terminalHolder()->currentTerminal()->impl();
+    impl->saveHistory(&file);
+    file.close();
+    QStringList args = Properties::Instance()->handleHistoryCommand.split(QLatin1Char(' '), QString::SkipEmptyParts);
+    if (args.isEmpty())
+        return;
```
yan12125: It should be better to have a message If handleHistoryCommand is not set, so that users who want to open files manually can get the filename.

yan12125:
By the way, the action name "Handle history" may not be intuitive. It might be better to call the action "Save history to a file" and rename "Handle history command" in preferences to "Command for saved history file".

PR #581, opened 2019-03-26
Create actions for switching to tab by index.

Andersbakken:


src/mainwindow.cpp
```diff
@@ -234,6 +236,30 @@ void MainWindow::setup_ActionsMenu_Actions()
     setup_Action(MOVE_RIGHT, new QAction(tr("Move Tab &Right"), settingOwner),
                  MOVE_RIGHT_SHORTCUT, consoleTabulator, SLOT(moveRight()), menu_Actions, data);
 
+    menu_Actions->addSeparator();
+    data.setValue(checkHasIndexedTab);
+
+    setup_Action(TAB_1, new QAction(QIcon::fromTheme(QStringLiteral("go-1")), tr("Tab &1"), settingOwner),
```
yan12125: Some comments:

* Is it possible to use a loop instead of 10 function calls?
* Is there an example icon theme provides `go-*` icons?

src/tabwidget.cpp
```diff
@@ -352,26 +353,66 @@ void TabWidget::removeCurrentTab()
...
+void TabWidget::switchTo9()
+{
+    switchTab(8);
+}
+
+void TabWidget::switchTo10()
+{
+    switchTab(9);
+}
+
```
yan12125: I guess these function are not necessary if 10 setup_Action calls are transformed into a loop.

src/mainwindow.cpp
Andersbakken: Yeah. I switched it to using QSignalMapper instead. This is indeed much nicer :-)

Not entirely sure about the icons. It seems a little weird to have an icon for a number. That kinda is your font, no? :-)

yan12125: Got it! Just FYI - QAction can be constructed without an icon ;-)

Andersbakken: Indeed. I removed the icon reference.

src/mainwindow.cpp
```diff
@@ -99,12 +99,12 @@ MainWindow::MainWindow(TerminalConfig &cfg,
         setStyleSheet(QStringLiteral(QSS_DROP));
     }
     else {
-	if (Properties::Instance()->saveSizeOnExit) {
-	    resize(Properties::Instance()->mainWindowSize);
-	}
-	if (Properties::Instance()->savePosOnExit) {
-	    move(Properties::Instance()->mainWindowPosition);
-	}
+    if (Properties::Instance()->saveSizeOnExit) {
```
yan12125: Indentation is messed up here.

src/tabwidget.cpp
```diff
@@ -244,7 +244,7 @@ void TabWidget::contextMenuEvent(QContextMenuEvent *event)
     } else if (action == rename) {
         emit tabRenameRequested(tabIndex);
     } else if (action == changeColor) {
-	emit tabTitleColorChangeRequested(tabIndex);
+    emit tabTitleColorChangeRequested(tabIndex);
```
yan12125: Indentation messed up, too.

src/mainwindow.cpp
```diff
@@ -234,6 +236,23 @@ void MainWindow::setup_ActionsMenu_Actions()
     setup_Action(MOVE_RIGHT, new QAction(tr("Move Tab &Right"), settingOwner),
                  MOVE_RIGHT_SHORTCUT, consoleTabulator, SLOT(moveRight()), menu_Actions, data);
 
+    menu_Actions->addSeparator();
+    data.setValue(checkHasIndexedTab);
+
+    const QString textBase = tr("Tab");
+    QSignalMapper *signalMapper = new QSignalMapper(this);
+    QObject::connect(signalMapper, SIGNAL(mapped(int)), consoleTabulator, SLOT(switchTab(int)));
+    QObject::connect(signalMapper, SIGNAL(mapped(int)), this, SLOT(gotshit(int)));
```
yan12125: Is it possible to use the new Qt5 syntax for `connect()` instead of the SIGNAL-SLOT syntax?

src/mainwindow.cpp
```diff
@@ -234,6 +236,23 @@ void MainWindow::setup_ActionsMenu_Actions()
...
+    for (int i=1; i<=10; ++i) {
+        QString num = QString::number(i);
+        QAction *action = new QAction(textBase + QLatin1Char(' ') + num, settingOwner);
+        action->setProperty("tab", i);
+        char name[16];
+        snprintf(name, sizeof(name), "Tab %d", i);
+        signalMapper->setMapping(action, i - 1);
+        setup_Action(name, action, NULL, signalMapper, SLOT(map()), menu_Actions, data);
```
yan12125: Could you move `Tab X` actions to a submenu (e.g., a submenu `Go to`). After this patch the actions menu appears bloated.

src/mainwindow.cpp
```diff
@@ -172,8 +172,8 @@ void MainWindow::setDropShortcut(QKeySequence dropShortCut)
     }
 }
 
-void MainWindow::setup_Action(const char *name, QAction *action, const char *defaultShortcut, const QObject *receiver,
-                              const char *slot, QMenu *menu, const QVariant &data)
+QAction *MainWindow::setup_Action(const char *name, QAction *action, const char *defaultShortcut, const QObject *receiver,
```
yan12125: Is there a reason to return the action here? Looks like no one uses the return value.

src/mainwindow.cpp
```diff
@@ -234,6 +236,23 @@ void MainWindow::setup_ActionsMenu_Actions()
     setup_Action(MOVE_RIGHT, new QAction(tr("Move Tab &Right"), settingOwner),
                  MOVE_RIGHT_SHORTCUT, consoleTabulator, SLOT(moveRight()), menu_Actions, data);
 
+    menu_Actions->addSeparator();
+    data.setValue(checkHasIndexedTab);
+
+    const QString textBase = tr("Tab");
+    QSignalMapper *signalMapper = new QSignalMapper(this);
```
yan12125: From https://doc.qt.io/qt-5/qsignalmapper.html:

> This class is obsolete. It is provided to keep old source code working. We strongly advise against using it in new code.

src/tabwidget.cpp
Andersbakken: I'll add a commit to reindent the whole file.

src/mainwindow.cpp
Andersbakken: I've removed the signalMapper since it turns out it's been deprecated.

src/mainwindow.cpp
Andersbakken: Didn't realize it had been deprecated. I dropped it in the last patch.

src/mainwindow.cpp
Andersbakken: I used it in an earlier iteration. Forgot to remove it.

src/mainwindow.cpp
Andersbakken: Done.

yan12125:
Oops seems I didn't say clearly.

> Could you move Tab X actions to a submenu (e.g., a submenu Go to). After this patch the actions menu appears bloated.

I meant a submenu `Go to` under `Actions`. Your approach also works, but I feel that's not how most applications places such actions.

Andersbakken:
> Oops seems I didn't say clearly.
> 
> > Could you move Tab X actions to a submenu (e.g., a submenu Go to). After this patch the actions menu appears bloated.
> 
> I meant a submenu `Go to` under `Actions`. Your approach also works, but I feel that's not how most applications places such actions.

Ah. That makes more sense. Done now.,

yan12125:
I pushed the last fix to eliminate the following warning:
```
/Users/yen/Projects/qterminal/src/forms/qterminal.ui: Warning: action `menu_GoTo' not declared
```
Apparently it's no longer used after moving the Go To menu under Actions.

Now it's merged into master. Thank you very much for efforts and patience invested in this useful feature!

Andersbakken:
Thanks. My pleasure.

PR #582, opened 2019-03-26
WIP: Proper version comparison.

Andersbakken:


agaida:
is this really needed?

tsujan:
> is this really needed?

I don't think so. Let's hear what @yan12125 might say.

yan12125:
The issue this tries to fix is real. However, I'd prefer another approach - just removing version conditions and migrate settings. I've heard many users follow git-master, making version numbers no longer accurate.

agaida:
given that we will have migrations from time to time - i've nothing against, ls long the implementation is clean.

agaida:
Btw - why is an exact match needed? Migrate all things < 0.14 and be done with. Or do i forget something?

yan12125:
Superseded by more-robust and simpler https://github.com/lxqt/qterminal/pull/617.

PR #588, opened 2019-05-08
Consider settings dialog frame on fitting it into screen

tsujan:
The frame thickness is taken from the terminal window, which is fully formed.

Also, by adding another scroll-bar to another page of the settings dialog, made it still more compact, so that now, it can fit into very small screens too.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -466,174 +466,211 @@
       </layout>
      </widget>
      <widget class="QWidget" name="behaviorPage">
-      <layout class="QGridLayout" name="gridLayout_5">
-       <item row="2" column="0">
-        <spacer name="verticalSpacer_2">
-         <property name="orientation">
-          <enum>Qt::Vertical</enum>
+      <layout class="QVBoxLayout" name="verticalLayout_5">
+       <property name="spacing">
+        <number>0</number>
+       </property>
+       <property name="leftMargin">
+        <number>0</number>
+       </property>
+       <property name="topMargin">
+        <number>0</number>
+       </property>
+       <property name="rightMargin">
+        <number>0</number>
+       </property>
+       <property name="bottomMargin">
+        <number>0</number>
+       </property>
+       <item>
+        <widget class="QScrollArea" name="scrollArea1">
+         <property name="frameShape">
+          <enum>QFrame::NoFrame</enum>
          </property>
-         <property name="sizeHint" stdset="0">
-          <size>
-           <width>20</width>
-           <height>57</height>
-          </size>
+         <property name="widgetResizable">
+          <bool>true</bool>
          </property>
-        </spacer>
-       </item>
-       <item row="1" column="0">
-        <widget class="QGroupBox" name="groupBox_4">
-         <property name="title">
-          <string>Emulation</string>
-         </property>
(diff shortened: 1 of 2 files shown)


yan12125:
Here are before-and-after screenshots for this patch on macOS/Aqua:

before

after

Maybe on macOS the titlebar height shouldn't be included?

tsujan:
Your screenshot shows no problem. All buttons are shown in it.

With the patch, the dialog isn't made compact without good reason but it can be as small as this when needed:

small_dialog

yan12125:
I wanted to say that the scrollbar is not necessary. A screenshot for the whole desktop should be clearer:
after
The initial size of the preferences window can be larger as there is still space.

tsujan:
The goal was is make the window *smaller* to fit it into small screens, not to make it big to fill other screens ;)

If the size provided by the `.ui` file fits into the screen, nothing will be done. The existence of absence of scrollbars depends on the style and isn't related to this.

tsujan:
A Qt style that doesn't waste space as much as the style of macOS does:

style

yan12125:
I guess I got the difference between your settings and mine. On Linux I use 13pt as the default font size for lxqt-qtplugin, and 700x700 (the default size of propertiesdialog) is not large enough for displaying all items.

Maybe those who like large texts should know what they're doing... Feel free to merge this PR by yourself, anyway.

tsujan:
@yan12125 The main difference between your and my screenshots is in style's layout spacing. the layout spacing of macOS's Qt style is large; hence the scrollbar in your case.

The size of settings dialog can also be remembered but I don't know whether I should add code for it or not.

yan12125:
Tried again with the default theme from Kvantum. Surprisingly, Chinese (zh_TW.UTF-8) and English (en_US.UTF-8) qterminal gives different results:

qterminal-pref-en_US
qterminal-pref-zh_TW

Looks like there's a bug in Qt around handling font sizes... That's beyond the scope of this pull request, anyway.

PS. Testing environment: 1440x900 display, default font for Qt: Noto Sans CJK TC 13pt

yan12125:
Gonna merge this as unnecessary scrollbars in some special cases are much less an issue than inaccessible buttons. Thanks for the work and the patience!

tsujan:
@yan12125 
When there's a scroll view, avoiding scroll-bars *in the code* is an impossible task and, here, there's no attempt to do so — the goal was fitting the dialog into small screens.

Frankly, I don't like unnecessary scroll-bars either. Therefore, in the latest release of FeatherPad, I made it to remember the size of its Preferences dialog, so that if the user sees a scroll-bar at first but is able to avoid it by resizing, the new size will be remembered.

It's possible to do it for QTerminal too. If you want it, I could make a PR later.

yan12125:
> When there's a scroll view, avoiding scroll-bars in the code is an impossible task and, here, there's no attempt to do so — the goal was fitting the dialog into small screens.

Indeed.

> It's possible to do it for QTerminal too. If you want it, I could make a PR later.

Remembering the size of the preferences window is not so important IMO. We can do it when other issues in qterminal are resolved and life is becoming boring :D

tsujan:
> We can do it when other issues in qterminal are resolved and life is becoming boring :D

I agree :) For now, I'm so busy with other codes that life is the opposite of boring but I'll do it when I find enough time — it should be easy to implement.

PR #591, opened 2019-05-17
fix: duplicated setHistorySize() call will overwrite the value set from Properties dialog

BLumia:
`setHistorySize()` already called inside `propertiesChanged()` which are using the value we set from the Properties dialog:

https://github.com/lxqt/qterminal/blob/cd59a92276af9a6962cd374d377851f267bfe5c0/src/termwidget.cpp#L91-L99

So it seems call `setHistorySize()` after the `propertiesChanged()` will overwrite the value with the hardcoded `5000`.

https://github.com/lxqt/qterminal/blob/cd59a92276af9a6962cd374d377851f267bfe5c0/src/termwidget.cpp#L54-L56

### DIFF
--- a/src/termwidget.cpp
+++ b/src/termwidget.cpp
@@ -53,8 +53,6 @@ TermWidgetImpl::TermWidgetImpl(TerminalConfig &cfg, QWidget * parent)
 
     propertiesChanged();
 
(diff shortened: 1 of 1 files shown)


yan12125:
Nice finding, thanks!

PR #593, opened 2019-05-24
WIP: Hide window decorations in Wayland

alarsyo:
Hi!

I open this PR to discuss a proper solution to this problem : on Wayland, in a tiling window manager (I'm using [sway](https://github.com/swaywm/sway)), QTerminal will display window decorations around the terminal, no matter what the settings for borders are.

One solution I've found from this discussion qutebrowser/qutebrowser#2267 is to set the `QT_WAYLAND_DISABLE_WINDOWDECORATION` environment variable.

Launching the current release of QTerminal with this environment variable set does the trick, but I think a proper setting inside QTerminal itself would be the best solution, hence this Pull Request :)

From the qutebrowser discussion above, I understand that this environment variable should be set quite early for Qt to account for it. I've put it inside `main.cpp`, and that works fine, but I'm sure there is a better place for this. Being not familiar with this project's architecture, I came here to ask.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -174,6 +174,8 @@ int main(int argc, char *argv[])
     app->installTranslator(&translator);
 
     TerminalConfig initConfig = TerminalConfig(workdir, shell_command);
+    if (Properties::Instance()->borderless)
+        setenv("QT_WAYLAND_DISABLE_WINDOWDECORATION", "1", 0);
     app->newWindow(dropMode, initConfig);
 
     int ret = app->exec();


yan12125:
Not sure if `QT_WAYLAND_DISABLE_WINDOWDECORATION` is the best choice. Here are my observations:

* The menu item View -> Hide Window Border no longer works on the fly. In other words, users cannot enable window borders without restarting qterminal if borders are disabled during the last qterminal run.
* The preferences window and the exit confirmation dialog (if enabled) are borderless, too.

Tested with Wayfire git-master + Qt 5.13 beta 2 on Arch Linux.

alarsyo:
Ah you're right, I hadn't thought about the fact that this could be toggled on the fly. If someone has a proper solution then, please share :)

yan12125:
There's a section commented code handling this:

https://github.com/lxqt/qterminal/blob/ca53eb288e1eb68479626f2e05ff91ce0747a6fd/src/mainwindow.cpp#L360-L364

I'm not sure how it's broken. If these lines work fine on Wayland, it might be an option to enable it on Wayland only.

tsujan:
> I'm not sure how it's broken.

@yan12125 Yes, it was broken and could cause a crash. The fix is in https://github.com/lxqt/qterminal/pull/664. Whether it works on Wayland or not depends on how much Qt methods work under Wayland -- WM-specific methods may not because Wayland is too restrictive.

tsujan:
Tested with Weston and, to my surprise, the borderlessness option of https://github.com/lxqt/qterminal/pull/664 worked under Wayland like X11:

weston

yan12125:
The latest qterminal 0.15.0 contains #664.

PR #602, opened 2019-07-27
Update qterminal_pt_BR.ts

bugcrazy:
Fix translation errors, removal of numeric characters in the translation that do not exist in the original.

### DIFF
--- a/src/translations/qterminal_pt_BR.ts
+++ b/src/translations/qterminal_pt_BR.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="pt_BR">
+<TS version="2.2" language="pt_BR">
 <context>
     <name>BookmarksWidget</name>
     <message>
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks for the fix! I wonder why the original translation passed the review :thinking:

bugcrazy:
I really like the LXQt project. I hope the project will evolve and advance to version 1.0.0

yan12125:
Glad to hear that :)

agaida:
PS: We use [Weblate](https://weblate.lxqt.org/translate/lxqt/qterminal/pt_BR/?checksum=&offset=&q=Which+behavior+to+emulate.+Note+that+this+d&search=ftx&source=on&target=on&type=all&ignored=False&date=&only_user=&exclude_user=) - and i am surprised, that this change is merged in Weblate - normally it breaks weblate and the work is lost :) - ok, including the work of other Translators in this file - so please:

**Translation changes only in Weblate! Thanks :smile:**

yan12125:
I remember GitHub pull requests for translations are not discouraged? Or things have changed?

agaida:
Only a year ago or so - not discouraged are PRs for desktop files and completly new translations.
But you are right, we should do some more documentation. To be honest i would like to postpone it a bit, just experimenting with wiki,js and it looks promising. And since yesterday i don't want to put all eggs into one basket to be honest. (@tsujans remarks about github restrictions)

tsujan:
> @tsujans remarks about github restrictions

Until now, those restrictions have been only for me (because of a new kind of discrimination) and they seem to be just a formality (there has been no practical change). I try to ignore them but, last night, I created a GitLab account and started mirroring from GitHub -- better be safe than sorry.

yan12125:
Ah, thanks for the explanations!

Issue #485, opened 2018-09-27
Cannot disable menu helpers (Alt-f, Alt-h, etc.)


ishovkun:
##### Expected Behavior

I should be able to use keys, such as Alt-f and Alt-h to control my shell as opposed to controlling qterminal. Fast access to the terminal menu entries is a secondary task. 

##### Current Behavior

Cannot disable menu helpers such as Alt-f, Alt-h. 

##### Possible Solution

Either introduce an option to disable the menu helpers, or simple disable them.

##### Steps to Reproduce (for bugs)

1. Open Qterminal.
2. Type some text.
3. Go to the beginning of the line.
4. Press Alt-f to jump forward.

##### Context

I want to use normal emacs-style shell navigation, among other things.

##### System Information

* Distribution & Version: Manjaro KDE testing
* Kernel: 4.17
* Qt Version: 5.11.1
* libqtxdg Version:  
* qtermwidget Version: 0.9.0
* lxqt-build-tools Version:
* Package version: qterminal 0.9.0

matrohin:
Hello!
I had the same issue and I solved it by disabling menu bar (#41 )

tsujan:
They're called "mnemonics" and I'm not sure how popular they are (I never use them in any app). They can be removed easily in the app code (ui files); I think it's a good idea to remove them from QTerminal's menubar, at least. @yan12125, what do you think?

BTW, KDE has a "feature" that's sometimes called "auto-mnemonics", so that even if "&" is removed from the ui files, KDE will add it! However, that's not our problem.

matrohin:
It was made on purpose at #139. So, maybe some people use it.
For me it's ok to have menu bar disabled (hidden). I rarely use it.
If I really need it I will do `Ctrl+Shift+M` (toggle menu bar visibility), then `Alt+<letter>` to access it.

agaida:
One should be able to toggle them on and off as they might interfere with applications that we don't know or don't care of.

tsujan:
> One should be able to toggle them on and off

I agree but there's no clean Qt method for that, as far as I know. Once you add "&" to a text, the mnemonic will be activated for its button/menubar-item/menu-item.... It's possible to change texts on the fly but I haven't seen that in any app.

agaida:
https://bugreports.qt.io/browse/QTBUG-49435?focusedCommentId=400096&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel 

[...]

PR #604, opened 2019-08-04
Made menubar accelerators optional

tsujan:
Because `Alt+F` and some other shortcuts have meaning for a terminal emulator.

Fixes https://github.com/lxqt/qterminal/issues/485

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -117,7 +117,7 @@
            </rect>
           </property>
           <layout class="QGridLayout" name="gridLayout_3">
-           <item row="17" column="1">
+           <item row="18" column="1">
             <widget class="QSpinBox" name="appTransparencyBox">
              <property name="suffix">
               <string> %</string>
@@ -194,7 +194,7 @@
              </layout>
             </widget>
            </item>
-           <item row="7" column="0">
+           <item row="8" column="0">
             <widget class="QCheckBox" name="showMenuCheckBox">
              <property name="text">
               <string>Show the menu bar</string>
@@ -204,7 +204,7 @@
            <item row="2" column="1">
             <widget class="QComboBox" name="colorSchemaCombo"/>
            </item>
-           <item row="20" column="0">
+           <item row="21" column="0">
             <widget class="QLabel" name="label_9">
              <property name="text">
               <string>Start with preset:</string>
@@ -214,7 +214,7 @@
              </property>
             </widget>
            </item>
-           <item row="20" column="1">
+           <item row="21" column="1">
             <widget class="QComboBox" name="terminalPresetComboBox">
              <item>
               <property name="text">
@@ -248,7 +248,7 @@
              </property>
             </widget>
            </item>
-           <item row="21" column="0">
+           <item row="22" column="0">
             <widget class="QLabel" name="label_15">
              <property name="text">
               <string>Terminal margin</string>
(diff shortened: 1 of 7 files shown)


yan12125:
Whoa, qterm* sees much faster development recently :)

A possible issue I see in this pull request: it might not work well with some languages. For example, in Traditional Chinese (zh_TW) `&Actions` is translated into `動作(&A)`. After stripping `&`, the menu text becomes `動作(A)`. A better result might be `動作` as `A` is no longer useful.

The same translation pattern appears in Japanese (ja) and Korean (ko_KR), too. For example `&Actions` is translated into `操作(&A)` and `동작(&A)`, respectively.

On the other hand, this pull request should work just fine with Greek and Russian. `&Actions` is `Ε&νέργειες` in Greek and `&Действия` in Russian. I guess Qt is good enough so that key sequences like `Alt + Д` just works.

tsujan:
@yan12125 I knew about Chinese and Japanese from some translations of my apps. Removal of ` (&X)` is possible but I couldn't convince myself that it's safe too. `X` not being useful just means the accelerator is visible but disabled (with the user's consent). Better safe than sorry, IMHO.

BTW, Qt even doesn't remove underlines with `qt_set_sequence_auto_mnemonic(false)` (which I didn't use because it couldn't be toggled on the fly).

> I guess Qt is good enough so that key sequences like Alt + Д just works.

Please note that accelerators should be removed from all menu-bar items and for all languages. Making exception for specific languages would create an maintainable code. After all, it's optional.

tsujan:
In a separate commit, I "took a risk" and removed `(&X)` for Chinese and Japanese, supposing that, in any language, the pattern `(&X)` couldn't be anything other than an accelerator.

If you merge it, please don't squash!

yan12125:
The code might be more readable with:
```Diff
diff --git a/src/mainwindow.cpp b/src/mainwindow.cpp
index 841ca79..8013406 100644
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -114,7 +114,7 @@ MainWindow::MainWindow(TerminalConfig &cfg,
     //consoleTabulator->setShellProgram(command);
 
     const auto menuBarActions = m_menuBar->actions();
-    for (auto action : menuBarActions)
+    for (const auto& action : menuBarActions)
         menubarOrigTexts << action->text();
 
     // apply props
@@ -646,7 +646,7 @@ void MainWindow::propertiesChanged()
     const auto menuBarActions = m_menuBar->actions();
     if (Properties::Instance()->noMenubarAccel)
     {
-        for (auto action : menuBarActions)
+        for (auto& action : menuBarActions)
         {
             QString txt = action->text();
             txt.remove(QRegularExpression(QStringLiteral("\\s*\\(&[a-zA-Z0-9]\\)\\s*"))); // Chinese and Japanese
@@ -657,7 +657,7 @@ void MainWindow::propertiesChanged()
     else if (menubarOrigTexts.size() == menuBarActions.size())
     {
         int i = 0;
-        for (auto action : menuBarActions)
+        for (auto& action : menuBarActions)
         {
             action->setText(menubarOrigTexts.at(i));
             ++i;
```

tsujan:
@yan12125 Yes. Please add a commit to PR!

yan12125:
> Please add a commit to PR!

Done!

tsujan:
@yan12125, thank you very much!

After these two PRs are merged, I'll make only one small PR before the next release and will request your and @agaida's reviews.

agaida:
GTM

Issue #492, opened 2018-10-13
why is qterminal so small ?


shirishag75:
See the following - 

![qterminal-height-weight-settings](https://user-images.githubusercontent.com/727626/46905961-f9af9a80-cf18-11e8-9391-5b690edb9d64.png)

and 

![qterminal-so-small-of-total-desktop-area](https://user-images.githubusercontent.com/727626/46905966-02a06c00-cf19-11e8-9f80-8e37e788278b.png)

for some reason by default the window is pretty small, even when I have asked it to use 100% width. 

FWIW, my system supports and uses 1600x900 screen so it's wide-screen. 

I did see that the size is hardcoded in qterminal.ini 

```
/.config/qterminal.org$ cat qterminal.ini | grep 640
size=@Size(640 480)
```

I say hardcoded as I changed it, rebooted the system and saw that the same settings were in there as earlier. 

The same I saw of keyboard shortcuts as well, changing any keyboard shortcuts in the .ini file has no effect and they go back to their default keybindings. 

I have given most of the info. of system specific at #491

shirishag75:
I did that, will know on next reboot whether it worked or not.

agaida:
@shirishag75 - but the idea of intiial settings without saved size or positions sounds nice to me, not that i missed it much until now, but i guess some people would love it.

tsujan:
> but the idea of intiial settings without saved size or positions sounds nice to me

Found this issue by chance.

I did it a long time ago but not in a way usable for all. My reason was that I wanted an appropriate, fixed, initial size. pcmanfm-qt has such a setting (I don't remember if it was there or I added it); FeatherPad too.

If I find time, I'll add it. For now, I only self-assign.

apjanke:
Related: #284. Sounds like pretty much the same request.

@agaida, @shirishag: Are you looking for a setting that will control the size of the window each time you launch `qterminal` or press Ctrl-N? Or just a setting that controls the size of the window the first time you run `qterminal` on a new machine or new desktop session?

tsujan:
@apjanke I've self-assigned but may not have time to do it soon. So, if you want to do it soon, please tell me and I'll remove the self-assignment -- 2 devs shouldn't work on the same thing.

apjanke:
Let's wait to hear back from agaida and shirishaG to confirm that this would work, but yeah, I'd like to give this a try later this week if this is what they're looking for.

[...]

PR #606, opened 2019-08-04
Allow fixed custom size for QTerminal

tsujan:
Closes https://github.com/lxqt/qterminal/issues/492

If the window size is not remembered on exiting, the custom width and height could be set in Preferences. Then, QTerminal will always have that size on starting. Of course, the maximized state is also saved/restored, as before.

Although the size is in pixels (not in characters), a button is added for getting the current size, so that the user could first resize QTerminal and then open Preferences and get the size. In this way, (1) lots of complex calculations are prevented, and (2) differences in tastes are respected.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -707,7 +707,7 @@
                  </widget>
                 </item>
                 <item>
-                 <widget class="QPushButton" name="fixedSizeButton">
+                 <widget class="QPushButton" name="getCurrentSizeButton">
                   <property name="text">
                    <string>Get current size</string>
                   </property>
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -588,7 +588,7 @@
                 </property>
                </widget>
               </item>
-              <item row="10" column="1">
+              <item row="11" column="1">
                <spacer name="verticalSpacer_4">
                 <property name="orientation">
                  <enum>Qt::Vertical</enum>
@@ -601,7 +601,7 @@
                 </property>
                </spacer>
               </item>
-              <item row="8" column="0" colspan="3">
+              <item row="9" column="0" colspan="3">
                <widget class="QCheckBox" name="useCwdCheckBox">
                 <property name="text">
                  <string>Open new terminals in current working directory</string>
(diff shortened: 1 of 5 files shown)


agaida:
Meh - i guess @tsimonq2 should review this too - will link his bug if i find the report

tsujan:
I thought 2 reviews were enough. This patch is very simple but I'm a fallible human being ;)

agaida:
hmm - i still not get, why snapping to full characters is important - but maybe this could the beginning of the implementation of this feature

tsujan:
> why snapping to full characters is important

Tastes differ. With this patch, the user can also resize in terms of characters; it's very easy (as explained in the PR comment).

yan12125:
Not sure if the current UI is intuitive or not. A possible alternative is:
```
Initial size:
(radio button) last size
(radio button) fixed size [600 px] x [400 px] {Get current size}
```

src/forms/propertiesdialog.ui
```diff
@@ -673,6 +673,64 @@
...
+                </item>
+                <item>
+                 <widget class="QSpinBox" name="fixedHeightSpinBox">
+                  <property name="suffix">
+                   <string> px</string>
+                  </property>
+                 </widget>
+                </item>
+                <item>
+                 <widget class="QPushButton" name="fixedSizeButton">
```
yan12125: Might be clearer if this button is named `getCurrentSizeButton`.

tsujan: > Might be clearer if this button is named getCurrentSizeButton.

Please do it by adding a commit to the PR.

> Not sure if the current UI is intuitive or not.

I think it is. An extra "last size" button would be redundant.

yan12125:
> Please do it by adding a commit to the PR.

Done!

> I think it is. An extra "last size" button would be redundant.

Got it!

0xdf0xdf:
I see this was in the newest version which was released about a month ago. Any idea when this will reach the kali distribution via apt upgrade?

tsujan:
> Any idea when this will reach the kali distribution via apt upgrade?

If Kali gets its packages from Debian, maybe never. @agaida was the Debian maintainer but he disappeared a few months ago.

But if Kali makes its own deb packages, you could ask its maintainers.

PR #610, opened 2019-08-18
No more incorrectly setting icon via OSC 1

yan12125:
Closes: https://github.com/lxqt/qtermwidget/issues/286

I leave other icon-handling code there as in the future I may add another escape sequence to change tab icons.


tsujan:
+1

Issue #609, opened 2019-08-06
Layout of preferences dialog is suboptimal


tsujan:
In some places, a lot of space is wasted and, especially, the list widget isn't fit to its content (its parameters are wrong).

A good layout should look like this with all languages:

![good_layout](https://user-images.githubusercontent.com/981076/62556172-987efc00-b889-11e9-9060-e767a943113f.png)

I'll fix it in a single PR alongside https://github.com/lxqt/qterminal/issues/608 because they're related to each other.

Issue #608, opened 2019-08-05
Feature: remember the size of preferences dialog


tsujan:
Since it needs no additional option and is very easy to implement, why not? It's more professional too.

I'll make a PR after https://github.com/lxqt/qterminal/pull/604 and https://github.com/lxqt/qterminal/pull/606 are decided (because of conflicts).

Issue #263, opened 2016-09-27
decrease minimal height


excalibur1234:
qterminal has a minimal height and width.
can its minimal height be decrease to e.g. the height of 2 lines?

i have many small scripts (e.g. to download youtube videos) and i like to run them in qterminal to see how much the script has progressed. but i do not need to see 12 lines of code (i see 12 lines of code in qterminal at minimum height with my current font size). 2 lines of code would be perfect.

i have used konsole for a short amount of time and it has a much lower minimal height. now, i miss such a feature in qterminal.

jcc242:
Just familiarizing myself with the code, decided to take a look at this. Removed the minimumSize property from TabWidget in qterminal.ui. See attached screenshot, top terminal is konsole for comparison, bottom terminal is qterminal. To the right is the absolute minimum size I could shrink them both to.

I could commit this and file a pull request if desired. Do we want to specify a minimum width/height? I think it would be more difficult to specify minumum number of lines as it changes based on the font size, would need a signal/slot on font size change. Original minimum size is 320x200 (pixels?).

![qterm-height](https://cloud.githubusercontent.com/assets/2275469/19011715/19e20df8-875c-11e6-9ccb-22ce0ed2e8f6.png)

excalibur1234:
wow, this looks great!

i think a minimum height/width is useful. specifying them in pixels is ok. 
i was mentioning "two line minimum height" more as a reference than something you would need to write code for.

200 pixels as the current minimum height sounds about right. this is about 1/5 of a 1920x1080 screen. as mentioned above, this is ok for doing terminal work, but not when you want a pop-up of qterminal in a screen edge with a progress bar of your running script.

i recommend to do a pull request and refer add a short text like "this PR solves #263". then, the lxqt team can discuss the advantages and disadvantages of your change.

Vladimir-csp:
Dupe of #181.
I find terminals without menus or scrollbars (or anything except the terminal itself) very handy, as they can be resized down to 1 short line to show some terse useful output.
No minimal size should be set, just increment window size hints.

agaida:
Closed with #265

PR #611, opened 2019-08-18
Remember the size of preferences dialog

tsujan:
Closes https://github.com/lxqt/qterminal/issues/609 and closes https://github.com/lxqt/qterminal/issues/608

The size is saved when the dialog is about to be closed in any way.

Also:

 * Several problems are fixed in the ui file;
 * The list widget is resized to fit its content with all languages and font sizes; and
 * Appropriate theme icons are added to the list widget (they exist in all standard icon sets and, if not existent, they just won't be shown).

NOTE: Fitting a list widget to its content can be done only in the code, as is done in this patch.

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -90,8 +90,6 @@ PropertiesDialog::PropertiesDialog(QWidget *parent)
             fixedHeightSpinBox->setValue(pSize.height());
         }
     });
-    fixedWithSpinBox->setMinimum(300);
-    fixedHeightSpinBox->setMinimum(200);
     QSize ag;
     if (parent != nullptr)
     {
@@ -106,18 +104,22 @@ PropertiesDialog::PropertiesDialog(QWidget *parent)
             }
         }
     }
-    if (ag.isValid())
+    if (!ag.isEmpty())
     {
         fixedWithSpinBox->setMaximum(ag.width());
         fixedHeightSpinBox->setMaximum(ag.height());
     }
+    fixedWithSpinBox->setMinimum(qMin(300, fixedWithSpinBox->maximum()));
+    fixedHeightSpinBox->setMinimum(qMin(200, fixedHeightSpinBox->maximum()));
 
     QStringList emulations = QTermWidget::availableKeyBindings();
     QStringList colorSchemes = QTermWidget::availableColorSchemes();
     colorSchemes.sort(Qt::CaseInsensitive);
 
     listWidget->setCurrentRow(0);
-    listWidget->setSizeAdjustPolicy(QAbstractScrollArea::AdjustToContentsOnFirstShow);
+    // resize the list widget to its content
+    listWidget->setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Expanding);
+    listWidget->setMaximumWidth(listWidget->sizeHintForColumn(0) + 2 * listWidget->frameWidth() + 4);
 
     colorSchemaCombo->addItems(colorSchemes);
     int csix = colorSchemaCombo->findText(Properties::Instance()->colorScheme);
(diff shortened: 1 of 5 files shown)


src/propertiesdialog.cpp
```diff
@@ -106,18 +104,22 @@ PropertiesDialog::PropertiesDialog(QWidget *parent)
...
         }
     }
-    if (ag.isValid())
+    if (!ag.isEmpty())
     {
         fixedWithSpinBox->setMaximum(ag.width());
         fixedHeightSpinBox->setMaximum(ag.height());
     }
+    fixedWithSpinBox->setMinimum(qMin(300, fixedWithSpinBox->maximum()));
+    fixedHeightSpinBox->setMinimum(qMin(200, fixedHeightSpinBox->maximum()));
```
agaida: if we could consider that the lowest possible screen resolution we support is about 800x480 or 800x600 the minimum size could be bigger.

tsujan: Right but they're just numbers > 0. I haven't touched them in this patch; just moved them downward in the code and added a qMin() for the sake of logic.

agaida: Was only a suggestion while we are at

tsujan: And it was a good suggestion. QTerminal can be resized to ridiculously small sizes -- you could almost make it disappear if there was no WM title-bar. I'll open an issue for it.

tsujan:
@agaida I did it here because of conflicting lines (didn't want to wait for this). Preferably, don't squash on merging.

agaida:
when i squash it is mostly an incident :) - that making the terminal insane small is not a bug, it is a feature :D

tsujan:
>  that making the terminal insane small is not a bug, it is a feature :D

A feature that's good for magicians ;) Only the WM can prevent it from disappearing, especially in a big screen.

agaida:
- this is a big one

tsujan:
You're kidding. It was 300×200 when you said: "if we could consider that the lowest possible screen resolution we support is about 800x480 or 800x600 the minimum size could be bigger". Now it's 400×200.

agaida:
A bit smalller

https://i.imgur.com/Pup7uUR.png

And i'm not kidding - i talked about the size of the settings window

tsujan:
> i talked about the size of the settings window

That's the size of the main window to be set *inside* the settings window.

Change the minimums to values you see appropriate, by adding a commit. I just wanted this to be ready before the next release.

agaida:
And btw - we should rework some things regarding visibility of some elements - really a NoGo :D 

Without joke:

I can live without the menu toggle in the view menu,

the context menu should/must have a toggle for window decorations

agaida:
@tsujan: https://github.com/lxqt/qterminal/issues/263

tsujan:
> #263 - minimum size

Odd decision, IMO.

But please add a commit! If you want, you could change all occurrences of 400×200 (twice, I think) to 1×1 but not 0×0. You could also add a comment to the line in `mainwindow.cpp`; in that way, it'll be obvious why 1×1 is chosen.

agaida:
ok - btw - the minimum size thing was discussed long time - in the end nothing was speaking against

tsujan:
OK, the decision was made and we follow it. But it's better to be in the code explicitly -- and 1×1, not 0×0.

tsujan:
Thanks!

But we should really talk about the above-mentioned decision later. It was totally wrong: the window can completely disappear (= made hidden) by mistake. 1×1 at least prevents it from hiding but, in the end, it'll make no difference.

agaida:
Suggestion: one line min height, 20 chars min witdh

tsujan:
> Suggestion: one line min height, 20 chars min witdh

I agree. Let's make it 10 chars wide for those who are in love with miniature.

This PR has become a mess. Let me make another one with the changes.

agaida:
ok

agaida:
Btw - no matter how small the window is - it will be in the tasklist - and if users have a use case for this - why should be prevent it?

tsujan:
QTerminal may be used in places without a panel or task list. 1 char high and 10 char wide should be the smallest possible size; smaller that that would be good for nothing.

agaida:
>Unix was not designed to stop you from doing stupid things, because that would also stop you from doing clever things.

    Doug Gwyn, in Introducing Regular Expressions (2012) by Michael Fitzgerald

tsujan:
Don't you know it? Doug agreed with a small change I made:

Unix was not designed to stop you from doing stupid things, because that would also stop you from doing clever things; it just stops you from doing stupidest things.

PR #614, opened 2019-08-18
Remember the size of preferences dialog

tsujan:
Closes https://github.com/lxqt/qterminal/issues/609 and closes https://github.com/lxqt/qterminal/issues/608

The size is saved when the dialog is about to be closed in any way.

Also:

 * Several problems are fixed in the ui file;
 * The minimum size is set to 10 spaces × font height, for the window not to disappear accidentally;
 * The list widget is resized to fit its content with all languages and font sizes; and
 * Appropriate theme icons are added to the list widget (they exist in all standard icon sets and, if not existent, they just won't be shown).

NOTE: Fitting a list widget to its content can be done only in the code, as is done in this patch.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -76,6 +76,17 @@ MainWindow::MainWindow(TerminalConfig &cfg,
 
     setupUi(this);
 
+    // Allow insane small sizes - reason:
+    // https://github.com/lxqt/qterminal/issues/181 - Minimum size
+    // https://github.com/lxqt/qterminal/issues/263 - Decrease minimal height
+    QFontMetrics metrics(Properties::Instance()->font);
+#if (QT_VERSION >= QT_VERSION_CHECK(5,11,0))
+    int spaceWidth = metrics.horizontalAdvance(QChar(QChar::Space));
+#else
+    int spaceWidth = metrics.width(QChar(QChar::Space));
+#endif
+    setMinimumSize(QSize(10 * spaceWidth, metrics.height()));
+
     m_bookmarksDock = new QDockWidget(tr("Bookmarks"), this);
     m_bookmarksDock->setObjectName(QStringLiteral("BookmarksDockWidget"));
     m_bookmarksDock->setAutoFillBackground(true);
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -14,7 +14,7 @@
    <string>Terminal settings</string>
   </property>
   <layout class="QGridLayout" name="gridLayout_4">
-   <item row="2" column="0" colspan="3">
+   <item row="1" column="0" colspan="2">
     <widget class="QDialogButtonBox" name="buttonBox">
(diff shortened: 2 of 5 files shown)


tsujan:
The reason why I included all these in a single PR was just they were related to each other and I wanted them to be in the next release. A bad practice; I know ;)

agaida:
Looks GTM

tsujan:
Haha... At least, without borders, there's something to right click :)

agaida:
i would suggest the double min height - so at least one line line is visible - and in my example the menu is toggled off ... - with menu turned on there is no terminal visible anymore :D

tsujan:
And this my screenshot; find the terminal in it ;)

tiny

agaida:
i see it :P

tsujan:
> i would suggest the double min height - so at least one line line is visible

That's really about how I improved Doug Gwyn's quotation: Such a small terminal is stupid but not the stupidest; the user can still find and resize it :)

agaida:
/me can hear the crying users of unknown but graphically advanced distributions ...

src/mainwindow.cpp
```diff
@@ -76,6 +76,17 @@ MainWindow::MainWindow(TerminalConfig &cfg,
...
+    // Allow insane small sizes - reason:
+    // https://github.com/lxqt/qterminal/issues/181 - Minimum size
+    // https://github.com/lxqt/qterminal/issues/263 - Decrease minimal height
+    QFontMetrics metrics(Properties::Instance()->font);
+#if (QT_VERSION >= QT_VERSION_CHECK(5,11,0))
+    int spaceWidth = metrics.horizontalAdvance(QChar(QChar::Space));
+#else
+    int spaceWidth = metrics.width(QChar(QChar::Space));
+#endif
+    setMinimumSize(QSize(10 * spaceWidth, metrics.height()));
```
agaida: , 2 * metrics. would be better - prevent follow up bugs

agaida: otherwise fine

tsujan: It would be too "big". My measures have changed since our last discussion: https://github.com/lxqt/qterminal/pull/611

Really... it can be made very small without vanishing or creating any problem.

tsujan:
tiniest

agaida:
looks good

tsujan:
Oh, I missed this:

> don't you think that a "Toggle Window Borders" without a checkbox would look better?

Yes, I do. But it isn't about https://github.com/lxqt/qterminal/pull/615; it can have/lack that checkbox everywhere, not just in the context menu, although it may not be bad either because window border is important.

src/mainwindow.cpp
agaida: ok, convinced

agaida:
@yan12125  - good to you too?

yan12125:
All looks great! Thank you very much for UI improvements!

tsujan:
Thank *you* for improving its core!

Issue #613, opened 2019-08-18
Context menu should have a toggle for Window Decorations / Window Border


agaida:
I can live without the menu toggle in the view menu, but the context menu should/must have a toggle for window decorations

![](https://i.imgur.com/KB4v0Cx.jpg)

Follow up to #611

tsujan:
I agree. For the next release or after it?

agaida:
if it is easy, for the next - otherwise after - doesn't really matter

tsujan:
Should be easy.

PR #615, opened 2019-08-18
Added "Hide Window Borders" to context menu

tsujan:
Closes https://github.com/lxqt/qterminal/issues/613


agaida:
Ah, ok - i see - so GTM

Issue #616, opened 2019-08-21
configuration updates fail - or messages are misleading


agaida:
started from terminal:
```
 % qterminal  
Properties constructor called  
Warning: Configuration file was written by a newer version  of QTerminal.  
Some settings might be incompatible.
```
Doesn't change when restarted, even if settings are changed.

##### System Information
latest git

agaida:
https://github.com/lxqt/qterminal/pull/582 - maybe working on this will help.

tsujan:
Lexically, the string "0.14.1" is less than "0.9.0", for example; hence the message.

tsujan:
The fix: https://github.com/lxqt/qterminal/pull/617

yan12125:
Thanks! Haven't had time to look into it :/

PR #617, opened 2019-08-22
Fixed version comparison

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/616

### DIFF
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -261,6 +261,30 @@ void Properties::saveSettings()
     m_settings->setValue(QLatin1String("PrefDialogSize"), prefDialogSize);
 }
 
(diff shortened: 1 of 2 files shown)


agaida:
https://doc.qt.io/qt-5/qversionnumber.html

tsujan:
No need to it.

agaida:
i just want to mention it.

tsujan:
Thanks! It's quite good to use it. Tastes differ though: I prefer to add my simple functions when a whole class isn't needed.

That being said, if you feel it's really necessary to add that class, it'll be OK to me.

agaida:
Your PR is good - and it is not needed to use qversionnumber

PR #620, opened 2019-08-27
Fix warning.

Andersbakken:
I got this one:

../src/main.cpp: In function ‘int main(int, char**)’:
../src/main.cpp:122:14: warning: ignoring return value of ‘int chdir(const char*)’, declared with attribute warn_unused_result [-Wunused-result]
         chdir(QDir::homePath().toLatin1().data());
         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

From gcc (Ubuntu 8.3.0-6ubuntu1) 8.3.0

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -119,7 +119,9 @@ int main(int argc, char *argv[])
         // On macOS, if qterminal.app is spawned by launchd (e.g., from Finder
         // or use `open qterminal.app`, $PWD is set to /. Workaround that by
         // go to $HOME first.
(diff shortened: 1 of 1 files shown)


agaida:
@yan12125  - what do you think?

yan12125:
Currently the error message reads: (after `chmod 000 $HOME` on macOS)

> Failed to chdir to $HOME "/Users/yen" Permission denied

Better to have a colon before the actual error (Permission denied). Not a big deal, anyway, as it's unlikely that chdir() to $HOME would fail.

Issue #634, opened 2019-10-17
noise when starting a terminal from command line


alvherre:
####  Expected behavior

When I start a new terminal from a command line (another terminal), I expect that the console that I launch it from should be silent. No other terminal that I use (I am switching from lxterminal, previously used rxvt and xterm) do any similar things.

#### Current behavior
This text appears:

```
Properties constructor called
TRANSLATIONS_DIR: Loading translation file "qterminal_es_CL.qm" from dir /usr/share/qterminal/translations
load success: true
QDir::exists: Empty or null file name
Canot open file "/home/alvherre/.config/qterminal.org/qterminal_bookmarks.xml"
Trying to load translation file from dir "/home/alvherre/.local/share"
Trying to load translation file from dir "/usr/local/share"
Trying to load translation file from dir "/usr/share"
Trying to load translation file from dir "/usr/share/qtermwidget5/translations"
Translations found in "/usr/share/qtermwidget5/translations"
default KB_LAYOUT_DIR:  "/usr/share/qtermwidget5/kb-layouts"
loadAllColorSchemes
Shell program: "/bin/bash"
("/bin/bash")
```

This looks like debugging output and is pretty uncomfortable.

##### Suggested solution

I suggest that some debugging option is turned on by default and it shouldn't.

##### System Information

* Distribution & Version: Debian 10.1
* Kernel: .19.67-2+deb10u1 (as packaged by Debian)
* Qt Version: 5.11.3
* qtermwidget Version: 0.14.1
* lxqt-build-tools Version: (no idea)
* Package version: 0.14.1-1

tsujan:
I second that.

IMHO, it's better to remove this kind of output everywhere. Developers could always add `qDebug() << ...` *temporarily* to find and fix problems; users -- including developers as users -- don't need to see it later. And if a GUI app needs to inform users about real problems, message boxes could be used.

tsujan:
Some apps may need it in a state of transition -- like when libfm was being ported into C++11 -- but I think we've passed that phase.

PR #635, opened 2019-10-17
Removed some debug outputs

tsujan:
Users don't need to seem them. If an info is really important, it should be shown in a message box, IMHO.

Closes https://github.com/lxqt/qterminal/issues/634 after https://github.com/lxqt/qtermwidget/pull/304 is merged.

### DIFF
--- a/src/bookmarkswidget.cpp
+++ b/src/bookmarkswidget.cpp
@@ -170,7 +170,7 @@ public:
         QFile f(fname);
         if (!f.open(QIODevice::ReadOnly))
         {
-            qDebug() << "Cannot open file" << fname;
+            //qDebug() << "Cannot open file" << fname;
             // TODO/FIXME: message box
             return;
         }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -177,18 +177,18 @@ int main(int argc, char *argv[])
     QString fname = QString::fromLatin1("qterminal_%1.qm").arg(QLocale::system().name().left(5));
     QTranslator translator;
 #ifdef TRANSLATIONS_DIR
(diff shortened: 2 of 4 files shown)


luis-pereira:
I do agree. What about using `QLoggingCategory` for logging/debug ? It's already used in some places.It allows enable/disable debug messages without recompilation.

agaida:
@luis-pereira - sounds good - maybe it is worth to have a look into calamares regarding the debugging level handling

tsujan:
> What about using QLoggingCategory for logging/debug ?

Too many instances of `qDebug()` not only aren't helpful to development but also clutter the code. A dev usually needs to add his own instances to find the causes of specific problems. So, IMO, removing them is more practical -- I commented them out though.

tsujan:
And, as someone who has hunted so many bugs, I've never found a real use for `QLoggingCategory`.

luis-pereira:
>  So, IMO, removing them is more practical -- I commented them out though.

Yeap. Commenting was the reason of comment. To me is remove or use `QLoggingCategory`

luis-pereira:
> And, as someone who has hunted so many bugs, I've never found a real use for `QLoggingCategory`

It allows a user that "can't" compile the project to enable the debug messages. In my experience it enhances bug reports.

tsujan:
As far as I've seen first hand, more than 2 qDebug instances can be an obstacle, while having many of them commented out could be handy: one or two could be uncommented temporarily with bug fixing. That's why, for example, I've kept most of PCMan's qDebugs in libfm-qt but out-commented them.

`QLoggingCategory` doesn't do that.

luis-pereira:
We are talking about different use-cases.
qDebug() or qCDebug() the clutter is the same.

Either way the PR does what it states.

yan12125:
+1 for commenting out qDebug() outputs. They are not really useful.

PR #664, opened 2020-02-29
Added an option for borderless state and remembered it

tsujan:
There was an incorrect assumption in the code, that had resulted in overlooking it.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -111,7 +111,7 @@
            </rect>
           </property>
           <layout class="QGridLayout" name="gridLayout_3">
-           <item row="18" column="1">
+           <item row="19" column="1">
             <widget class="QSpinBox" name="appTransparencyBox">
              <property name="suffix">
               <string> %</string>
@@ -198,7 +198,7 @@
            <item row="2" column="1">
             <widget class="QComboBox" name="colorSchemaCombo"/>
            </item>
-           <item row="21" column="0">
+           <item row="22" column="0">
             <widget class="QLabel" name="label_9">
              <property name="text">
               <string>Start with preset:</string>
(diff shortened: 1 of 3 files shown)


tsujan:
Oh, there is still a problem. I'll tell you when it's fixed.

agaida:
ok, fine - what do you think about an explicit option for?

tsujan:
I fixed the problem. It can be tested now.

> what do you think about an explicit option for?

No objection. Let the main bug be fixed by this patch and then we could add an option for it to Preferences, although I think it'll be a sort of duplication.

tsujan:
An explanation:

When I said "there is still a problem", I saw that changing an option in Preferences made the borderless window disappear. I think the original author encountered the same problem and ignored borederlessness because of it. But it had a simple solution. that is inside the patch now.

tsujan:
@agaida I listened to you and added a separate commit to this PR for an option in Preferences dialog. There was a duplicate option for hiding menubar; so, why not another one for borderlessness? Moreover, a future PR would change this patch; hence doing them together, in the current PR.

agaida:
@tsujan - thank you very much. I see the possibility  of cluttering the settings page too - with a big but: BUT we should integrate senseful options, no matter how hard it is for us. And i really mean it: For me it is easy to ask for an option that it is otherwise not visible (at a first sight) for the normal user. And i really mean it - Read the not yet done fucking manual can't be the only answer, so this is a really sensible thing - we know "option orgies" in other applications, might it be terminals or editors - I would've be fine if your answer had been: Not a visible option first hand, but doable in the config file-

tsujan:
@agaida After giving it more thought and considering different use cases, I came to the same conclusion; otherwise, I wouldn't change the title. So, I completely agree with you :)

travankor:
you guys should cut a new release once you merge this

tsujan:
> you guys should cut a new release once you merge this

As far as I know, the whole LXQt will be released soon, the last release being really old now.

tsujan:
I think this can be merged for git users. @yan12125, do you see any problem?

yan12125:
Sorry I haven't had time to actually test this. The code looks OK, so let's give it a try!

PR #666, opened 2020-03-13
Update disabled actions after a tab dies.

Andersbakken:
Actions do not get properly enabled/disabled after closing a tab.

To reproduce:

Start qterminal
split tab in two (vertical or horizonal)
create another tab
close this tab (Ctrl-D)
Notice how the actions for switching between the subterminals on the first tab are disabled


tsujan:
Thanks for describing the issue and trying to fix it! But the problem was a little deeper.

Replaced by https://github.com/lxqt/qterminal/pull/930

PR #676, opened 2020-05-10
Make close-on-middle-click optional

chrcuray:
It's too easy to close an important terminal when the middle button is also used to paste text.


tsujan:
Please don't include translations in a PR for the code.

chrcuray:
Ok.
Do you prefer I revert with a new commit or I redo the full PR ?

tsujan:
Whichever you're more comfortable with.

PR #678, opened 2020-05-10
Enhance the custom tab name behavior

chrcuray:
Modify custom tab name behavior:
- use available session name during session edit (was empty)
- restoring to terminal title when custom name is empty (was not possible)


yan12125:
Sorry for the long delay. Overall this looks good. There is one minor issue: when there isn't any custom tab name yet, using an empty string in the "Tab name" dialog makes the tab name empty. Probably better to keep the current tab name (i.e., do nothing) in this case.

chrcuray:
> Sorry for the long delay. Overall this looks good. There is one minor issue: when there isn't any custom tab name yet, using an empty string in the "Tab name" dialog makes the tab name empty. Probably better to keep the current tab name (i.e., do nothing) in this case.

PR #687, opened 2020-06-05
Loop variable copy

luis-pereira:


tsujan:
GTM

It was strange previously; probably, a result of multiple changes.

yan12125:
GTM, too. Thanks for catching this!

Most likely something was missed when support for binding multiple shortcuts to one action is added.

PR #695, opened 2020-06-20
Open new a tab to the right of the active tab

uffejakobsen:
This pull request adds a new featuer to qterminal:

Open new a tab just to the right of the active tab

This feature is inspired by a similar feature in XFCE Terminal

The feature is by default off - meaning that the existing functionality is preserved - and the new tab will be opened as the rightmost tab

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -575,7 +575,7 @@
               <string>Behavior</string>
              </property>
              <layout class="QGridLayout" name="gridLayout_2">
-              <item row="12" column="0">
+              <item row="13" column="0">
                <widget class="QLabel" name="label_17">
                 <property name="toolTip">
                  <string>This command will be run with an argument containing the file name of a tempfile containing the scrollback history</string>
@@ -643,7 +643,7 @@
                 </item>
                </layout>
               </item>
-              <item row="11" column="1">
+              <item row="12" column="1">
                <widget class="QComboBox" name="termComboBox">
                 <property name="sizePolicy">
                  <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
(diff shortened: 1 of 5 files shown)


yan12125:
Thanks for this patch. Could you use `insertTab`? That should be better than add and then move the new tab.

uffejakobsen:
@yan12125 sure - will take a look at that at home - and modify this pull req :-)

uffejakobsen:
@yan12125 pull request updated
BTW: it can be further optimized into only using insertTab() for all "add tab" scenarios - but I do not know if that is what you wanted ?

yan12125:
> BTW: it can be further optimized into only using insertTab() for all "add tab" scenarios - but I do not know if that is what you wanted ?

The current patch is good enough. Feel free to improve it further if you like :)

shlyakpavel:
Really useful feature. Thank you!

uffejakobsen:
@yan12125 pull request updated optimized/simplified - now insertTab() is used in every scenario

I have no plans for any other revisions of the code - hoping that the code is ok for you guys :-)

src/forms/propertiesdialog.ui
```diff
@@ -689,7 +689,17 @@
                 </property>
                </widget>
               </item>
-              <item row="11" column="0">
+              <item row="11" column="0" colspan="2">
+               <widget class="QCheckBox" name="openNewTabRightToActiveTab">
```
yan12125, line 693: In general we append widget type to their names. `openNewTabRightToActiveTab` -> `openNewTabRightToActiveTabCheckBox`

src/tabwidget.cpp
```diff
@@ -88,7 +88,9 @@ int TabWidget::addNewTab(TerminalConfig config)
     connect(console, &TermWidgetHolder::termTitleChanged, this, &TabWidget::onTermTitleChanged);
     connect(this, &QTabWidget::currentChanged, this, &TabWidget::currentTitleChanged);
 
-    int index = addTab(console, label);
+    const int newIndex = (Properties::Instance()->m_openNewTabRightToActiveTab ? currentIndex()+1 : count());
```
yan12125, line 91: Please add spaces besides `+`.

uffejakobsen: Done in new revision of PR

src/forms/propertiesdialog.ui
uffejakobsen: Done in new revision of PR

uffejakobsen:
@yan12125 new revision of PR with the code style changes adapted :-)

yan12125:
Thanks!

PR #698, opened 2020-07-15
Add blinking cursor option to keyboard cursor submenu

uffejakobsen:
This pull request adds a new feature to qterminal:
It is now possible to choose from the keyboard cursor submenu if the cursor should be blinking or not.


yan12125:
Thanks for this useful feature and sorry for delayed response.

As there is only one new action, and it is unlikely that there will be another action in the new group, could you drop QActionGroup and just use QAction?

yan12125:
Ah, somehow GitHub did not notify me for the update in March.

Did you check if using `QAction` instead of `QActionGroup` works?

By the way, it will be greater if a checkbox is added to the Appearance page of the preferences dialog. There is already a combo box for default cursor shape.

uffejakobsen:
@yan12125 Thanks for your reply.
Actually I just wanted to rebase my branch onto the latest source - so this push carries no improvement at all :-)
I'll take a look at your suggestions - thanks :-)

yan12125:
No problem and don't hestitate if you're unsure about anything (coding style, conventions, ...) :)

stefonarch:
Noticed that hanging around. Compiling I noticed that settings for presets ( "start with horizontal terminals ecc) are broken, opens only one. But maybe because it's not rebased.

yan12125:
In general, I don't fix pull requests but let the authors do. If an author is unresponsive for some time, others can create a new pull request to replace the current one.

uffejakobsen:
Thanks for your comments
This PR is incomplete
It only adds the option to the menu - not to the preferences dialog...
It should be reworked
I think that this PR is obsolete and should be closed.

uffejakobsen:
Closing - reasons given earlier in this thread

PR #699, opened 2020-07-16
Add shortcut editing instructions

radiocane:
I've added some instructions for editing QTerminal shortcuts, just above the Shortcut/Key list.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -710,7 +710,7 @@
                 </property>
                </widget>
               </item>
-              <item row="5" column="0" colspan="2"> 
+              <item row="5" column="0" colspan="2">
                <widget class="QCheckBox" name="closeTabOnMiddleClickCheckBox">
                 <property name="text">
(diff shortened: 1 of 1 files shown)


tsujan:
Labels can take too much space and make the widget too wide, especially with translations. Tooltips (and "What's This") specialize in showing info without cluttering the GUI. The same info can be added as the tooltip of `shortcutsWidget`.

BTW, `3 click on a Shortcut` isn't necessary. The user can press `Enter`.

radiocane:
Bumping after commit

tsujan:
Requested @yan12125's review. He knows QTerminal better than all.

yan12125:
Thanks for this! Looks good overall. Just some minor comments:

* Better to add a dot after numbers -> `1.` `2.`
* Better to capitalize the first word for each item

> `3 click on a Shortcut` isn't necessary.

Well, clicking on the shortcut is what I do. It is more intuitive for me. I would put both approaches in the tooltip.

yan12125:
Thanks!

Issue #729, opened 2020-10-31
Shortcuts can be messed up easily


tsujan:
The way to reproduce the bug:

 1. Make a backup of `~/.config/qterminal.org/qterminal.ini` because you'll need to restore it.
 2. Open Preferences → Shortcuts and sort shortcuts by pressing the Key column.
 3. Remove a shortcut from the end of the list.

Then, most probably, all your shortcuts will be ruined.

tsujan:
I corrected the report. I thought the problem was new but it isn't.

PR #730, opened 2020-10-31
Fixed shortcut setting when the Key column is sorted

tsujan:
The bug was very old. I noticed it by chance.

Fixes https://github.com/lxqt/qterminal/issues/729

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -391,7 +391,12 @@ void PropertiesDialog::saveShortcuts()
         const QString& keyValue = shortcutKeys.at(x);
         QAction *keyAction = actions[keyValue];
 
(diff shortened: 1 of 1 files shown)


tsujan:
@yan12125, I requested a review from you. Please tell me if you don't have time before November 4 because, IMHO, it should be in 0.16.0.

yan12125:
Had a quick look and it seems reasonable. If you want to have it in 0.16, go ahead! Otherwise, this needs to be postponed to 0.17 if you want an in-depth review.

tsujan:
@yan12125 
In the old code, it was supposed that the order of actions should be the same in the app and in the shortcut table. Since that mistake could totally ruin shortcuts, I think this should be in 0.16.0.

tsujan:
BTW, the remainder of the old shortcut code has another problem, which it isn't critical and can wait.

Issue #731, opened 2020-11-01
Shortcuts are always in English in Preferences dialog


tsujan:
##### Steps to Reproduce (for bugs)

1. Open QTerminal with `LANGUAGE=fr qterminal`.
2. Show the Preferences dialog.
3. Go to "Raccourcis".

Then the first column is in English, while it should be in French.

##### Possible Solution

The first column should be populated by translated strings. Using `tr()` inside `PropertiesDialog::setupShortcuts()` doesn't do anything other than returning a `QString` — `tr()` should have a literal string as its argument and it can't have it inside that function.

##### System Information

* Latest git LXQt

tsujan:
@stefonarch, could you confirm this issue with your native language? I only recently installed QTerminal translations.

stefonarch:
Always english this column, opening with  `LANGUAGE=it qterminal
` or from runner.

tsujan:
@stefonarch, thanks!

Not only the action names but also the shortcut strings are in English (`Shift` should be `Maj` in French, for example). Something is very wrong in the code.

In comparison, lximage-qt doesn't have this problem.

tsujan:
Came back to this and found the problems:

 1. The app doesn't install Qt translations.
 2. The app doesn't use translated strings in the shortcut editor.

Both issues can be fixed easily. I'll make a PR.

On a side note, the core shortcut code is weird: adding strings to `config.h`, using a method like `MainWindow::setup_Action()`, etc., are all weird and unnecessary. I guess they were written when Qt didn't translate shortcuts (Qt3?!). IMHO, after fixing issues like this one, we should modernize the core code.

Issue #570, opened 2019-03-09
Shortcut keys mislabeled in Preferences on macOS


apjanke:
On macOS, in the Preferences panel for Shortcuts, the modifier keys look incorrectly labeled. Most of them are “Cmd-something”, when the default bindings are actually “Cmd-something”.

<img width="723" alt="Screen Shot 2019-03-09 at 4 55 27 PM" src="https://user-images.githubusercontent.com/2618447/54077884-dd781580-428c-11e9-9b2b-abec087fa9c9.png">

If I double-click on the shortcut key combination to modify it, then it flips over to be “Cmd-something”, displayed using the conventional macOS abbreviations.

![qterminal-shortcut-keys-changing](https://user-images.githubusercontent.com/2618447/54077885-e0730600-428c-11e9-89fd-f7af35679390.gif)

##### Steps to Reproduce (for bugs)
1. Run `qterminal` on macOS
2. Press ⌘-, to bring up Preferences
3. Click on Shortcuts in the left-hand menu
4. Double-click on each of the Key combinations for the listed shortcuts.

##### System Information
* Distribution & Version: macOS 10.14.3 with Xcode 10.1
* Kernel: Darwin Kernel Version 18.2.0
* Qt Version: Qt 5.12.1
* libqtxdg Version:  ???
* qtermwidget Version: head of `master` (commit e292763)
* lxqt-build-tools Version: 0.6.0
* Package version: head of `master` (commit aefdf8e)

tsujan:
This is a wild guess; don't count on it:

I had a similar situation in FeatherPad. The reason was that shortcuts were saved in the "NativeText" format. Once I changed it to "PortableText" format, the problem was fixed.

PR #732, opened 2020-11-01
Fixed translation of shortcuts in Preferences

tsujan:
Also, worked around the Qt's bug about the Meta key by preventing multiple shortcuts.

Fixes https://github.com/lxqt/qterminal/issues/731

NOTE:

Actually, entering multiple shortcuts into `QKeySequenceEdit` can never result in multiple shortcuts but just creates a complex shortcut, all of whose elements should be pressed in a sequence. That's a small design flaw we have to tolerate.

That being said, (1) multiple shortcuts for a single action aren't very useful either because Qt only shows the first one in the main GUI, and (2) The separator `|` for multiple shortcuts isn't removed from the code because it can be used by editing the config file manually.

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -30,6 +30,14 @@
 #include "config.h"
 #include "qterminalapp.h"
 
+void KeySequenceEdit::keyPressEvent(QKeyEvent* event)
+{
+    // by not allowing multiple shortcuts,
+    // the Qt bug that makes Meta a non-modifier is worked around
+    clear();
+    QKeySequenceEdit::keyPressEvent(event);
+}
+
 Delegate::Delegate (QObject *parent)
     : QStyledItemDelegate (parent)
 {
@@ -392,7 +400,10 @@ void PropertiesDialog::saveShortcuts()
         QAction *keyAction = actions[keyValue];
 
         QTableWidgetItem *item = nullptr;
-        auto items = shortcutsWidget->findItems(tr(keyValue.toStdString().c_str()), Qt::MatchExactly);
+        QString txt = keyAction->text();
+        txt.remove(QRegularExpression(QStringLiteral("\\s*\\(&[a-zA-Z0-9]\\)\\s*")));
+        txt.remove(QLatin1Char('&'));
+        auto items = shortcutsWidget->findItems(txt, Qt::MatchExactly);
         if (!items.isEmpty())
             item = shortcutsWidget->item(shortcutsWidget->row(items.at(0)), 1);
         if (item == nullptr)
@@ -401,7 +412,7 @@ void PropertiesDialog::saveShortcuts()
         QList<QKeySequence> shortcuts;
         const auto sequences = item->text().split(QLatin1Char('|'));
         for (const QString& sequenceString : sequences)
-            shortcuts.append(QKeySequence(sequenceString));
+            shortcuts.append(QKeySequence(sequenceString, QKeySequence::NativeText));
         keyAction->setShortcuts(shortcuts);
     }
     Properties::Instance()->saveSettings();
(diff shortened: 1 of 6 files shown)


tsujan:
@stefonarch, please test it when you find the time. All actions and they shortcuts should appear in the GUI language inside the shortcut editor or on menus, not necessarily in English. Also, shortcuts should be modified without problem. Here, it worked with French.

stefonarch:
As for the column it looks fine, but the shortcuts are in still in english (example "shift" instead of "maiusc" in italian) but I'm not sure if those are translated anywhere.

I'd ship it, if there are problems: point release...

tsujan:
> example "shift" instead of "maiusc" in italian

I'm sure that's how Qt handles `Shift` in Italian — right or wrong — because, for example, I see `Shift+Destra` and `Shift+Sinistra` when I run QTerminal+patch with `LANGUAGE=it qterminal` (I mean "Destra" and "Sinistra").

I've seen a similar Qt problem about Dutch but with Right/Left.

> I'd ship it, if there are problems: point release...

Let's wait for @yan12125; he's the master of QTerminal.

yan12125:
Could this be postponed to the next version? In my memory the part for loading translations is fragile and I prefer to properly review it before merging.

stefonarch:
> I'm sure that's how Qt handles `Shift` in Italian

Yeah, didn't look properly, saw just "shift" and didn't look further.

tsujan:
> Could this be postponed to the next version?

I have no problem with that.

> In my memory the part for loading translations is fragile

I didn't touch it; I just added Qt translations, in the way all LXQt components do it.

src/propertiesdialog.cpp
```diff
@@ -423,9 +436,12 @@ void PropertiesDialog::setupShortcuts()
 
         const auto shortcuts = keyAction->shortcuts();
         for (const QKeySequence &shortcut : shortcuts)
-            sequenceStrings.append(shortcut.toString());
+            sequenceStrings.append(shortcut.toString(QKeySequence::NativeText));
 
-        QTableWidgetItem *itemName = new QTableWidgetItem( tr(keyValue.toStdString().c_str()) );
+        QString txt = keyAction->text();
+        txt.remove(QRegularExpression(QStringLiteral("\\s*\\(&[a-zA-Z0-9]\\)\\s*")));
+        txt.remove(QLatin1Char('&'));
```
yan12125, line 443: Could you extract these lines to a function? The complex regular expression may be changed in the future and I'd like to manage them in one place.

yan12125 (approved):
> I didn't touch it; I just added Qt translations, in the way all LXQt components do it.

Yep it turns out I worried too much. I have some bad experiences when testing translations across different platforms (macOS & Linux), but this pull request looks platform-independent when I checked again.

Looks good overall. Just a minor coding-style stuff.

src/propertiesdialog.cpp
tsujan: Will do it.

tsujan:
> I have some bad experiences when testing translations across different platforms (macOS & Linux)

If shortcuts aren't correct in macOS, please tell me! They should be correct because they're saved in the portable format and shown in the native format but who knows what might happen with a complex code?

yan12125:
Ah, that reminds me of https://github.com/lxqt/qterminal/issues/570. I guess that issue will also be fixed after this pull request? I don't have access macOS now and thus I cannot check.

tsujan:
> Ah, that reminds me of #570.

Yes, that's what I meant.

> I guess that issue will also be fixed after this pull request?

In the case of FeatherPad, for example, after I respected the above-mentioned format rules, macOS issues were fixed. Here, the code is inherited from the old times and is a little weird; so, I can't be sure. I don't have macOS; if you do, would you please test and tell me?

tsujan:
> I don't have access macOS 

Oh, sorry, I missed this part of your comment.

tsujan:
@yan12125 
In [a separate commit](https://github.com/lxqt/qterminal/pull/732/commits/527b19fcd6b2227ed3c4391ae6bd4a884cacb4ec), I added a static method to `Properties` for removing accelerators — didn't find a better place for it. If you had/have another place in mind, please tell me!

yan12125:
Thanks! A static function in Properties looks good.

PR #734, opened 2020-11-09
Fixed a mistake in settings dialog layout

tsujan:
The background image line-edit was hidden behind transparency spinbox!

@yan12125, do this and https://github.com/lxqt/qterminal/pull/732 (after it's completed) justify a point release?


yan12125:
Thanks for fixing it!

> do this and #732 (after it's completed) justify a point release?

Yep, a point release will be good.

PR #737, opened 2020-11-14
translations: Use .yaml for .desktop translation

palinek:
ref. lxqt/lxqt#1044

depends on lxqt/lxqt-build-tools#59


stefonarch:
```
 ls - /usr/share/applications/|grep qterm
-rw-r--r--   1 root root 2,3K 18 nov 08.37 qterminal.desktop
-rw-r--r--   1 root root 2,7K 18 nov 08.37 qterminal-drop.desktop
-rw-r--r--   1 root root 2,8K 24 ott 10.31 qterminal_drop.desktop
```
Dropdown desktop file needs underscore _ , otherwise  there are  2 files in /usr/share/applications

palinek:
> Dropdown desktop file needs underscore _ , otherwise there are 2 files in /usr/share/applications

There won't be. The one with underscore will not be used any more.

stefonarch:
> There won't be

On every existing installation there will still remain the old one.. Or remove it or use the same name, I think the latter is better.

palinek:
> On every existing installation there will still remain the old one.

No, it won't remain. With the upgrade the files from older package are removed and the new ones are installed.

stefonarch:
Not sure if so also  for people on arch who use AUR git packages.

palinek:
Definitely yes, that's the purpose of a packaging system.

stefonarch:
So only my issue that I'm compiling.
Testing it in weblate now?

apt-ghetto:
Would it be possible to mention the filename change (qterminal_drop.desktop.in -> qterminal-drop.desktop.in) in the changelog?

It is not really obvious to packagers, that this change was indeed intentional.

yan12125:
Is the reason behind the rename not to confuse scripts that determine langauge codes by underscores?

palinek:
> not to confuse scripts that determine langauge codes by underscores

Exactly.

yan12125:
Thanks for clarification. @palinek Could you add an item to https://github.com/lxqt/qterminal/blob/master/CHANGELOG? Usually that file is updated before a release but it is still good to add it now to avoid being forgotten during the release.

PR #739, opened 2020-11-18
unifying qterminal_ru.desktop

stefonarch:
weblate:
Avoid having translation files for both the plain language code and its equivalent territory designation (for example de and de_DE).


palinek (approved):
GTM

PR #749, opened 2020-11-28
Fixed quotes and missing lines

stefonarch:


src/translations/qterminal_el.desktop.yaml
```diff
@@ -1,4 +1,4 @@
+Desktop Entry/Name: "QTerminal"
+Desktop Entry/GenericName: "Προσομοιωτής τερματικού""
```
yan12125, line 2: There are duplicated quotes - I pushed a fix directly to git master.

src/translations/qterminal_pt.desktop.yaml
```diff
@@ -1,2 +1,4 @@
+Desktop Entry/Name: "Qterminal"
+Desktop Entry/GenericName: "Emulador de terminal"
```
yan12125, line 2: Hmm, authorship by `Sérgio Marques <smarquespt@gmail.com>` is lost and weblate is confused (https://github.com/lxqt/qterminal/commit/dce787632f5558ceb08fe82245dd1b90c8722123) after adding new translation here.

stefonarch: I found several commits already merged with the missing quotes.
The doubled ones come from my setting in Featherpad, should change it, some slipped through I see...

yan12125: I'm afraid not if we don't want force pushes to this git repository, so I Ieave a comment here.

Issue #741, opened 2020-11-18
qterminal keyboard operated scroll-back not working


uffejakobsen:
qterminal keyboard operated scroll-back not working

In qterminal-0.15 and earlier versions it was possible to traverse the scroll-back buffer (history) using SHIFT+PGUP/PGDN

This does not work with qterminal-0.16 - even with a history of 1000 lines you can only scroll back half a page using SHIFT+PGUP/PGDN

If you use the mouse to operate the scroll back slider it works as expected

Seen with qterminal-0.16 on ArchLinux

yan12125:
I cannot reproduce the issue with qterminal 0.16.1-1 on Arch Linux. Arelated change might be https://github.com/lxqt/qtermwidget/pull/352. Do you use a custom keyboard layout?

uffejakobsen:
I'm also on qterminal 0.16.1-1 on Arch Linux and I still see the problem...

>  Do you use a custom keyboard layout?
I'm not really sure what you refer to here - is it my keyboard language layout within the Desktop Env - then yes (Danish keyboard layout)

yan12125:
> A second option is the option inside qterminal called "emulation" in the "behaviour" section - is guess this is the target for your question ?

Yes that's what I'm asking. Sorry for not being clear enough.

> Anyway the emulation setting in qterminal is "default"

Hmm, that should work correctly.

Let's check if this is a regression first. I created an Arch Linux package with https://github.com/lxqt/qtermwidget/pull/352 reverted. Could you install that and see if the the issue persists?

[qtermwidget-0.16.1-1-x86_64.pkg.tar.zst.zip](https://github.com/lxqt/qterminal/files/5577424/qtermwidget-0.16.1-1-x86_64.pkg.tar.zst.zip)
(Uploaded as ZIP as GitHub does not like pkg.tar.zst)

uffejakobsen:
Hello, 
Thanks for your effort and clarification.
I accidentally stumbled over an explanation for the problem - without testing your library

The preferences option "Behavior" / "Action after paste" was the cause of the problem.
Setting it to "No move" enables me to traverse the whole scroll-back buffer operating by SHIFT+PGUP/PGDN
Changing the "Action after paste" to anything other than "No move" will give the behavior that started this issue/ticket

It doesn't really make any sense that this option should affect scroll-back keyboard operation
[...]

PR #752, opened 2020-11-30
Added swap Mouse Button 2 and 3 Option

macrl2000:
If this would be intresting - i added a putty style mouse button option, you can swap middle and right button to have
same behaviour as in putty (paste with right mouse button, context menu middle button).
In config.ini one can now set SwapMouseButtons2and3=true to have button 2 and 3 functionality swapped.
regards


yan12125:
Thanks for this pull request. The idea is good, but changed codes interfere with my ongoing fix for a bug in paste handling (#741), and it is likely that this PR will need major changes after that. I will update the status when needed.

yan12125:
Hmm here is something wrong - the base branch should be `master` instead of `releases-0.14.x`. Please rebase and then change the base branch.

yan12125:
> Please rebase and then change the base branch.

Oh, seems I can change the base branch. Please just rebase the branch.

PR #762, opened 2020-12-11
Use explicit QFlags default constructor instead of nullptr

luis-pereira:
The use of nullptr is deprecated.


luis-pereira:
Merging it. Trivial stuff.

PR #768, opened 2020-12-15
Removed two warnings that were shown when QTerminal started

tsujan:
They were identical and would be seen if Qt warnings weren't suppressed (in ~/.config/QtProject/qtlogging.ini): "QDir::exists: Empty or null file name"


yan12125:
Nice fix! I noticed that but forgot about it for being busy on other issues.

tsujan:
I hadn't seen them because I've put this in `~/.config/QtProject/qtlogging.ini` to prevent Qt from cluttering system logs with useless warnings:
```
[Rules]
*.warning=false
```
Once in a while, I remove the above lines to check apps.

Issue #43, opened 2014-04-11
Hitting Ctrl buffers a scroll-to-bottom event


jleclanche:
When you hit "Ctrl", or "Alt" or similar modifier keys, the next scroll event will scroll all the way to the bottom. This is very annoying as it means that if you do something like Ctrl+Shift+C to copy some text and then scroll up, it'll scroll all the way back down.

worblehat:
I guess this is the reason:
https://github.com/qterminal/qtermwidget/blob/master/lib/TerminalDisplay.cpp#L2634-L2658

Actually I don't really understand the purpose of that `if(event->modifiers().testFlag(Qt::ShiftModifier)...` branch. Appearently it always scrolls the screen as specified in the "motion after pasting"-setting whenever a modifier key is pressed.

worblehat:
What should the "Action after paste"-setting actually do? I see no difference when pasting no matter what option of the setting I choose...

PR #771, opened 2020-12-18
Move paste-related codes to qtermwidget & some fixes for pasting

yan12125:
Most codes are from https://github.com/lxqt/qterminal/pull/309

Fixes https://github.com/lxqt/qterminal/issues/43
Fixes https://github.com/lxqt/qterminal/issues/741

### DIFF
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -111,7 +111,8 @@ void Properties::loadSettings()
     /* default to BlockCursor */
     keyboardCursorShape = m_settings->value(QLatin1String("KeyboardCursorShape"), 0).toInt();
     hideTabBarWithOneTab = m_settings->value(QLatin1String("HideTabBarWithOneTab"), false).toBool();
-    m_motionAfterPaste = m_settings->value(QLatin1String("MotionAfterPaste"), 0).toInt();
+    // For "Motion after paste", 2 (scrolling to bottom) makes more sense
+    m_motionAfterPaste = m_settings->value(QLatin1String("MotionAfterPaste"), 2).toInt();
     m_disableBracketedPasteMode = m_settings->value(QLatin1String("DisableBracketedPasteMode"), false).toBool();
 
     /* fixed tabs width */
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -786,6 +786,9 @@
                 <property name="text">
                  <string>Forcefully disable bracketed paste mode</string>
                 </property>
+                <property name="toolTip">
+                 <string>Bracketed paste mode is useful for pasting multiline strings.</string>
+                </property>
                </widget>
               </item>
              </layout>
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -187,7 +187,7 @@ PropertiesDialog::PropertiesDialog(QWidget *parent)
     /* actions by motion after paste */
 
     QStringList motionAfter;
-    motionAfter << tr("No move") << tr("Move start") << tr("Move end");
+    motionAfter << tr("No move") << tr("Scrolling to top") << tr("Scrolling to bottom");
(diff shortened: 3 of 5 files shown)


yan12125:
Hi @tsujan could you have a look at this pull request and the dependent one https://github.com/lxqt/qtermwidget/pull/387? Specifially, I'd like to know if it is a good idea to remove the "Action after paste" option.

tsujan:
Yes, of course! I might find the time tomorrow night or a little later but will definitely do it.

tsujan:
As for "Action after paste", I can tell you right now that anything other than scrolling to the bottom feels counterintuitive to me. So, I agree with its removal.

ito32bit:
I use "No move" of "Action after paste" option.
However, I may be the only user using this option.

yan12125:
> I use "No move" of "Action after paste" option.

Thanks a lot for the feedback! Could you share about the need of that option?

Actually, I have "No move" before I got the bug report #741. I don't remember why I used that. Maybe because it is the default, or "Move end" is broken (#43, #741), which will be fixed after this PR.

tsujan (approved):
I couldn't find a problem in the code of this PR or that of https://github.com/lxqt/qtermwidget/pull/387. I also tried them after reading their corresponding bug reports and by testing various combinations of settings. In all cases, they worked as expected.

GTM.

tsujan:
> I use "No move" of "Action after paste" option.

IMHO, the lack of scrolling down on pasting was dangerous (I always scrolled down manually before pasting) and scrolling up didn't make sense. With these PRs, the view is scrolled down when something is pasted, so that the user couldn't miss the pasted text (before hitting `Enter`).

tsujan:
@yan12125 
This isn't directly about the PRs but only a suggestion: What about adding this tooltip to the option that disables the bracketed mode (my old PR): "Bracketed paste mode can be useful with multi-line texts."

ito32bit:
> Thanks a lot for the feedback! Could you share about the need of that option?

For example, when I install a package, I do the following:

1. type `apt depends lxqt lxqt-core` and press Enter key (List the dependent packages)
1. type `sudo apt install `
1. scroll with mouse or Shift+PgUp or Shift+Up ...
1. [1] select a package name from "Recommends:" or "Suggests:" with the mouse
1. [2] press Shift+Insert
1. repeat [1] and [2]
1. press Shift+End to scroll to bottom
1. press Enter key to install packages

yan12125:
> For example, when I install a package, I do the following:

Thanks a lot for sharing it! It is indeed a common scenario, so I dropped the last commit and just make "Scrolling to bottom" the default for new users.

src/properties.cpp
```diff
@@ -111,7 +111,8 @@ void Properties::loadSettings()
     /* default to BlockCursor */
     keyboardCursorShape = m_settings->value(QLatin1String("KeyboardCursorShape"), 0).toInt();
     hideTabBarWithOneTab = m_settings->value(QLatin1String("HideTabBarWithOneTab"), false).toBool();
-    m_motionAfterPaste = m_settings->value(QLatin1String("MotionAfterPaste"), 0).toInt();
+    // For "Motion after paste", 2 (scrolling to botoom) makes more sense
```
tsujan, line 114: Typo: botoom → bottom

yan12125: Oops fixed. Thanks!

yan12125:
> What about adding this tooltip to the option that disables the bracketed mode (my old PR): "Bracketed paste mode can be useful with multi-line texts."

Sorry forgot about this. Added in the third commit.

yan12125:
@tsujan does the latest revision of this PR still look good?

tsujan:
@yan12125
Yes, it's good. I don't think there's any problem but, if there is, we'll find it in practice.

yan12125:
Thanks a lot for reviewing! I merged this and the depended PR in qtermwidget.

PR #772, opened 2020-12-20
Added an option to disable bracketed paste mode

tsujan:
Follows and depends on https://github.com/lxqt/qtermwidget/pull/390.

The bracketed paste mode may not be comfortable for all users — it's so different from what happens in text editors. Hence, an option to forcefully disable it.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -575,7 +575,7 @@
               <string>Behavior</string>
              </property>
              <layout class="QGridLayout" name="gridLayout_2">
-              <item row="13" column="0">
+              <item row="14" column="0">
                <widget class="QLabel" name="label_17">
                 <property name="toolTip">
                  <string>This command will be run with an argument containing the file name of a tempfile containing the scrollback history</string>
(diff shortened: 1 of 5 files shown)


src/forms/propertiesdialog.ui
```diff
@@ -781,6 +781,13 @@
                 </property>
                </widget>
               </item>
+              <item row="3" column="0" colspan="2">
+               <widget class="QCheckBox" name="disableBracketedPasteModeCheckBox">
+                <property name="text">
+                 <string>Disable bracketed paste mode</string>
```
yan12125, line 787:
```suggestion
                 <string>Forcefully disable bracketed paste mode</string>
```

yan12125 (approved):
Looks good overall. Just a minor suggestion - as this feature provides a little security, I prefer to use strong words in the hope of that users will think for a few more seconds before disabling it.

src/forms/propertiesdialog.ui
tsujan: Done and squashed.

yan12125:
Thanks!

tsujan:
Thank *you* for reviewing!

Issue #830, opened 2021-05-14
Double click on tab bar to create new tab, does not apply preset


mhd-jalilvand:
I'm using version 0.14.1
when I open a new tab using File> New Tab then it loads as I expect
but when I create a new tab by double-click on tabs bar, it creates a single tab and presets does not apply

stefonarch:
> t when I create a new tab by double-click on tabs bar

Don't understand this - I see no way to do this by double click anywhere.
This said 0.14 is way outdated.

mhd-jalilvand:
![oie_kRH5C5MUvJkN](https://user-images.githubusercontent.com/14037860/118261319-9cad6400-b4c8-11eb-84e6-72cf0a1b7ac9.png)
0.14 version is latest version by installing debian 10

stefonarch:
Thank you, I double clicked ten times  there already in current version and nothing happens. 
Looks like this was possible once.

tsujan:
I can confirm that the preset (Terminal settings →  Appearance → Start with preset) won't be applied in the latest version of QTerminal if the tab is opened by double clicking the tab-bar. Reopening...

@stefonarch, it's strange that double clicking doesn't work for you. The code opens a new tab. Another bug that needs investigation?

tsujan:
There's yet another problem related to this:

The option says, "Start with preset". Therefore, logically, *QTerminal itself* should also be started with the preset, not just its new tabs. If there's another option that neutralizes it, the behavior is counter-intuitive; if not, it's about the same bug.

stefonarch:
> @stefonarch, it's strange that double clicking doesn't work for you. The code opens a new tab. Another bug that needs investigation?

No, the bug was inside me here, sorry. Related https://github.com/lxqt/qterminal/issues/351 ?

tsujan:
> Related to #351 ?

Oh, yes, it's the second problem. I'll take a look at both when I find the time; hadn't used the preset before.

tsujan:
A fix for both problems is here: https://github.com/lxqt/qterminal/pull/831

mhd-jalilvand:
Thank you @tsujan , It work as expected

tsujan:
Thanks for testing!

PR #352, opened 2017-08-29
Apply terminals preset settings to the first tab of a window as well

yan12125:
Fixes #351

PR #831, opened 2021-05-14
Respect preset on opening new window and double clicking tabbar

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/830 and fixes https://github.com/lxqt/qterminal/issues/351

The patch also fixes:

  * The problem about the focused terminal with horizontal/vertical splitting (it was wrong when a new window was opened); and
  * The issue in 4-terminal splitting.

### DIFF
--- a/src/tabwidget.cpp
+++ b/src/tabwidget.cpp
@@ -279,11 +279,17 @@ bool TabWidget::eventFilter(QObject *obj, QEvent *event)
         if (index == -1)
         {
             if (Properties::Instance()->terminalsPreset == 3)
+            {
                 preset4Terminals();
+            }
             else if (Properties::Instance()->terminalsPreset == 2)
+            {
                 preset2Vertical();
+            }
             else if (Properties::Instance()->terminalsPreset == 1)
+            {
                 preset2Horizontal();
+            }
             else
             {
                 TerminalConfig defaultConfig;
--- a/src/tabwidget.cpp
+++ b/src/tabwidget.cpp
@@ -564,6 +564,8 @@ void TabWidget::preset4Terminals()
     TermWidgetHolder* term = reinterpret_cast<TermWidgetHolder*>(widget(ix));
 
     // see preset2Horizontal() for an explanation
+    // Waiting for the first focus change is enough because, after it happens,
+    // the window is active and the other events happen serially.
     QObject::disconnect(mFocusConnection);
     mFocusConnection = connect(term, &TermWidgetHolder::termFocusChanged, this, [this, term] {
         QObject::disconnect(mFocusConnection);
--- a/src/tabwidget.cpp
+++ b/src/tabwidget.cpp
@@ -64,7 +64,9 @@ TabWidget::TabWidget(QWidget* parent) : QTabWidget(parent), tabNumerator(0), mTa
 }
 
 TabWidget::~TabWidget()
-{}
+{
+    QObject::disconnect(mFocusConnection);
+}
 
 TermWidgetHolder * TabWidget::terminalHolder()
 {
@@ -276,8 +278,17 @@ bool TabWidget::eventFilter(QObject *obj, QEvent *event)
         int index = tabBar()->tabAt(e->pos());
         if (index == -1)
         {
-            TerminalConfig defaultConfig;
-            addNewTab(defaultConfig);
+            if (Properties::Instance()->terminalsPreset == 3)
+                preset4Terminals();
+            else if (Properties::Instance()->terminalsPreset == 2)
+                preset2Vertical();
+            else if (Properties::Instance()->terminalsPreset == 1)
+                preset2Horizontal();
+            else
+            {
+                TerminalConfig defaultConfig;
+                addNewTab(defaultConfig);
+            }
         }
         else
             renameSession(index);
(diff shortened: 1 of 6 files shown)


stefonarch:
Looks working fine.

tsujan:
Thanks for testing!

tsujan:
Oh, now I see that @yan12125 did a part of this in https://github.com/lxqt/qterminal/pull/352 but didn't merge it because of the 4-terminal preset (which I mentioned in the NOTE of the PR comment). I think https://github.com/lxqt/qterminal/issues/351#issuecomment-325703245 is a separate issue that needs a separate fix.

tsujan:
OK, I fixed all problems (the 4-terminal issue too) and included all fixes in this PR.

@stefonarch, please test again when you find the time!

stefonarch:
Didn't find any misbehaviour anymore both in dropdown and terminal, thanks!

src/tabwidget.cpp
```diff
@@ -276,8 +278,17 @@ bool TabWidget::eventFilter(QObject *obj, QEvent *event)
         int index = tabBar()->tabAt(e->pos());
         if (index == -1)
         {
-            TerminalConfig defaultConfig;
-            addNewTab(defaultConfig);
+            if (Properties::Instance()->terminalsPreset == 3)
+                preset4Terminals();
```
yan12125, line 282: Could you add curly brackets for all if/else blocks? I always do that after Apple's "goto fail;" bug revealed.

src/tabwidget.cpp
```diff
@@ -512,33 +523,52 @@ void TabWidget::preset2Horizontal()
...
-    // switch to the 1st terminal
-    term->directionalNavigation(NavigationDirection::Top);
+    // see preset2Horizontal() for an explanation
+    QObject::disconnect(mFocusConnection);
+    mFocusConnection = connect(term, &TermWidgetHolder::termFocusChanged, this, [this, term] {
+        QObject::disconnect(mFocusConnection);
+        term->splitHorizontal(term->currentTerminal());
+        term->directionalNavigation(NavigationDirection::Left);
+        term->splitHorizontal(term->currentTerminal());
+        // switch to the 1st terminal (the focus is already changed)
```
yan12125, line 567: Is this statement always true? Otherwise, the focus may be at the left-bottom subterminal if the next line (`directionalNavigation`) is run before the terminal is actually split. Not a big deal, though.

src/termwidgetholder.cpp
```diff
@@ -390,6 +390,7 @@ void TermWidgetHolder::setCurrentTerminal(TermWidget* term)
         {
             emit termTitleChanged(windowTitle(), QString{});
         }
+        emit termFocusChanged();
```
yan12125, line 393: Could you use `Q_EMIT` instead? In the long term, I'd like to introduce `QT_NO_KEYWORDS` into qterminal/qtermwidget. Feel free to replace other `emit` usage you see nearby :)

yan12125 (commented):
Thanks a lot for this brilliant solution! Also thanks others for testing!

src/tabwidget.cpp
tsujan: Sure.

src/tabwidget.cpp
tsujan: > Is this statement always true?

If you mean "the focus is already changed", yes, it's always true at this stage. Our problem was that we didn't wait for the window activation. By waiting for `termFocusChanged`, we guarantee that the focus is changed correctly *afterward*.

src/termwidgetholder.cpp
tsujan: Sure.

tsujan:
Done.

As for `Q_EMIT`, I changed 2 other instances near my code line but you might want to check and replace all instances of `emit` with it in another PR — probably, in `qtermwidget` too.

src/tabwidget.cpp
yan12125: > If you mean "the focus is already changed"

Yep

> By waiting for `termFocusChanged`, we guarantee that the focus is changed correctly _afterward_.

I think only the focus change signal from `splitVertical` is waited. The focus change from the second `splitHorizontal` call in the inner lambda function is not?

yan12125:
> you might want to check and replace all instances of `emit` with it in another PR

Yep, "when I have time" LOL

src/tabwidget.cpp
tsujan: Waiting for the first focus change is enough because, after it happens, the window is active and the other events happen serially.

yan12125: Ah, that makes sense. Could you add that explanation to the comments?

tsujan: > Could you add that explanation to the comments?

I already did :) I also added two instances of this comment for the code reader:

```c++
// see preset2Horizontal() for an explanation
```

tsujan: You read these codes more that the rest of us; feel free to add to the branch the explanation you see fit. Sorry, I don't know what you have in mind — to me, that comment explains everything concisely.

yan12125: I added a commit with expanded explanations. Could you check if I understand you correctly?

tsujan: > Could you check if I understand you correctly?

I think we both understand it correctly :)

yan12125:
Thanks! Let's move on!

tsujan:
Thanks for reviewing! Having lots of firsthand experience in it, I know how boring it can be.

yan12125:
Not at all :) Reviewing your patches is delightful thanks to the high quality.

PR #842, opened 2021-07-06
Trim shell QString.

textbrowser:


yan12125:
Thanks for the fix! Could you do trimming inside `getShell()`? If there are other usages of `getShell()` in the future, they will be benefited as well.

textbrowser:
Also changed getShell().

textbrowser:
How about trimming the strings in the conditionals? For example
```if(!m_shell.trimmed().isEmpty())```.

yan12125:
> How about trimming the strings in the conditionals?

That sounds better!

yan12125:
Thanks! I think the change in  src/termwidget.cpp is no longer needed?

textbrowser:
Yes, please discard that change.

yan12125:
> Yes, please discard that change.

GitHub does not seem to allow merge only some commits. Could you do that locally with git? Squashing previous commits is fine.

textbrowser:
I've reverted the "index-verification" push. I'm sorry for including that revision with this pull request.

yan12125:
No problem and thanks!

PR #852, opened 2021-09-03
Adding same new issue templates as in lxqt

stefonarch:


yan12125:
Actually there is already an issue template https://github.com/lxqt/qterminal/blob/master/.github/ISSUE_TEMPLATE.md. We cannot stop users from ignoring it, sigh.

yan12125:
Or do you mean to replace the existing issue template with those new ones? Then remove the old one.

At a first glance,  `liblxqt Version` in bug_report.md is not needed for qterminal as liblxqt is not a dependency

stefonarch:
The existing one will not be used anymore. To see iin action: https://github.com/lxqt/lxqt/issues/new/choose


> At a first glance, `liblxqt Version` in bug_report.md is not needed for qterminal as liblxqt is not a dependency

Sure, I just copied over the templates which I used in lxqt/lxqt, will remove this.

stefonarch:
adapted to Qterminal now

yan12125:
Looks great! Could you replace all "Qterminal" with "QTerminal"? That seems the official name https://github.com/lxqt/qterminal#readme

> The existing one will not be used anymore. 

Then go ahead and remove it!

stefonarch:
Done, also reworded the warnings about LTS user in bug report template.

yan12125:
Awesome! Thanks a lot.

PR #860, opened 2021-10-11
If Wayland environment is detected, QTerminal doesn't move window position on starting

selairi:
This fixes the wrong position of menus. If you work on Wayland environment and save the Qt window position and move the window position to save one on application start, menus will be shown in the wrong position:

There is a vertical gap that increases every time you close and open the application.
Finally, the menu is drawn outside of the screen and it cannot be used.

It is easy to avoid: Don't move window position on starting.

Maybe this is not a LXQt bug, it is a bug of Qt, but it is really annoying.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -23,7 +23,7 @@
 #include <QStandardPaths>
 #include <QTimer>
 #include <functional>
-#include <QProcessEnvironment>
+#include <QGuiApplication>
 
 #ifdef HAVE_QDBUS
 #include <QtDBus/QtDBus>
@@ -117,7 +117,7 @@ MainWindow::MainWindow(TerminalConfig &cfg,
             resize(Properties::Instance()->fixedWindowSize);
         }
         if (Properties::Instance()->savePosOnExit && !Properties::Instance()->mainWindowPosition.isNull()
-            && QProcessEnvironment::systemEnvironment().value(QStringLiteral("XDG_SESSION_TYPE")) != QStringLiteral("wayland")
+            && QGuiApplication::platformName() != QStringLiteral("wayland")
             ) {
             move(Properties::Instance()->mainWindowPosition);
         }


src/mainwindow.cpp
```diff
@@ -115,7 +116,9 @@ MainWindow::MainWindow(TerminalConfig &cfg,
         else if (Properties::Instance()->fixedWindowSize.isValid()) {
             resize(Properties::Instance()->fixedWindowSize);
         }
-        if (Properties::Instance()->savePosOnExit && !Properties::Instance()->mainWindowPosition.isNull()) {
+        if (Properties::Instance()->savePosOnExit && !Properties::Instance()->mainWindowPosition.isNull()
+            && QProcessEnvironment::systemEnvironment().value(QStringLiteral("XDG_SESSION_TYPE")) != QStringLiteral("wayland")
```
tsujan, line 120: We have `QGuiApplication::platformName()`.

tsujan:
I'm not saying this as a reviewer:

To me, the strange thing about this is that, `QWidget::move()` should not be able to move a window under Wayland and so, this change should not be needed. But, in all my tests under Plasma-Wayland, the option about saving window position was unchecked. Will check that option and test later.

tsujan:
OK, I tested under Weston and can confirm @selairi's observation: with window position saved, a gap starts to appear on closing and opening QTerminal. @selairi's solution prevents that from happening.

As to why it happens, I haven't started to investigate it yet but I think it may be about how menus are positioned relative to their parent windows. The problem is in Qt.

I also guess that it may not happen under Plasma-Wayland because, somehow, KDE devs have worked around similar problems in their Wayland compositor. But a test is needed.

tsujan (approved):
GTM.

I'll make similar changes to lximage-qt and two of my apps; the Qt bug is too nasty.

tsujan:
> I think that Wayland ignores all requests about move windows, but Qt is saving the updated position, then the popups are placed in wrong positions.

That's exactly what I think too. You explained it succinctly.

> lximage-qt works properly under Wayland. 

Yes, but if you're saying that its menu positions are correct under Wayland, it's just by chance. IMO, your workaround is needed everywhere window positions are saved and restored.

tsujan:
> By the way, I noticed that the window gets smaller and smaller on Wayfire if "Save Size when closing" is checked.

If so, either it should happen under X11 too (which would mean our code has a problem) or the issue is related to Wayfire. There should be no difference between X11 and Wayland with regard to window resizing.

I'll check everything soon, under Plasma-Wayland too.

yan12125:
Indeed! A good news is that it's not a bug in Qt or QTerminal.

tsujan:
Then, do you approve of this for a merge? I requested your review.

yan12125:
Hmm, interesting. Sounds more like a Qt bug.

Issue #874, opened 2021-11-21
dropdown qterminal width


onesimply:
##### Expected Behavior

Keep the width of the dropdown terminal at 100% 

##### Current Behavior

If you change the width to 100% the next time you open the preferences is set to 99%.

##### Steps to Reproduce (for bugs)

1. Set the width of the terminal to 100% and exit the preferences window.
2. Open again the preferences window and the width is set to 99%

##### System Information

* Distribution & Version: Devuan
* Kernel: 5.10
* Qt Version: 5.15.2
* liblxqt Version: 0.16.0-1
* Package version: 1.0.0

tsujan:
Thanks for reporting! It was caused by a small mistake in the code: https://github.com/lxqt/qterminal/pull/875

PR #875, opened 2021-11-21
Fixed max. value of drop-down width/height spin-box

tsujan:
It seemed like a copy-paste typo.

Fixes https://github.com/lxqt/qterminal/issues/874


stefonarch:
Fixed, yes.

luis-pereira (approved):
Works as expected.
GTM.

PR #887, opened 2022-01-13
Enable bookmark filtering

tsujan:
Filtering did nothing before.

NOTE: This patch only enables bookmark filtering. The filter line-edit needs to be focused when the bookmark pane is shown; some other changes may also be needed. I'll make another PR after this one is reviewed and merged.

### DIFF
--- a/src/bookmarkswidget.cpp
+++ b/src/bookmarkswidget.cpp
@@ -349,7 +349,9 @@ QModelIndexList BookmarksModel::allChildRows(const QModelIndex& parent) const
         auto child = index(row, 0, parent); // we only need the row
         list << child;
         if (hasChildren(child))
+        {
             list << allChildRows(child);
+        }
     }
     return list;
 }
@@ -424,7 +426,9 @@ void BookmarksWidget::filter(const QString& str)
                 treeView->setRowHidden(index.row(), index.parent(), false);
             }
             else
+            {
                 treeView->setRowHidden(index.row(), index.parent(), true);
+            }
         }
     }
 }
--- a/src/bookmarkswidget.cpp
+++ b/src/bookmarkswidget.cpp
@@ -341,6 +341,19 @@ int BookmarksModel::rowCount(const QModelIndex &parent) const
     return parentItem->childCount();
 }
 
+QModelIndexList BookmarksModel::allChildRows(const QModelIndex& parent) const
+{
+    QModelIndexList list;
+    for (int row = 0; row < rowCount(parent); ++row)
+    {
+        auto child = index(row, 0, parent); // we only need the row
+        list << child;
+        if (hasChildren(child))
+            list << allChildRows(child);
+    }
+    return list;
+}
+
 #if 0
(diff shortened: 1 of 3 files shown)


stefonarch (approved):
Working fine, and  I see it filters also  the "non visible" parts (the bash command).

tsujan:
> I see it filters also the "non visible" parts

If you mean that parent branches (which have bold font) are always shown, yes. I saw no practical reason to add code for showing/hiding them after checking whether all their child branches are hidden.

stefonarch:
Yes, the commands, noticed it by chance. At first view one could think they are false results. 
`    <command name="Edit bookmarks" value="featherpad ~/.config/qterminal.org/qterminal_bookmarks.xml"/>
` showed up searching "term"

tsujan:
It can easily be removed (I added it intentionally) but, IMO, it can be useful. Imagine you have lots of bookmarks but don't remember their names. If you remember only parts of their commands, you'll benefit from filtering.

stefonarch:
I remember well my first glance years ago at "bookmarks" - didn't understand what their use could be and ignored them for years.
Only after typing real to much` cd ../lxqt-*` I started to use them for the folders.

But given the way to add or sort them I doubt people will end up with large lists. Nice would be an DND and right click on selected command → add to bookmarks and so on but I guess this would mean a lot of work.

tsujan:
> but I guess this would mean a lot of work.

Who knows? We might have them later. For now, let's make this functionality really usable and fix its issues. It's good that both of us started to use it and find its issues.

src/bookmarkswidget.cpp
```diff
@@ -341,6 +341,19 @@ int BookmarksModel::rowCount(const QModelIndex &parent) const
...
 
+QModelIndexList BookmarksModel::allChildRows(const QModelIndex& parent) const
+{
+    QModelIndexList list;
+    for (int row = 0; row < rowCount(parent); ++row)
+    {
+        auto child = index(row, 0, parent); // we only need the row
+        list << child;
+        if (hasChildren(child))
+            list << allChildRows(child);
```
yan12125, line 352: Better to always have curly brackets for `if` ;)

src/bookmarkswidget.cpp
```diff
@@ -393,3 +408,23 @@ void BookmarksWidget::handleCommand(const QModelIndex& index)
...
+        AbstractBookmarkItem *item = static_cast<AbstractBookmarkItem*>(index.internalPointer());
+        if (item && item->type() == AbstractBookmarkItem::Command)
+        {
+            if (item->value().contains(str, Qt::CaseInsensitive)
+                || item->display().contains(str, Qt::CaseInsensitive))
+            {
+                treeView->setRowHidden(index.row(), index.parent(), false);
+            }
+            else
+                treeView->setRowHidden(index.row(), index.parent(), true);
```
yan12125, line 427: ditto

yan12125 (approved):
Thanks, nice catch! The overall logic looks good, just some minor coding style hints~

It's great to refine some "hidden" features :)

Issue #886, opened 2022-01-13
[Feature request] Improve bookmarks panel


stefonarch:
**Is your feature request related to a problem? Please describe.**

I never ever use one of the "local bookmarks". Instead of `HOME`  most users will use `cd`, most terminal user never enter "Music" folder and so on.

**Describe the solution you'd like**

Hardcoded bookmarks should be removed in favor of  only user defined bookmarks, shipping just a list of useful examples. 
Also distinction  "local" and "synchronized" section would be  gone, just "Bookmarks", all of the XDG_* commands would work also if saved remotely,                                      

##### Context

https://github.com/lxqt/qterminal/issues/789#issuecomment-1012060723

tsujan:
> Hardcoded bookmarks should be removed in favor of only user defined bookmarks

I agree. What's the use of adding `cd ...` for someone that works with a terminal emulator? If he doesn't know about `cd`,  `PWD`, `HOME`, etc., he needs to learn them elsewhere.

tsujan:
If we remove "Local Bookmarks", I think we should also hide the top branch "Synchronized Bookmarks", so that the user could have his own meaningful tree:

![tree](https://user-images.githubusercontent.com/981076/149363009-44e7deaf-c798-444d-88e1-6044cd2bd0e6.png)

tsujan:
Apparently, the bookmark functionality isn't used by many; otherwise, there would be an open issue about bookmark filtering, which doesn't work.

stefonarch:
> Apparently, the bookmark functionality isn't used by many; otherwise, there would be an open issue about bookmark filtering, which doesn't work.

True, never used it as they aren't so many here. I specially use them for locations. "Edit bookmartks" is for using featherpad to modify them, downside they get not reloaded on save.
![schermata-01-13-17-15-01](https://user-images.githubusercontent.com/10681413/149367297-cc236f6e-0b6e-4622-b103-3886cce21ee2.png)

stefonarch:
IMHO a good and various  collection of bookmarks in the default file will show well the potentials.

tsujan:
I think, before any design change (which will need @yan12125's presence), the filtering should be enabled.

stefonarch:
Step by step, no hurry.

tsujan:
> Step by step, no hurry.

I agree. Mixed PRs can be a nightmare for reviewers — they are for me.

yan12125:
[...]

PR #893, opened 2022-01-30
Cleaned up bookmarks tree

tsujan:
This small patch

 * Removes hard-coded bookmarks;
 * Makes the top branch "Bookmarks" hidden;
 * Focuses the filtering line-edit on showing the bookmarks dock (inside the main window); and
 * Clears the filtering line-edit on pressing Escape inside the dock.

### DIFF
--- a/src/bookmarkswidget.cpp
+++ b/src/bookmarkswidget.cpp
@@ -18,6 +18,7 @@
 
 #include <QDebug>
 #include <QStandardPaths>
+#include <QShortcut>
 
 #include "bookmarkswidget.h"
 #include "properties.h"
@@ -98,58 +99,6 @@ public:
         : AbstractBookmarkItem(AbstractBookmarkItem::Group, parent)
     {
         m_display = name;
-    }
-};
-
-class BookmarkLocalGroupItem : public BookmarkGroupItem
-{
-public:
-    BookmarkLocalGroupItem(AbstractBookmarkItem *parent)
-        : BookmarkGroupItem(QObject::tr("Local Bookmarks"), parent)
-    {
-        QList<QStandardPaths::StandardLocation> locations;
-        locations << QStandardPaths::DesktopLocation
-                  << QStandardPaths::DocumentsLocation
-                  << QStandardPaths::TempLocation
-                  << QStandardPaths::HomeLocation
-                  << QStandardPaths::MusicLocation
-                  << QStandardPaths::PicturesLocation;
-
-        QString path;
-        QString name;
-        QString cmd;
-        QDir d;
-
-        // standard $HOME subdirs
-        for (const QStandardPaths::StandardLocation i : qAsConst(locations))
-        {
-            path = QStandardPaths::writableLocation(i);
-            if (path.isEmpty() || !d.exists(path))
-            {
-                continue;
-            }
-            name = QStandardPaths::displayName(i);
-
-            path.replace(QLatin1String(" "), QLatin1String("\\ "));
-            cmd = QLatin1String("cd ") + path;
-
-            addChild(new BookmarkCommandItem(name, cmd, this));
-        }
-
-        // system env - include dirs in the tree
-        QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
-        const auto keys = env.keys();
-        for (const QString &i : keys)
-        {
-            path = env.value(i);
-            if (path.isEmpty() || !d.exists(path) || !QFileInfo(path).isDir())
-            {
-                continue;
-            }
-            path.replace(QLatin1String(" "), QLatin1String("\\ "));
-            cmd = QLatin1String("cd ") + path;
-            addChild(new BookmarkCommandItem(i, cmd, this));
-        }
     }
 };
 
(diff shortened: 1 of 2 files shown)


stefonarch:
All working fine (I removed for me already the hardcoded ones but didn't manage to hide "Bookmarks".)

No idea if this is possible, similar to the panels main menu:
* clearing filter on hide
* using arrows  to run through results and launch bookmark with enter

tsujan:
> clearing filter on hide

Good idea.

> using arrows to run through results

You can, already. When the filter line-edit is focused, press Tab; then you could use arrows and Enter.

stefonarch:
> You can, already. When the filter line-edit is focused, press Tab; then you could use arrows and Enter.

press TAB shortly, otherwise you end up in the terminal. 
I noticed that if the panel is in focus fast typing first characters and enter works very fine too, without using the filter, arrow keys too. There is no shortcut for focusing the panel true?

tsujan:
> press TAB shortly, otherwise you end up in the terminal.

Right; just pressing + releasing. It's possible to make arrow keys work from the line-edit but it's beyond this small PR. I might do it in a subsequent PR.

> There is no shortcut for focusing the panel true?

If, while the dock is shown (and attached), you're typing inside terminal view and want to focus the doc, the only way is pressing its shortcut twice (to hide and show the doc again). IMHO, adding yet another shortcut just for focusing the doc would be an overkill because it should belong to the whole terminal window (→ shortcut editor). Moreover, focusing a *detached* doc may be impossible under Wayland.

> maybe better focus on the panel on opening and a shortcut for the filter instead?

The filter is the most important part. So, IMO, it should be focused on opening, as this PR does it. Using arrow keys is a secondary job.

tsujan:
Oh, since the bookmarks shortcuts can be used twice for refocusing, I think we shouldn't clear the filter on hiding the dock. After all, pressing Escape (inside the dock) is easy.

So, I think this PR doesn't need a commit for it.

stefonarch:
> So, I think this PR doesn't need a commit for it.

I agree, there are so many ways. In the last weeks I used it always open and mostly using the mouse/touchpad.

stefonarch (approved):

tsujan:
> In the last weeks I used it always open and mostly using the mouse/touchpad.

You and I may be the only ones who have discovered its usefulness. I wouldn't try it if you didn't mention it, but now it's very handy to me.

tsujan:
> As removing local bookmarks is a backward-incompatible change

It isn't. Local bookmarks are as valid as before.

yan12125:
Ah, the term `Local bookmarks` is indeed ambiguous. You know what I mean :) Maybe something like `Hard-coded bookmarks (also known as "Local Bookmarks") are removed ...` in future release notes. I prefer to include the term `Local Bookmarks` as it was the tree name shown in the bookmarks pane.

tsujan:
> Ah, the term Local bookmarks is indeed ambiguous

It isn't used anywhere in GUI. The user is just informed that he/she could add bookmarks.

On the other hand, the hard-coded bookmarks were practically useless (terminal users know how to change dir), cluttered the bookmarks dock and weren't removable. The design was bad.

tsujan:
I have no objection to adding of a note when the next release comes. I'm just saying that nothing useful is lost.

yan12125:
> It isn't used anywhere in GUI.

They do!

screenshot-43581b3e


> I'm just saying that nothing useful is lost.

Yes, so I'm +1 on removing them.

tsujan:
> They do!

That's *before* this PR ;)

terminal

yan12125:
Err, of course things removed only exist before this PR

tsujan:
Sorry, I don't understand your point. As I said, I'm not against a release note in the spring. If you think that this PR needs a change, please tell me!

stefonarch:
We just have to make the naming clear in the release notes. "local" and "sincronized" are now just "Bookmarks", because it's both: still possible to sync them (anybody tested this?)

tsujan:
> We just have to make the naming clear in the release notes. "local" and "sincronized" are now just "Bookmarks"

No objection to this either. My point is:

If those items were "Local Bookmarks", where were "Global Bookmarks"? It simply didn't make sense. Only after you opened an issue, I found that "local" was a strange misnomer for a hard-coded job as trivial as `cd ...` (who misses it?). More importantly, it hid the real usage of bookmarks.

That's why I don't see this as a migration but a bug fix.

yan12125:
> If you think that this PR needs a change, please tell me!

The only thing to change is removing the now-unneeded line `#include <QStandardPaths>`. All others are ideas about release notes and can be postponed until the next release.

> It simply didn't make sense.

I agree.

> who misses it?

Though unlikely, some people may still find it useful.

> That's why I don't see this as a migration but a bug fix.

This change indeed improves the bookmark feature by removing the bad part. Nevertheless, some long-existing stuff is removed, so release notes should indicate it. It's fine to call it something other than "migration". Just to make sure that users will understand they need some manual steps if they are using the removed stuff.

> We just have to make the naming clear in the release notes. "local" and "sincronized" are now just "Bookmarks", because it's both: still possible to sync them (anybody tested this?)

Sounds a good statement if someone confirms it by testing.

stefonarch:
> Though unlikely, some people may still find it useful.

I added some of the hardcoded ones in the draft here: https://github.com/lxqt/qterminal/pull/895/files#diff-e0ed6e34a80a1ed73f45da40aea97df67dbea34fac278e40e5034b0bccfdc4c5

yan12125:
Those are all nice-to-have additions! We can continue the discussion at https://github.com/lxqt/qterminal/issues/886

PR #895, opened 2022-01-31
Draft for bookmarks examples

stefonarch:
Some ideas for the example file. 
I see also why it was hard to use them before: file hasd to be created  by the user first.

To do list:
* save file at first run together with qterminal.ini
* Rework "Find..." button: rename "Load"; open `~/.config/qterminal.org` folder; limit to `*.xml` files
* Message box for backwards compatibility?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,6 +227,11 @@ install(FILES
     DESTINATION "${CMAKE_INSTALL_DATADIR}/appdata"
 )
 
+install(FILES
+    qterminal_bookmarks_example.xml
+    DESTINATION "${CMAKE_INSTALL_DATADIR}/qterminal"
+)
+
 if(NOT APPLEBUNDLE)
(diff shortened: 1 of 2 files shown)


yan12125:
> To do list:

Will those improvements part of this PR or other future PRs? If the latter, https://github.com/lxqt/qterminal/issues/886 may be a better place for those notes.

Besides that, this PR looks good. The "Edit bookmarks" one is nice! Although people may not use it often, it provides a good hint.

By the way, a comment like `<!-- Some examples to demonstrate how to use bookmarks. Adjust them to fit your needs. -->` should make things clearer.

stefonarch:
I see this as a single  PR only related to the the content, https://github.com/lxqt/qterminal/issues/886 is more the "collection" of all ideas, there could come more even,

yan12125:
Got it! Let's move on!

Issue #897, opened 2022-01-31
Closing last tab with "confirm exit" and answer "no" results in blank window


stefonarch:
File > quit works well, with answer "no", closing last tab with  exit, ctrl+d or  x button 
and answering "no" results in blank window:

![schermata-01-31-19-26](https://user-images.githubusercontent.com/10681413/151851374-1169ef23-6b4c-4c20-af22-e00448a50906.png)

##### Expected Behavior

Never close terminal when answering "no" at confirmation dialog.
##### Current Behavior

It closes terminal but not window.
##### Steps to Reproduce (for bugs)

1. Preferences, behaviour, check "ask for confirmation when closing
2.Close tab with either cltr-d, x button or `exit`
3. Answer "no" in the dialog

##### System Information

* QTerminal version: git, 1.0.0.
* Distribution & Version: manjaro, arch

stefonarch:
cltr+shift+t doesn't reopen tab, file > new tab does.

tsujan:
Confirming. I don't use confirmation before closing. That's why I didn't see it.

stefonarch:
Neither me, activated it only while checking old issues.
Related https://github.com/lxqt/qterminal/issues/746

stefonarch:
A clean approach should be: "confirm close tab" for all tabs. It makes no sense just ask if it's the last one, it should be similar ti browser.
Would fix also https://github.com/lxqt/qterminal/issues/509 and https://github.com/lxqt/qterminal/issues/496

tsujan:
The situation was worse. See the PR's comment: https://github.com/lxqt/qterminal/pull/898

> A clean approach should be: "Confirm tab closing" for all tabs with all 

That's a separate functionality. Here, we're talking about the *exit* prompt, which can be triggered by closing the window or its last tab.

PR #898, opened 2022-02-02
Fixed closing of last tab with prompt

tsujan:
There were a few odd inconsistencies and a dangerous situation regarding tabs:

 1. If the close button of the last tab was pressed and the exit prompt was rejected, a state would happen, in which the terminal window had no tab. That could easily result in a crash if the user called an action that was related to the active terminal (like "Clear Active Terminal" from the menu bar).
 2. Moreover, in the above-mentioned state, it was impossible to add a new tab by using the shortcut.
 3. With a single tab, if the "Close Tab" action was used from the menu bar — instead of the tab's close button — and the exit prompt was rejected, the tab wouldn't be removed. That was actually a good behavior but it was inconsistent with 1.

This patch removes the inconsistencies and avoids the crash danger by choosing 3 as the only behavior. In other words, closing the last tab is exactly like closing the window, with or without the exit prompt. A tabless terminal is also made impossible.

Fixes https://github.com/lxqt/qterminal/issues/897

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -68,7 +68,6 @@ MainWindow::MainWindow(TerminalConfig &cfg,
 #ifdef HAVE_QDBUS
     registerAdapter<WindowAdaptor, MainWindow>(this);
 #endif
-    m_removeFinished = false;
     QTerminalApp::Instance()->addWindow(this);
     // We want terminal translucency...
     setAttribute(Qt::WA_TranslucentBackground);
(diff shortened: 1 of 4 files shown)


stefonarch (approved):

tsujan:
Since here the goal wasn't the implementation of a new feature but making the current one self-consistent, and because the patch prevents the possibility of crash, I merge it, considering that it's tested by @stefonarch and me.

There are feature requests (some may be duplicates) about other kinds of prompts (e.g., see https://github.com/lxqt/qterminal/issues/897#issuecomment-1026545958). Most importantly, I agree that showing prompt only for closing the whole terminal or its last tab isn't enough but that's a separate matter.

PR #349, opened 2017-08-13
WIP current appdata

agaida:
refs #346

PR #901, opened 2022-02-15
Fix appdata.xml

stefonarch:
Fixes https://github.com/lxqt/qterminal/pull/349
I saw this used in discover, could be others too.

schermata-02-15-08-53


qterminal.appdata.xml
```diff
@@ -14,30 +14,22 @@
...
+      <li>Shortcut editor</i>
+      <li>Bookmarks for commands</li>
     </ul>
   </description>
   <screenshots>
     <screenshot type="default">
-      <image>http://worblehat.github.io/storage/qterminal_1.png</image>
-      <caption>The default QTerminal window.</caption>
+      <image>https://lxqt-project.org/images/screenshots/qterminal-1.png</image>
+      <caption>QTerminal with splitted terminals.</caption>
```
tsujan, line 29: I think it should be "split terminals".

stefonarch: You think right.

stefonarch:
I wonder if we should add those files also to screengrab, lximage-qt and lxqt-archiver.
https://www.freedesktop.org/wiki/Distributions/AppStream/

I   noticed that for example screengrab is not found by "discover" app.

stefonarch:
There are more issues here, maybe rewrite the file with the generator.
```
$ appstreamcli validate qterminal.appdata.xml 
E: ~:2: metainfo-ancient

? Validation failed: errors: 1

```

tsujan:
> I wonder if we should add those files also to screengrab, lximage-qt and lxqt-archiver.

I'm not familiar with these things (I don't use apps like `discover`) but a user kindly made this for FeatherPad and I merged it: https://github.com/tsujan/FeatherPad/blob/master/featherpad/data/featherpad.metainfo.xml. Perhaps, similar files could be added for LXQt.

stefonarch:
I digged in a little bit, my  experience from the the ubuntu forum tells me too to stay away from those software centers, but many users like and use  them, so I think we could add them to the applications but not to all components. I learned now why all those kde desktop files have those DNS-like desktop names, the [generator](https://www.freedesktop.org/software/appstream/metainfocreator/#/guiapp) creates those (`io.github.lxqt.creengrab.desktop for example`). Don't think we need this, but the standalone apps could all have their appdata.file

> but a user kindly made this for FeatherPad and I merged it

I looked at` /usr/share/metainfo/` -  it should be also there I learned.

tsujan:
> my experience from the the ubuntu forum tells me too to stay away from those software centers

I completely agree. Only `pacman` for Arch and its derivatives; only `apt` for Debian.

> I looked at /usr/share/metainfo/ - it should be also there I learned.

FeatherPad just has `/usr/share/metainfo/featherpad.metainfo.xml`. I won't add more.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html says, "component metadata of type desktop-application ... can be installed with an `.appdata.xml` extension as well **for historical reasons**" and also, "AppStream tools scan the /usr/share/appdata/ path **for legacy compatibility** as well. It should not be used anymore by new software though." So, I think that they may change their rules once in a while but, currently, `/usr/share/metainfo/XXX.metainfo.xml` is the right place.

tsujan:
I think `qterminal.metainfo.xml` (inside /usr/share/metainfo/) may be better than `qterminal.appdata.xml` as the file name. But, as I said, I'm not very familiar with these things.

tsujan:
It should be OK now, but please test it yourself.

stefonarch:
```
:metainfo$ appstreamcli validate qterminal.metainfo.xml  

? Validation was successful: pedantic: 3

```

stefonarch:
I made a test with inttools and it could be added in weblate, but this would need adding tools in lxqt-build-tools for the *.po files.
https://www.freedesktop.org/software/appstream/docs/sect-Quickstart-Translation.html
Not sure if it's worth the effort.

PR #902, opened 2022-02-15
Update README.md

stefonarch:
And another last one l10n...
Maybe a list
### Features
* Tabs
* Split
* etc

could be added too here.


tsujan:
> ... could be added too here.

I think it's better to merge this first.

Issue #904, opened 2022-02-20
Blank terminal is still possible


tsujan:
https://github.com/lxqt/qterminal/pull/898 fixed closing of the last tab but another case is still possible.

##### Steps to Reproduce (for bugs)

1. Check "Ask confirmation when closing".
2. Type `exit` and press `Enter`

Then the terminal gets blank.

##### Context

This is similar to https://github.com/lxqt/qterminal/issues/897 but without using the shortcut, tab button or title-bar's close button.
##### System Information

Latest git

PR #834, opened 2021-05-21
Adds extra options to confirm before exit window and close tabs

mhd-jalilvand:
Adds extra options to confirm before exit window and close tabs :
- [x] `0` Always ask before close the window
- [x]  `1` Never ask
- [x]  `2` Ony ask if the window contains more than one tabs
- [x]  `3` Ask confirmation not only on closing window but also for each individual tabs
- [x]  `4` show confirmation for main window and also for each tab which is splited to more than one session

related to #173 
maybe solves #746 
and solves #509

PR #906, opened 2022-02-20
Never show close prompt when explicitly ending session from terminal

tsujan:
Because, for example, it doesn't make sense to show a prompt after `ctrl-d`. Showing a prompt in this case would result in a blank view, that is prone to crash. The close prompt is for the GUI.

Fixes https://github.com/lxqt/qterminal/issues/904

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -592,7 +592,9 @@ void MainWindow::toggleBookmarks()
 void MainWindow::closeEvent(QCloseEvent *ev)
 {
     if (!Properties::Instance()->askOnExit
-        || !consoleTabulator->count())
+        || consoleTabulator->count() == 0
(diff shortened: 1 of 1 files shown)


tsujan:
@yan12125
I requested a review from you because I wanted to know whether you agree with the logic; otherwise, the change is very elementary.

@stefonarch 
As always, your tests are needed and appreciated.

stefonarch (approved):
It's the same behavior as with  `exit`-
GTM

yan12125:
Yes, the idea is good.

By the way, maybe https://github.com/lxqt/qterminal/pull/834 is no longer necessary with this fix?

stefonarch:
> By the way, maybe #834 is no longer necessary with this fix?

Yes and no, afaik there is  no possibility to ask for  confirm closing tab.

yan12125:
Got it, thanks for checking it!

Issue #905, opened 2022-02-20
Drop-down shortcut works even when terminal has a child dialog


tsujan:
##### Steps to Reproduce (for bugs)

1. Open a drop-down terminal.
2. Open its Preferences dialog.
3. Press the drop-down shortcut.

Then, the terminal disappears and only the dialog remains.

4. Close the dialog.

The terminal exits (no crash) because its last window is closed.

That's also bad because the user doesn't know that the terminal has exited. If he presses the shortcut, nothing will happen.

##### Context

This is an old problem.
##### System Information

Latest git.

tsujan:
Even worse, if you press the shortcut twice while Preferences dialog is shown, the dialog goes behind the terminal.

The solution is simple: the shortcut shouldn't work when there is a child dialog.

PR #907, opened 2022-02-20
Don't toggle drop-down terminal if it has modal dialog

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/905


stefonarch (approved):
Works perfect, thanks!

PR #910, opened 2022-02-21
Use a shortcut entry for drop-down shortcut

tsujan:
In this way, the user doesn't need to worry about key names.

The entry reacts to `Enter/Return` as well as `Escape` like other entries (spin-boxes and line-edits): `Enter/Return` applies the changes and closes the Preferences dialog, while `Escape` cancels everything and closes the dialog.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -953,19 +953,13 @@ To remove/disable a Shortcut, at point 2 press only a modifier (like Shift)</str
         </widget>
        </item>
        <item>
-        <layout class="QFormLayout" name="formLayout_2">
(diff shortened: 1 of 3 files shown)


stefonarch (commented):
schermata-02-21-13-18


Just a detail: could it be not so long? Similar size to the above?

stefonarch (approved):

tsujan:
> could it be not so long? 

Yes. I'll edit the ui file.

tsujan:
I played with the width a little by changing the layout, adding a horizontal spacer and trying different size policies. All in all, it wasn't worth the complication. Let's leave it as it was before this patch! Line-edits normally take the available width.

stefonarch:
Ok, it's not a setting that somebody will touch often anyway.

Issue #921, opened 2022-03-07
Improve interaction between dropdown and fullscreen


AKSoo:
##### Expected Behavior
* Activating fullscreen (default F11) should make a dropdown terminal fill the screen immediately.
* Pressing the shortcut should return the terminal to the user's dropdown width/height.

##### Current Behavior
* After activating fullscreen, you need to close and reopen the dropdown to make it fill the screen.
* Deactivating a fullscreen dropdown returns the terminal to the last saved **window** sizes, not the dropdown width/height.

##### Context
Would allow users to go in and out of fullscreen from dropdown

##### System Information
* QTerminal version: 1.0.0-1
* Distribution & Version: Arch Linux
* Kernel: Linux 5.15.26-1-lts
* Qt Version: 5.15.3+kde+r132-1

tsujan:
> Activating fullscreen (default F11) should make a dropdown terminal fill the screen immediately.

It does, with the latest git.

> Pressing the shortcut should return the terminal to the user's dropdown width/height.

If by shortcut you mean fullscreen shortcut, it does, with latest git.

Recently, the master branch received a commit that fixed several issues in the drop-down mode. Of course, it doesn't exist in 1.0.0.

However, I don't close this report because I found a minor issue in using the fullscreen shortcut with the drop-down mode. It isn't mentioned in the report but I'll explain (and probably, fix) it later if I find the time.

tsujan:
I've described the issue in the PR's comment: https://github.com/lxqt/qterminal/pull/922

PR #922, opened 2022-03-07
Fixed a minor problem in full-screening in drop-down mode

tsujan:
This small patch fixes the following issue in the drop-down mode:

 1. The window became fullscreen on pressing the fullscreen shortcut, as it should.
 2. After that, the window disappeared on pressing the drop-down shortcut, as it should.
 3. The window appeared with the correct size on pressing the drop-down shortcut again, as it should.

Then, the window wouldn't become fullscreen unless you pressed the fullscreen shortcut twice.

Closes https://github.com/lxqt/qterminal/issues/921

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -762,9 +762,18 @@ void MainWindow::showHide()
     }
 
     if (isVisible())
+    {
         hide();
+    }
     else
     {
+        // The checked state of the fullscreen action should be reset; otherwise, its shortcut
+        // might need to be pressed twice later to make the window fullscreen. We don't consult
+        // "isFullScreen()" because it will return "false" if the window has been deactivated.
+        if (auto a = actions.value(QLatin1String(FULLSCREEN)))
+        {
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch, I'm sorry for bothering you with testing PRs one after the other! But I'll keep doing it ;) Because your tests are really needed. There's no rush though.

tsujan:
OK. You could also show the menubar and do fullscreening from there. After 3, the fullscreen menu item should be unchecked, while it was checked before the patch.

stefonarch:
I saw that a shortcut can be set, so F11 works now (could be default?).
So I can confirm the issue and that it's fixed now.

stefonarch (approved):
Gtm

tsujan:
Thanks!

> so F11 works now (could be default?).

I thought it was default; I don't remember touching it because I don't like full-screening.

stefonarch:
Same, I never use it fullscreen. We should set up some counter for bugfixes here ;)

Issue #928, opened 2022-03-15
Problem with bookmarks dock in drop-down mode


tsujan:
##### Steps to Reproduce (for bugs)

1. Show bookmarks dock in drop-down mode.
2. Close bookmarks dock.

Then, the terminal view won't have focus. You should first click in it.

##### Possible Solution

Will look into it.

##### System Information

* Latest git

PR #929, opened 2022-03-16
Fixed focus change on showing/hiding bookmarks dock

tsujan:
Previously,

 * The focus didn't return to the drop-down terminal on hiding the bookmarks dock.
 * The detached dock wasn't activated when it was shown by using the shortcut (in contrast to detaching with mouse).
 * The main window wasn't activated after triggering a bookmark from a detached dock.

Fixes https://github.com/lxqt/qterminal/issues/928

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -582,10 +582,6 @@ void MainWindow::showFullscreen(bool fullscreen)
 void MainWindow::toggleBookmarks()
 {
     m_bookmarksDock->toggleViewAction()->trigger();
-    if (m_bookmarksDock->isVisible())
-    {
-        m_bookmarksDock->widget()->setFocus();
-    }
 }
 
 
@@ -852,6 +848,10 @@ void MainWindow::newTerminalWindow()
 
 void MainWindow::bookmarksWidget_callCommand(const QString& cmd)
 {
+    if (m_bookmarksDock->isFloating())
+    {
+        activateWindow();
+    }
     consoleTabulator->terminalHolder()->currentTerminal()->impl()->sendText(cmd);
     consoleTabulator->terminalHolder()->currentTerminal()->setFocus();
 }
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
Would you please confirm that the PR does its job?

stefonarch:
I never hide it, so I didn't notice that .Pr works as expected.
GTM

tsujan:
Thanks!

> I never hide it

I always do, after using it ;)

stefonarch:
I'm not sure when this started, but now with bookmark docks always open hiding/showing (F12) the focus will be  on the bookmarks panel always on reopen.

tsujan:
> ... the focus will be on the bookmarks panel filter always when reopen.

You mean that I had to make an exception for startup? Because this patch gives the focus to the bookmarks dock when it appears.

Making an exception for startup makes sense to me; will look into it.

stefonarch:
It's true also for startup, but real annoying is the show/hide toggle.

tsujan:
> but real annoying is the focus switch with show/hide toggle.

That was *the purpose*. When the user shows the bookmarks dock explicitly, he wants to *use* it. Why should the focus remain on the main view?

tsujan:
I also found a related crash bug and fixed it:

 1. Show the bookmarks dock.
 2. type `exit` inside the view.

Then the app will crash.

I'll make a PR. But please first tell me whether I am missing something because, when I consider the UX, I think that:

 1. The bookmarks dock should get the focus if the user shows it explicitly, by using its shortcut.
 2. The view should get the focus if the window is started with the bookmarks dock.

stefonarch:
> 2, type `exit` inside the view.

If you mean in the filter line edit I cannot reproduce it.

tsujan:
> If you mean in terminal I cannot see see the difference.

You don't see a difference probably because the crash happens on exiting. You'll see it if you open a terminal by entering `qterminal` in another (without https://github.com/lxqt/qterminal/pull/936, of course).

Anyway, the crash happened because, after entering `exit`, there was no "focus proxy" (a Qt concept). I added a nullity check in  https://github.com/lxqt/qterminal/pull/936.

tsujan:
BTW, I found another old issue (not a crash) in the drop-down mode. Most users don't (and didn't) encounter it but I'll try to fix it before 1.1.0.

PR #930, opened 2022-03-17
Fixed updating of shortcuts on tab switching/removal/creation

tsujan:
It's enough to update shortcuts on tab switching; tab removal and creation are automatically covered.

This PR supersedes and completes https://github.com/lxqt/qterminal/pull/666, which didn't cover problems in tab switching like the following:

 1. Split terminal.
 2. Open a new tab by using shortcut.
 3. Switch to the new tab and back to the previous one by clicking on tabs.

Then, before the current patch, shortcuts for sub-terminal switching didn't work until the Actions menu-bar item was shown.


stefonarch (approved):
Issue is fixed.

Issue #413, opened 2018-04-25
Right click over empty area in tab bar misbehaves


aramazano:
When right clicked on such an area, context menu still pops up. I assume it pops up on behalf of the foreground session, as it is the only explanation I can find as to why it should pop up in the first place.

Then, according to the action taken, it behaves this way:

* Close session
Nothing happens.

* Rename sessionn
a) If entered NULL then nothing happens.
b) If anything (including space) is entered, then segfaults.

* Change title color
Nothing happens.

I would suggest either disabling right click menu when the pointer is hovering over unoccupied tab bar space, or popping it up on behalf of the foreground tab (even if the pointer is not on it).

Thanks and regards.

##### System Information
* Distribution & Version: lxqt.debian.net experimental-snapshots on top of Debian Testing
* Kernel: 4.9.0-4-amd64
* Qt Version: 5.10.1
* libqtxdg Version: libqt5xdg3 / 3.1.1~13-gb5ddbd4-1
* qtermwidget Version: 0.8.1~24-g4cc05f9-1
* lxqt-build-tools Version: Not installed
* Package version: 0.8.1~36-g1a4452a-1 (About dialog: 0.8.0)

tsujan:
That context menu is so buggy that can cause a crash when shown outside all tabs:
```
#0  0x00007fe07482b68a in QObject::setProperty(char const*, QVariant const&) () at /usr/lib/libQt5Core.so.5
#1  0x000055a77d1c6c90 in  ()
#2  0x000055a77d1e88ab in  ()
#3  0x00007fe074823616 in QMetaObject::activate(QObject*, int, int, void**) () at /usr/lib/libQt5Core.so.5
#4  0x000055a77d1d02ad in  ()
#5  0x00007fe0755b1fab in QWidget::event(QEvent*) () at /usr/lib/libQt5Widgets.so.5
#6  0x00007fe075570fec in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt5Widgets.so.5
#7  0x00007fe075579770 in QApplication::notify(QObject*, QEvent*) () at /usr/lib/libQt5Widgets.so.5
#8  0x00007fe0747f2cf0 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt5Core.so.5
#9  0x00007fe0755cca82 in  () at /usr/lib/libQt5Widgets.so.5
[...]

PR #931, opened 2022-03-17
Fixed right clicking an empty area of tabbar

tsujan:
Previously, it would result in a crash if the Rename action of the context menu was clicked and wouldn't do anything in the case of other actions.

Fixes https://github.com/lxqt/qterminal/issues/413


tsujan:
Thanks for both reviews!

stefonarch:
Thanks for fixes all those bugs! I think you need a day to write the changelog ;)

tsujan:
Ha ha... Writing changelogs is easy but boring. I'm happy that it's needed only twice a year ;)

PR #933, opened 2022-03-19
Consider Enter from numeric pad in shortcut editors

tsujan:
We want to apply the shortcut with Enter/Return but Enter from numeric pad is accompanied by `KeypadModifier`. The patch considers that.

Previously, if Enter was used from the numeric pad, the shortcut editors would accept it as a shortcut, which was bad.

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -58,7 +58,9 @@ bool Delegate::eventFilter(QObject *object, QEvent *event)
         QKeyEvent *ke = static_cast<QKeyEvent *>(event);
         int k = ke->key();
         // commit data and close the editor with Enter/Return
-        if (ke->modifiers() == Qt::NoModifier && (k == Qt::Key_Return || k == Qt::Key_Enter)) {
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
If you have Enter in the numeric pad of your keyboard, please test it. If not, tell me; we need this in 1.1.0.

stefonarch:
I've enter in the Numpad, but I've mapped it to equal (=) for krunner's calculator, will undo that and test.

stefonarch:
Ok, I experienced now the bug... couldn't "enter" the branch to compile, had first to unset the shortkey...

GTM

tsujan:
> couldn't "enter" the branch to compile, had first to unset the shortkey...

Yes, we shouldn't accept `Enter` as a shortcut because it's for confirming a shortcut in the editor. However, `Enter` will be acceptable as a shortcut if it's accompanied by modifiers (like `Shift+Enter`). `Return` didn't have this problem but it's also covered for the sake of certainty.

Thanks!

Issue #932, opened 2022-03-19
Suspicious code: hidden `QWidget`


tsujan:
In `mainwindow.cpp`, the hidden widget `settingOwner` is created, with an empty geometry, to be the parent of actions (with shortcuts). This choice of parent is not only very unusual but also unneeded. Moreover, it may cause unpredictable troubles. We had another instance of such a code that was the source of several bugs and was fixed in https://github.com/lxqt/qterminal/commit/004f1ddb8ca828399aa6f9b56b608199f9dddc8b.

I opened this issue as reminder. IMO, QTerminal's actions code needs be rewritten properly.

PR #896, opened 2022-01-31
No hidden parent for dropdown window

tsujan:
The code had made the dropdown window be the child of a hidden tool window. I have no idea why the original coder thought that a main window needed a hidden parent. However, for some reason (perhaps deep in Qt code), that hidden parent caused trouble with sub-dialogs of the properties dialog and made the app exit after closing them. So, it's removed here.

Also, a memory leak is fixed in showing the properties dialog — previously, it wasn't destroyed until the app exited.

Fixes https://github.com/lxqt/qterminal/issues/894

Issue #358, opened 2017-09-29
Right Clicking on a link in QTerminal makes "Open Link" and "Copy Link Address" Appear but go away


tsimonq2:
The title explains it all, [here's](https://youtu.be/_AoMEhoVhfg) a quick unlisted YouTube video that shows it in real action.

I'm using my daily PPA which is the tip of the QTerminal master tree.

tsimonq2:
This is going to be a tricky one, I don't think it's a regression in QTerminal but rather QTermWidget... I'll leave this up to the people who know the codebase a bit better than I.

yan12125:
Here's a possible fix: https://github.com/lxde/qtermwidget/tree/wip/issue358 with https://github.com/lxde/qterminal/tree/wip/issue358. I don't think it's ready for a pull request...

The cause is: when the image (contents in the terminal) changes, all filter hotspots (e.g. recognized URLs) are deleted, and Open & Copy actions use filter objects as parents, so they are deleted, too.

The idea of my fix is: use the popu menu as parents of Open & Copy actions and defer deletion of filter objects. This is achieved by recording that a filter object is used by someone else (```_hasAnotherParent```). When both Open & Copy actions are deleted, the relevant filter object is deleted, too. Here I use ```std::shared_ptr``` to record whether a filter object is really no longer used by anyone or not.

My test case:
1. Open a text file with a URL via NeoVim. I have ```set mouse=a``` in my ```~/.config/nvim/init.vim```
2. Right click on the URL and choose either Copy Link Address or Open Link.

My checks:
* Open & Copy actions appear and gone before patching, and they're working fine after the patch
* valgrind doesn't complain about illegal access or memory leaks about filters

BTW, Konsole just don't allow me to right click a URL in NeoVim. I was trying to find an elegant solution there :(

tsimonq2:
Bump; I'm willing to put a bounty on this too if it would help.

I'm personally affected by this and I keep getting user reports about it.

yan12125:
@tsimonq2: I created pull requests from my last attempt. I'm not 100% sure those changes are correct. Please test it, thanks!

agaida:
https://youtu.be/nlRjQ9Gc8E0 - not sure why there is a bug - can't reproduce it

yan12125:
This bug occurs is terminal texts are modified after the click. For example in vim/neovim with `:set mouse=a`.

agaida:
ahhh - ok - thanks for the clarification

[...]

PR #934, opened 2022-03-19
Don't use a hidden `QWidget` as the parent of actions

tsujan:
A hidden `QWidget` isn't a good choice as the parent of actions and may cause unpredictable problems. `QObject` should be used instead of it.

Closes https://github.com/lxqt/qterminal/issues/932, for now.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -150,8 +150,7 @@ void MainWindow::rebuildActions()
 {
     // Delete all setting-related QObjects
     delete settingOwner;
-    settingOwner = new QWidget(this);
-    settingOwner->setGeometry(0,0,0,0);
+    settingOwner = new QObject(this);
 
     // Then create them again
     setup_FileMenu_Actions();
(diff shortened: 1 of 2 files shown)


tsujan:
@yan12125
Any objection to merging this?

yan12125:
Hmm, I asked a question about QWidget and QObject 5 years ago, and I'm not sure if I understand the response - https://github.com/lxqt/qterminal/pull/313#discussion_r107137479.

tsujan:
> ... and I'm not sure if I understand the response

That response doesn't make sense to me. Moreover, several weird bugs were caused by a similar parenting (https://github.com/lxqt/qterminal/pull/896); they weren't fixed for a long time because no one suspected that a hidden parent QWidget could have side effects.

yan12125:
Sounds good. Let's move on!

tsujan:
IMHO, the current way of creating/updating/deleting actions is problematic and could (or should) be replaced by a robust method later. But, for now, avoiding a hidden parent widget may be the best option.

I've also seen other parts of the code, where a class assumes certain things about another class. Although they work OK, they aren't safe, especially if some of them are changed for fixing bugs or adding features. We might need to rewrite them too.

yan12125:
> the current way of creating/updating/deleting actions is problematic

Yeah, that reminds me of https://github.com/lxqt/qterminal/issues/358 :sweat_smile:

PR #936, opened 2022-03-27
Give focus to bookmarks dock only if it's shown explicitly

tsujan:
Also, a crash is fixed on using `exit` or `ctr+d` while the bookmarks dock is shown.


stefonarch (approved):
GTM, didn't find any  issues now with focus handling.

tsujan:
Thanks!

PR #979, opened 2022-08-01
adding a new knob to the bookmark xml called newtab

mc36:
once this newtab is set, the given command will be executed on a new tab
the rationale here is that is speeds up commection manager use cases:
one just opens the bookmarks and clicks through the needed boxes,
and don't have to deal with the creation of new tabs for each...

### DIFF
--- a/src/bookmarkswidget.cpp
+++ b/src/bookmarkswidget.cpp
@@ -33,9 +33,10 @@ public:
         Command = 2
     };
 
-    AbstractBookmarkItem(ItemType type, AbstractBookmarkItem* parent = nullptr)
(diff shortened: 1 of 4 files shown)


yan12125:
Thanks, this is a useful feature! The code looks good. Could you also update [the bookmarks example](https://github.com/lxqt/qterminal/blob/master/qterminal_bookmarks_example.xml) and [the default bookmarks](https://github.com/lxqt/qterminal/blob/11489b08af48981f762335045f27d20c5f600455/src/propertiesdialog.cpp#L555) to show how to use this feature?

PR #1002, opened 2022-10-26
Updated *.ts files

stefonarch:


stefonarch:
I've no clue why some translations are included here, french is ok because of that nonbreaking space.

yan12125:
> I've no clue why some translations are included here,

Hmm which commands did you use? Probably some command pulls the latest translations from Weblate.

> french is ok because of that nonbreaking space.

Looks like another case of https://github.com/lxqt/pcmanfm-qt/pull/1498#issuecomment-981602181?

stefonarch:
Not really, we just learned some month ago that weblate changes all the correct spaces coming from Qt's transupdate when a translations is done. 
So I try to run `transupdate` on all just before the release.

> Hmm which commands did you use? Probably some command pulls the latest translations from Weblate.

I made a fresh checkout, same result running only `lxqt-transupdate`

yan12125:
Looking into Chinese translations, I guess Qt copies the translated string from

https://github.com/lxqt/qterminal/blob/14399e1bf1aaa335f02e16460d0d4cdc007f3b39/src/translations/qterminal_zh_TW.ts#L90-L91

to

https://github.com/lxqt/qterminal/blob/14399e1bf1aaa335f02e16460d0d4cdc007f3b39/src/translations/qterminal_zh_TW.ts#L749-L750

as both have the same source string. It's actually a nice feature :)

> Not really, we just learned some month ago that weblate changes all the correct spaces coming from Qt's transupdate when a translations is done.

Oh, I see.

stefonarch:
I think we can trust Qt here, merging.

PR #1014, opened 2022-11-24
Specified the parent of main context menu for use on Wayland

tsujan:
On Wayland, a parentless context menu might pop up in an incorrect position and even have a window decoration if the window is inactive when right clicked.

### DIFF
--- a/src/termwidget.cpp
+++ b/src/termwidget.cpp
@@ -137,7 +137,7 @@ void TermWidgetImpl::propertiesChanged()
 
 void TermWidgetImpl::customContextMenuCall(const QPoint & pos)
 {
-    QMenu menu;
+    QMenu menu(this);
     QMap<QString, QAction*> actions = findParent<MainWindow>(this)->leaseActions();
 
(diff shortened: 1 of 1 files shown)


yan12125:
Ownership of context menus has been a rabbit hole in qterminal and causes issues like https://github.com/lxqt/qterminal/issues/358. As this change makes sense and does not make existing issues worse on either X11 or Wayland (tested on weston), I tend to merge it.

tsujan:
> ... issues like https://github.com/lxqt/qterminal/issues/358

I hadn't seen it before; will look into it if I find the time.

> ... Menus seem fine.

Yes. The ownership just may help with positioning (in Qt's source).

yan12125:
tsujan ***@***.***> 於 2022年11月26日 週六 00:30 寫道：

> ... issues like #358 <https://github.com/lxqt/qterminal/issues/358>
>
> I hadn't seen it before; will look into it if I find the time.
>
Thanks! There is no hurry as workarounds exist.

PR #1016, opened 2022-12-03
Improve User Experience, change default values properties.cpp

lopezac:
* Change default useCWD value from false to true.

useCWD is the "Open new terminals in current working directory" option.

I think it improves the user experience, because it pissed me off, working in the terminal, to open a new tab, and have to move all the way to the current working folder, and do that for every new tab. That is the default behavior a user wants, to open new tabs in the current working directory.


yan12125:
That makes sense, thanks!

Issue #355, opened 2017-09-17
ASCII BEL support


silversquirl:
It'd be nice if QTerminal handled ASCII BEL in a similar way to most terminals (audible beep). Currently, highlights on IRC, etc. can go unnoticed for quite some time because of the lack of feedback.

This might be more applicable to QTermWidget. Let me know if you'd like me to recreate this issue there.

jleclanche:
It's fine in either, a PR would be appreciated though, QTerminal doesn't really have active development.

yan12125:
I got a quick-and-dirty proof of concept at https://github.com/lxde/qterminal/tree/wip-support-bell. You'll need to install libcanberra (possibly as well as libcanberra-pulse if you're using PulseAudio) and rebuild qterminal.

By the way, I like the bell sound from Arch's default sound-theme-freedesktop :)

testote:
Maybe it is time to commit this?

sudo -B is really waiting for it.

yan12125:
Oops, forgot about this feature after initial testing. With some refinements, I think it's ready: https://github.com/lxqt/qterminal/pull/1020. Testing are welcome!

PR #1020, opened 2022-12-25
Handle bell (BEL, '\a') via libcanberra

yan12125:
Tested with `echo -e "\07"`

Closes https://github.com/lxqt/qterminal/issues/355

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -3,7 +3,7 @@ set -ex
 source shared-ci/prepare-archlinux.sh
 
 # See *depends in https://github.com/archlinuxcn/repo/blob/master/archlinuxcn/qterminal-git/PKGBUILD
-pacman -S --noconfirm --needed git cmake lxqt-build-tools-git qt5-tools qtermwidget-git qt5-x11extras qt5-translations
+pacman -S --noconfirm --needed git cmake lxqt-build-tools-git qt5-tools qtermwidget-git qt5-x11extras qt5-translations libcanberra
 
 cmake -B build -S .
 make -C build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,8 @@ if (BUILD_TESTS)
 endif()
 
 include(LXQtPreventInSourceBuilds)
+include(FindPkgConfig)
+pkg_check_modules(LIBCANBERRA libcanberra)
 include(LXQtTranslateTs)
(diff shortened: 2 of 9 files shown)


tsujan:
Works as intended.

tsujan:
Nice feature, BTW. My terminal isn't silent anymore :)

yan12125:
By the way, I'm not sure if people feel terminal bells noisy. Pulseaudio allows disabling sounds from specific applications (ex: qterminal), and I believe other sound servers provide similar functionality. This should not be an issue, anyway.

Chiitoo:
> By the way, I'm not sure if people feel terminal bells noisy. Pulseaudio allows disabling sounds from specific applications (ex: qterminal), and I believe other sound servers provide similar functionality. This should not be an issue, anyway.

Definitely something I (and others I know) have on a checklist to disable when installing something like an Ubuntu workstation (though I usually install LXQt and friends as well so I won't be using the GNOME things in any case).

My personal machines use plain ALSA only, so silencing a single application isn't that simple, so it would certainly be nice to have an option to disable this feature (while 'libcanberra' may be installed for other reasons).

tsujan:
If an option for disabling it is going to be added, please do it after merging this PR. IMHO, it's good to keep this PR only for its implementation.

yan12125:
Thanks for the information about ALSA. I will add an option after this PR.

luis-pereira (approved):
It works Ok. The code is Ok.
GTM.

luis-pereira:
Wrote https://github.com/lxqt/qterminal/pull/new/add-audible-bell-option for enabling/disabling audible bell.

yan12125:
Thanks! Could you push new commits to this branch (wip-support-bell) instead? It's better to merge them together.

luis-pereira:
> Thanks! Could you push new commits to this branch (wip-support-bell) instead? It's better to merge them together.

Done. Force pushed due to changes in my commit message.
Consider removing the wip from the branch name,

yan12125:
Thanks, your part looks good and it works as expected.

A side note: `<tabstop>dropShortCutEdit</tabstop>` in the UI file is no longer relevant after https://github.com/lxqt/qterminal/pull/910.

> Consider removing the wip from the branch name,

Apparently GitHub does not allow changing the source branch for an existing pull request. The source branch name does not appear in a commit message with Rebase and merge or Squash and merge, anyway.

Issue #1026, opened 2023-01-20
Possible crash in `QxtGlobalShortcut` under Wayland


tsujan:
First, sorry for the bad report!

I found a short note on my Desktop about a QTerminal crash under Wayland. Most probably, I was busy with coding when the crash happened, so that I only had time to put these lines inside a note for future investigation:

```
#0  0x00007f76f6e1dde5 in XkbUseExtension () from /usr/lib/libX11.so.6
#1  0x00007f76f6e16167 in ?? () from /usr/lib/libX11.so.6
#2  0x00007f76f6e167e8 in XKeysymToKeycode () from /usr/lib/libX11.so.6
```

I don't remember how QTerminal crashed or whether I used its drop-down mode (I usually don't under X11, let alone under Wayland), but `XKeysymToKeycode`  is used in `qxtglobalshortcut_x11.cpp`, which is full of X11 codes. Has it been excluded when the app is used under Wayland?

EDIT: Oh, the name of the note was "qterminal drop wayland". So, for whatever reason, I might have launched the drop-down mode.

stefonarch:
How can you start it in dropdown mode? Here it  always dumps core.
```
$ qterminal -d
Segmentation fault (core dumped)
```
probably when trying lo load the shortcut


```
Core was generated by `qterminal -d'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f3b21ad5dbe in XKeysymToKeycode () from /usr/lib/libX11.so.6
[Current thread is 1 (Thread 0x7f3b1f810840 (LWP 665410))]
(gdb) where
#0  0x00007f3b21ad5dbe in XKeysymToKeycode () from /usr/lib/libX11.so.6
#1  0x000055969cb282c2 in ?? ()
#2  0x000055969cb27471 in ?? ()
#3  0x000055969caed57c in ?? ()
#4  0x000055969caee8b0 in ?? ()
#5  0x000055969caf5bd9 in ?? ()
#6  0x000055969caeae1a in ?? ()
#7  0x000055969caea5f4 in ?? ()
#8  0x00007f3b1fe3c290 in ?? () from /usr/lib/libc.so.6
#9  0x00007f3b1fe3c34a in __libc_start_main () from /usr/lib/libc.so.6
#10 0x000055969caeac25 in _start ()
(gdb) 

```

stefonarch:
To emulate dropdown I use in `wayfire.ini`
```
binding_qterminal = KEY_F12
[...]

PR #1027, opened 2023-01-27
Fix crash of dropdown mode under Wayland

yan12125:
Closes https://github.com/lxqt/qterminal/issues/1026

### DIFF
--- a/src/third-party/qxtglobalshortcut_x11.cpp
+++ b/src/third-party/qxtglobalshortcut_x11.cpp
@@ -36,6 +36,7 @@
  *
  ****************************************************************************/
 
+#include <QGuiApplication>
 #include <QVector>
 #include <QtX11Extras/QX11Info>
 
@@ -99,7 +100,7 @@ public:
 
     bool isValid()
     {
-        return m_display != nullptr;
+        return QGuiApplication::platformName() == QStringLiteral("xcb") && m_display != nullptr;
     }
 
     Display *display()


tsujan:
I'm not under Wayland for now, but if this change prevents the crash, it's preferable to disabling the drop-down mode. Moreover, a simple change that does a job is preferable in general.

stefonarch:
I tested under X11 and under wayland, having now a simple `qterminal -d` in session settings > autostart, one issue less. fopr session switching and autostart.

Looks like the window size in xx% yy% is  not respected under wayland, which is natural I guess.

tsujan:
> Looks like the window size in xx% yy% is not respected under wayland, which is natural I guess.

The size should be respected anywhere. Wayland ignores the position.

stefonarch:
Probably because it is a percentage of the screen it's not working, I can also resize it.

tsujan:
> Probably because it is a percentage of the screen

The screen size should be reported correctly under Wayland. Wayland is only the enemy of global positioning and activating.

stefonarch:
So maybe the method it get's screensize is only X11. I've 90% for both x/y and get window maximized or resized as I want under wayland with all compositors.

tsujan:
Yes, I also think the problem you see should be in the code but haven't found time to investigate.

tsujan:
> I've 90% for both x/y and get window maximized or resized

Are you sure you haven't a Wayland rule for setting the geometry?

stefonarch:
Good catch, I forgot about the rules ;)
And only the first window was `qterminal -d`, I didn't update my shortcut. I see it crash now with win+right click (resize). 

And I see an inconsistency (probably we have some issue already about that): modifying size for dropdown in normal window's pref is overwritten when` qterminal -d` exits.

stefonarch:
Tested on labwc, it doesn't crash on resize and doesn't respect the geometry.

tsujan:
>  I see it crash now with win+right click (resize).

The backtrace could be informative (after `where`).

stefonarch:
Often it crashes the whole session now. And no `where` possible:
```
g)                                                                                                                                
       Message: Process 665410 (qterminal) of user 1000 dumped core.
                
                Stack trace of thread 665410:
                #0  0x00007f3b21ad5dbe XKeysymToKeycode (libX11.so.6 + 0x93dbe)
                #1  0x000055969cb282c2 n/a (qterminal + 0x5d2c2)
                #2  0x000055969cb27471 n/a (qterminal + 0x5c471)
                #3  0x000055969caed57c n/a (qterminal + 0x2257c)
                #4  0x000055969caee8b0 n/a (qterminal + 0x238b0)
                #5  0x000055969caf5bd9 n/a (qterminal + 0x2abd9)
                #6  0x000055969caeae1a n/a (qterminal + 0x1fe1a)
                #7  0x000055969caea5f4 n/a (qterminal + 0x1f5f4)
                #8  0x00007f3b1fe3c290 n/a (libc.so.6 + 0x23290)
                #9  0x00007f3b1fe3c34a __libc_start_main (libc.so.6 + 0x2334a)
                #10 0x000055969caeac25 _start (qterminal + 0x1fc25)
                
                Stack trace of thread 665411:
                #0  0x00007f3b1ff1437f __poll (libc.so.6 + 0xfb37f)
                #1  0x00007f3b1f96dbef n/a (libglib-2.0.so.0 + 0xacbef)
                #2  0x00007f3b1f915132 g_main_context_iteration (libglib-2.0.so.0 + 0x54132)
                #3  0x00007f3b206d8b2c _ZN20QEventDispatcherGlib13processEventsE6QFlagsIN10QEventLoop17ProcessEventsFlagEE (libQt5Core.so.5 + 0x2d8b2c)
                #4  0x00007f3b206865ac _ZN10QEventLoop4execE6QFlagsINS_17ProcessEventsFlagEE (libQt5Core.so.5 + 0x2865ac)
                #5  0x00007f3b204e725f _ZN7QThread4execEv (libQt5Core.so.5 + 0xe725f)
                #6  0x00007f3b21bafcba n/a (libQt5DBus.so.5 + 0x23cba)
                #7  0x00007f3b204e432a n/a (libQt5Core.so.5 + 0xe432a)
                #8  0x00007f3b1fe9f8fd n/a (libc.so.6 + 0x868fd)
                #9  0x00007f3b1ff21d20 n/a (libc.so.6 + 0x108d20)
                
                Stack trace of thread 665412:
                #0  0x00007f3b1fe9c4b6 n/a (libc.so.6 + 0x834b6)
                #1  0x00007f3b1fe9ecd0 pthread_cond_wait (libc.so.6 + 0x85cd0)
                #2  0x00007f3b204eb7c4 _ZN14QWaitCondition4waitEP6QMutex14QDeadlineTimer (libQt5Core.so.5 + 0xeb7c4)
                #3  0x00007f3b1d122cc4 n/a (libQt5WaylandClient.so.5 + 0x72cc4)
                #4  0x00007f3b204e432a n/a (libQt5Core.so.5 + 0xe432a)
                #5  0x00007f3b1fe9f8fd n/a (libc.so.6 + 0x868fd)
                #6  0x00007f3b1ff21d20 n/a (libc.so.6 + 0x108d20)
                
                Stack trace of thread 665413:
                #0  0x00007f3b1ff1437f __poll (libc.so.6 + 0xfb37f)
                #1  0x00007f3b1d122d1d n/a (libQt5WaylandClient.so.5 + 0x72d1d)
                #2  0x00007f3b204e432a n/a (libQt5Core.so.5 + 0xe432a)
                #3  0x00007f3b1fe9f8fd n/a (libc.so.6 + 0x868fd)
                #4  0x00007f3b1ff21d20 n/a (libc.so.6 + 0x108d20)
                ELF object binary architecture: AMD x86-64

Cannot access "/var/lib/systemd/coredump/core.qterminal.1000.56e12f0358a34c04ab7052c034949a4b.665410.1674201766000000.zst": File o directory non esistente    
```

yan12125:
Hey, don't discuss window size issues under a pull request about shortcuts :)

stefonarch:
> Hey, don't discuss window size issues under a pull request about shortcuts :)

Short version: first it segfaulted on wayland, now it starts fine but could crash in some compositor  under some conditions :)

GTM for me.

tsujan:
Having logged into Wayfire, I can confirm that the crash is fixed. The size is correct too.

However, probably after this PR is merged, a patch is urgently needed for preventing the disappearance of the drop-down mode on losing the focus under Wayland, because, otherwise, the user will have no way of showing the window again (another `qterminal -d` will just open another instance, and the user might end up with several hidden instances by doing so).

I think that a fundamental solution, under both X11 and Wayland, is preventing a second instance of the drop-down mode by using D-Bus. Then, the hidden instance could be shown by `qterminal -d`.

tsujan:
@stefonarch, I think the issues you encountered might have been a result of having several hidden instances of the drop-down mode (see my comment above). Just a wild guess though.

stefonarch:
Partially yes: I see now that on the first meta+right click the window just disappears and gets invisible,  and then a second instance is freezing and crashing as described. But also with only one instance  running: if I try to drag  the border to resize the window the whole session crashes, which maybe a wayfire issue when trying to resize a fixed-size window.

tsujan:
> ...which maybe a wayfire issue when trying to resize a fixed-size window.

As soon as you try to resize a Qt window in Wayfire (and some X11 WMs as well), the window gets inactive. You may not notice that with a style like Fusion or Breeze, but it happens. Now, in the case of the drop-down QTerminal, the window also gets hidden, and that might trigger a bug in Wayfire, which doesn't expect the window to disappear on resizing.

Let's not make @yan12125 angry. He's right; we've polluted this page with our discussion — although it was very useful, IMO.

yan12125:
Gonna merge as https://github.com/lxqt/qterminal/pull/1027#issuecomment-1406561816 looks like an approval, and three devs (me included) have tested it.

For the size issue, please copy relevant notes here to a new issue or a discussion thread. They will be more useful if they can be easily found :)

yan12125:
(https://github.com/lxqt/qterminal/pull/1027#issuecomment-1407373035)
> a patch is urgently needed for preventing the disappearance of the drop-down mode on losing the focus under Wayland

Oops forgot about this. Will have a look when I find time.

stefonarch:
Disappearing is the feature, it can be solved like @tsujan said:

> I think that a fundamental solution, under both X11 and Wayland, is prevent a second instance of the drop-down mode by using D-Bus. Then, the hidden instance could be shown by `qterminal -d`.

I've a shortcut f12 under wayland which activates an existing qterminal window.

PR #1034, opened 2023-02-19
Fix license in qterminal.metainfo.xml

### LINKED ISSUE #1032: License specification conflict between qterminal.metainfo.xml and README.md
Problem noticed in the source code of QTerminal 1.2.0. I have verified that it still exists in the current Git master.

The README.md files specifies:

> This project is licensed under the terms of the [GPLv2](https://www.gnu.org/licenses/gpl-2.0.en.html) or any later version. See the LICENSE file for the full text of the license.

The qterminal.metainfo.xml file specifies:

> <project_license>GPL-2.0-only</project_license>

Whenever someone can get around to it, it would be great if one or the other of these files could be changed to match the license QTerminal is actually under. Because the README.md file seems to spell things out clearly, I assume that GPL-2+ is the correct license.

yan12125:
Closes #1032

### DIFF
--- a/qterminal.metainfo.xml
+++ b/qterminal.metainfo.xml
@@ -6,7 +6,8 @@
   <summary>A lightweight multiplatform terminal emulator</summary>
 
   <metadata_license>CC-BY-4.0</metadata_license>
-  <project_license>GPL-2.0-only</project_license>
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
GTM, thanks for looking so deeply in it!

PR #1037, opened 2023-03-15
Use qWarning().noquote() on warnings with escaped characters

isf63:
Since Qt 5.4 `qWarning() <<` prints escaped characters with the `\`. Using `qWarning().noquote()` prints them as intended.


luis-pereira:
@isf63 Thanks.

PR #1044, opened 2023-04-13
Fixed a crash in context menu

tsujan:
The context menu needs a parent for being positioned correctly under Wayland (→ https://github.com/lxqt/qterminal/commit/27f2e7ce0a2a1cd476fb4abe8b97888fe22f0e4f). But the parent shouldn't be `TermWidgetImpl` because, otherwise, the app will crash on clicking the action `SUB_COLLAPSE` (which deletes the parent before the menu is closed).

The patch prevents the crash by choosing the main window as the parent.

### DIFF
--- a/src/termwidget.cpp
+++ b/src/termwidget.cpp
@@ -137,8 +137,13 @@ void TermWidgetImpl::propertiesChanged()
 
 void TermWidgetImpl::customContextMenuCall(const QPoint & pos)
 {
-    QMenu menu(this);
-    QMap<QString, QAction*> actions = findParent<MainWindow>(this)->leaseActions();
+    auto mainWindow = findParent<MainWindow>(this);
+    if (mainWindow == nullptr)
+    {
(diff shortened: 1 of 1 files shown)


tsujan:
This is urgent. I tested it under X11 and Wayland alike.

stefonarch (commented):
Confirmed the  bug and on wayland this patch resolves it, to test x11

stefonarch (approved):
Fine also under x11.

yan12125:
At first glance it looks good. I don't have time to look into details these two days. Let's merge it now and add follow-up fixes in 1.3.1 if needed.

yan12125:
There is only one minor issue: `mainWindow == nullptr` may never happen (otherwise original codes will crash on both X11 and Wayland), so the check may be replaced with an assert.

tsujan:
> ... may be replaced with an assert.

Right.

PR #1058, opened 2023-06-29
Let QTermWidget library set TERM

luis-pereira:
Just pass the TERM variable value to QTermWidget. QTermWidget will set it.


yan12125:
It turns out this PR does not depend on https://github.com/lxqt/qtermwidget/pull/511. This PR itself is a good change, so let's move on!

PR #1060, opened 2023-07-07
Reorder propertiesdialog.ui XML to correct tab order

isf63:
Affected files:
- src/forms/propertiesdialog.ui

The other .ui files had correct tab order.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -14,16 +14,6 @@
    <string>Terminal settings</string>
   </property>
   <layout class="QGridLayout" name="gridLayout_4">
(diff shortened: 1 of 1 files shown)


yan12125:
Thank you for the work. Most changes look good. The only issue is that in the Dropdown page, the shortcut editor is focused after OK/Cancel/Apply. This is probably because the shortcut editor is added via C++ [1] instead of declared in `propertiesdialog.ui`. Let's merge this first and investigate that later if someone feels it annoying.

[1] https://github.com/lxqt/qterminal/blob/1.3.0/src/propertiesdialog.cpp#L249

PR #1078, opened 2023-10-17
Avoid null pointer dereference

luis-pereira:
If w is nullptr we must not dereference it.
qFatal is a cleaner solution.


yan12125:
That makes sense, thanks!

PR #1101, opened 2024-02-16
Add word characters preference option (fixes #1098)

### LINKED ISSUE #1098: [Feature request] Option to customize set of semantic selection split characters
**Is your feature request related to a problem? Please describe.**

It seems that the set in question is hardcoded to something like whitespaces and `=<>[]{}()#`, which makes doubleclick selections pretty greedy.

**Describe the solution you'd like**

An option in preferences to customize this set of characters.

**Describe eventual alternatives you've considered**

ahigerd:
gnome-terminal supports configuring word selection characters, but QTerminal doesn't. This PR adds a configuration option to address that shortcoming.

I was going to make a feature request, but I found [this feature request](https://github.com/lxqt/qterminal/issues/1098) that beat me to it. @yan12125 had a partial implementation that I referenced, but I discovered that it was going to require a change to QTermWidget as well.

QTermWidget dependency: https://github.com/lxqt/qtermwidget/pull/531

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -69,7 +69,7 @@
       <enum>QFrame::StyledPanel</enum>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
      </property>
      <widget class="QWidget" name="appearancePage">
       <layout class="QVBoxLayout" name="verticalLayout_4">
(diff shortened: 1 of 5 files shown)


yan12125:
Thanks!

ahigerd:
Whoops. I found a bug in this PR: When the preference is changed, it isn't saved to disk, so it only persists as long as QTerminal is running. I only noticed this because I rebooted yesterday.

I'll get a fix in later once I have time to sit down with it when I'm not at work.

PR #1106, opened 2024-02-20
save word characters preference to disk

ahigerd:
In [my previous PR](https://github.com/lxqt/qterminal/pull/1101) I mistakenly forgot to save the new preference to the settings.


yan12125:
Nice catch!

Issue #1111, opened 2024-04-25
Use layershell for dropdown terminal under wayland


stefonarch:
Use layershell for positioning and layer, similar to lxqt-runner and lxqt-leave dialog windows.

##### Expected Behavior

Dropdown settings for size are working under wayland and do not show it in taskbars.
##### Current Behavior

* Position has to be manually fixed for centering and top panel by window rules and will affect all qterminal windows, e.g.
`<action name="MoveTo" x="80" y="38"/>` under labwc.

* Dropdown terminal shows up in taskbar

##### Possible Solution

 
Use similar code as we do in runner already:
https://github.com/lxqt/lxqt-runner/blob/3deaf15726c59c90f4a53c2d4678799d79889e26/dialog.cpp#L206

##### Context

Should be among wayland improvements in  LXQt 2.1  
Related issues are https://github.com/lxqt/qterminal/issues/1028 and https://github.com/lxqt/qterminal/issues/1059.

With a working single mode `qterminal -d`  can be used to reopen the existing dropdown QTerminal.

##### System Information

* QTerminal version: qt6

marcusbritanicus:
@stefonarch I think this issue can be closed, now that #1113 is merged.

PR #1113, opened 2024-04-25
Set layer overlay for dropdown terminal on wayland

stefonarch:
Fix for https://github.com/lxqt/qterminal/issues/1111
 
Remaining issues:
- [ ]  Configuration window is below the dropdownterminal and eventually completely unaccessible
- [ ] Consider screen width instead of available width for centering. With a 100% size and a right or left panel it goes out of screen (for the panel-width).
- [ ] Add an info and disable not-working shortcut selection in preferences.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,11 +24,13 @@ endif()
 set(LXQTBT_MINIMUM_VERSION "2.0.0")
 set(QT_MINIMUM_VERSION "6.6.0")
 set(QT_MAJOR_VERSION "6")
+set(SHELLQT_MINIMUM_VERSION "6.0.0")
 
 find_package(Qt6Core ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6Gui ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6LinguistTools ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6Widgets ${QT_MINIMUM_VERSION} REQUIRED)
+find_package(LayerShellQt ${SHELLQT_MINIMUM_VERSION} REQUIRED)
 if(UNIX)
     find_package(Qt6DBus ${QT_MINIMUM_VERSION} REQUIRED)
     find_package(Qt6 COMPONENTS Core REQUIRED)
@@ -206,6 +208,7 @@ target_link_libraries(${EXE_NAME}
     Qt6::Gui
     Qt6::Widgets
     qtermwidget6
+    LayerShellQtInterface
 )
 if(QXT_FOUND)
     target_link_libraries(${EXE_NAME} ${QXT_CORE_LIB} ${QXT_GUI_LIB})
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -44,6 +44,9 @@
 #include "qterminalapp.h"
 #include "dbusaddressable.h"
 
+#include <LayerShellQt/Shell>
+#include <LayerShellQt/Window>
+
 typedef std::function<bool(MainWindow&, QAction *)> checkfn;
 Q_DECLARE_METATYPE(checkfn)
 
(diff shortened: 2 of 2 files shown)


marcusbritanicus:
This works great..!! Two problems I face at the moment. 
1. Multiple instances.
2. Toggling the terminal window.

**Solution for 1.**
We can use a single-instance class derived from QApplication. When starting in drop-down mode, we can lock the instance. In all other cases, leave it unlocked.

**Solution for 2.**
Call `qterminal -d` to toggle and start the drop-down mode. Since the first instance is already locked, the subsequent instances can request the fist instance to toggle the window. All non-drop-down instances will not be affected by this since they're not locked.

stefonarch:
There are already open issues for that - listed in the the issue which will be fixed. At the moment I've a[ rather hacky script ](https://github.com/stefonarch/lxqt-labwc-session/blob/lxqt2-labwc-session/toggledropdown.in)which I use for the shortcut.

marcusbritanicus:
> There are already open issues for that - listed in the the issue which will be fixed. At the moment I've a[ rather hacky script ](https://github.com/stefonarch/lxqt-labwc-session/blob/lxqt2-labwc-session/toggledropdown.in)which I use for the shortcut.

Well.. I ma working on this now. Let's see how it goes.

stefonarch:
Testing combined with https://github.com/lxqt/qterminal/pull/1115 the issue with the settings window below is less severe as users could use F12 again to remove the terminal. Ideally doing this in the code would best IMO.

Regarding the second listed issue: it is exactly half panel width which moves the terminal out of center or out of screen if set to 100% width. 
With 2 identical panels on both sides it turns centered.

marcusbritanicus:
> Regarding the second listed issue: it is exactly half panel width which moves the terminal out of center or out of screen if set to 100% width. With 2 identical panels on both sides it turns centered.

Panel should be a `TopLayer`. Since QTerminal will be `OverlayLayer`, there should not be any conflict at all. Is this behaviour with all compositors (including kwin) or only a particular one?

stefonarch:
>Is this behaviour with all compositors (including kwin) or only a particular one?

I saw exactly the same with kwin too - actually it was kwin and another user where I played with 100px panels adding and removing (and hating the jumping icon at cursor at F12 press).  In theory you're right, it may be a bug in layershell-qt.

stefonarch:
Added another issue to fix:
 >Add an info and disable not-working shortcut selection in preferences.

tsujan:
The conflict with the current master aside, the combination of of this and https://github.com/lxqt/qterminal/pull/1131 resulted in a good experience here, under LabWC.

However, we may need to address some issues. For example, I noticed that the Preferences dialog goes behind the main window. I was able to work with it after clicking into the main window and then somewhere else to hide the main window, but that's not good for UX.

tsujan:
The problem with Preferences dialog seems to be a problem in Wayland (WM): it's the job of a WM to put a modal dialog in front of its parent window. So, we don't need to deal with it.

marcusbritanicus:
> The problem with Preferences dialog seems to be a problem in Wayland (WM): it's the job of a WM to put a modal dialog in front of its parent window. So, we don't need to deal with it.

We have multiple levels of problems here:
1. The main window is now a layer-surface, and an overlay surface. What this implies is that:
   a. It's not a xdg-shell window. So any windows that this layer-surface window spawns will be a simple, xdg-shell window.
   b. All windows that this layer-surface window spawns will be placed below this surface, because you have asked the compositor to place this on top.
2. Wayland, at the moment, does not have the concept of window modality. Even if it did, 1a and 1b would make it a moot point.

The solution to this is to hide the drop-down surface, and then show it again when the options dialog is closed.

Alternatively, show the options dialog as a layer-surface.

stefonarch:
Just noticed that `qterminal -d` and `lxqt-runner` on sway have the same issue: missing keyboard focus. Actually I don't remember if I already opened an issue about that.

tsujan (approved):
GTM

marcusbritanicus:
> @marcusbritanicus Thanks for your explanation! I get your point; I just don't think it's _our_ problem to solve. But yes, we may need a workaround, in which case, it could be added in another PR, after this is merged.

@tsujan Just a quick update: There is a xdg-dialog in wayland-protocols. It might take time for it to be implemented in Qt, but it _might_ solve our problem of the dialog going below. I do not expect that to directly work - it will require changes to layershell-qt.

tsujan:
> It might take time for it to be implemented in Qt, but it might solve our problem of the dialog going below.

Thanks for the info! indeed problems like that should be solved upstream.

Issue #1059, opened 2023-07-05
[Feature request] Only allow one instance of QTerminal dropdown per session


isf63:
**Is your feature request related to a problem? Please describe.**
It is currently possible to run multiple instances of QTerminal dropdown on the same session. This leads to unanticipated behavior.

**Describe the solution you'd like**
`qterminal --drop` does nothing when already running a dropdown instance, per X11/Wayland session. If none is running, it works as existing.

**Describe eventual alternatives you've considered**
Same, but show/focus the dropdown when an instance is already running.

**Context**
Extra instances of `qterminal --drop` in the same session are inaccessible, and only able to be exited via ending the PID.

**EDIT**: Other relevant flags to handle are `--execute`, `--profile`, `--workdir`. Those should only apply to a new instance.

yan12125:
There are also discussions about single-instance mode for the dropdown terminal at https://github.com/lxqt/qterminal/issues/1028. Probably two birds can be killed with one stone.

Issue #1028, opened 2023-01-29
QTerminal -d on wayland vanishes if loosing focus and can't be recovered


stefonarch:
As it fails to register the shortcut there's no way to make it visible again.

##### Possible Solution

1. Allow only one instance of `qterminal -d `(https://github.com/lxqt/qterminal/pull/1027#issuecomment-1407373035)
2.  or fix the registering of the shortcut

##### Steps to Reproduce (for bugs)

1. Open `qterminal -d` on wayland
2. change focus
4.  it will disappear also from taskbar like yatbfw

                                                 --->
* QTerminal version:
* Distribution & Version:
* Kernel:
* Qt Version:

tsujan:
IMHO, the drop-down mode is already used like a single-instance app; having multiple instances of it running together doesn't make much sense under X11 either.

yan12125:
The current qterminal already allows showing/hiding the dropdown window via a dbus method. For example:
```
qdbus org.lxqt.QTerminal-612287 / org.lxqt.QTerminal.Process.toggleDropdown
```
However, the service name is not fixed, as dbus interfaces in qterminal are designed to support multiple instances. Moving to a single-instance model is a big work and I don't think I can do it in near future. See also https://github.com/lxqt/qterminal/issues/644 for previous discussions about single-instance model.

stefonarch:
Could `dbus-send something` be used with a wildcard which always matches the dropdown terminal?

This could be used in a shortcut on all wayland compositors.
Looking for the clients class there are the same, would it be possibile to have a different class for the dropdown-window?

```
Window a8b59ab0 -> stef@archlinux: ~:
        at: 41,41
        size: 1283,998
        workspace: 3 (3)
        floating: 0
        monitor: 0
        class: qterminal
        title: stef@archlinux: ~
        pid: 2301858
        xwayland: 0
        pinned: 0
        fullscreen: 0
        fullscreenmode: 0
        fakefullscreen: 0
        grouped: 0
        swallowing: 0
-----------------------------
Window a8b59ab0 -> stef@archlinux: ~:
        at: 41,41
        size: 1283,998
        workspace: 3 (3)
        floating: 0
        monitor: 0
        class: qterminal
        title: stef@archlinux: ~
        pid: 2301858
        xwayland: 0
        pinned: 0
        fullscreen: 0
        fullscreenmode: 0
        fakefullscreen: 0
        grouped: 0
        swallowing: 0


```

yan12125:
[...]

PR #1114, opened 2024-04-25
Single instance dropdown

marcusbritanicus:
This PR adds preliminary support for single instance mode for drop-down mode. Additionally, it adds support for toggling the drop-down instance via `qterminal -d` command line. On wayland, we can simply add a command-binding pair for most compositors. On wayfire, this would simply be adding the following lines under the `[command]` section.

```
binding_qterminal_dd: KEY_F12
command_qterminal_dd: qterminal -d
```

If an instance of qterminal in the drop-down mode exists, it will toggle it. Otherwise, a new drop-down instance will be created.

Cc: @yan12125 @stefonarch @tsujan 

Edit:
This fixes both #1028 and #1059.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,11 +24,13 @@ endif()
 set(LXQTBT_MINIMUM_VERSION "2.0.0")
 set(QT_MINIMUM_VERSION "6.3.0")
 set(QT_MAJOR_VERSION "6")
+set(SHELLQT_MINIMUM_VERSION "6.0.0")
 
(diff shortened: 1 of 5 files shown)


stefonarch:
Very nice! But we should follow our rules: One issue - one commit/PR otherwise it's hard to keep track. For me single instance + `-d` option could go together though.

marcusbritanicus:
I'll close this PR and open a new one. I'm finding it a pain to rebase it on master.

PR #1115, opened 2024-04-25
Add preliminary single-instance support for drop-down mode

marcusbritanicus:
This PR adds preliminary support for single instance mode for drop-down mode. Additionally, it adds support for toggling the drop-down instance via `qterminal -d` command line. On wayland, we can simply add a command-binding pair for most compositors. On wayfire, this would simply be adding the following lines under the `[command]` section.

```
binding_qterminal_dd: KEY_F12
command_qterminal_dd: qterminal -d
```

If an instance of qterminal in the drop-down mode exists, it will toggle it. Otherwise, a new drop-down instance will be created. The normal qterminal instances are unaffected.

This fixes both #1028 and #1059. Since it's not possible to fix one without the other, I am breaking the one-issue-one-PR rule.

Cc: @yan12125 @stefonarch @tsujan

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,6 @@ endif()
 set(LXQTBT_MINIMUM_VERSION "2.0.0")
 set(QT_MINIMUM_VERSION "6.3.0")
 set(QT_MAJOR_VERSION "6")
-set(SHELLQT_MINIMUM_VERSION "6.0.0")
 
 find_package(Qt6Core ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6Gui ${QT_MINIMUM_VERSION} REQUIRED)
@@ -41,8 +40,6 @@ find_package(lxqt2-build-tools ${LXQTBT_MINIMUM_VERSION} REQUIRED)
 if (BUILD_TESTS)
     find_package(Qt6 ${QT_MINIMUM_VERSION} CONFIG REQUIRED Test)
 endif()
-
-find_package(LayerShellQt ${SHELLQT_MINIMUM_VERSION} REQUIRED)
 
 include(LXQtPreventInSourceBuilds)
 include(FindPkgConfig)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -24,11 +24,13 @@ endif()
 set(LXQTBT_MINIMUM_VERSION "2.0.0")
 set(QT_MINIMUM_VERSION "6.3.0")
 set(QT_MAJOR_VERSION "6")
+set(SHELLQT_MINIMUM_VERSION "6.0.0")
```
stefonarch, line 27: A leftover from the layershell PR

CMakeLists.txt
```diff
@@ -40,6 +42,8 @@ if (BUILD_TESTS)
     find_package(Qt6 ${QT_MINIMUM_VERSION} CONFIG REQUIRED Test)
 endif()
 
+find_package(LayerShellQt ${SHELLQT_MINIMUM_VERSION} REQUIRED)
```
stefonarch, line 45: Same

CMakeLists.txt
```diff
@@ -210,7 +216,9 @@ target_link_libraries(${EXE_NAME}
     Qt6::Core
     Qt6::Gui
     Qt6::Widgets
+    Qt6::Network
     qtermwidget6
+    LayerShellQtInterface
```
stefonarch, line 221: Same

stefonarch (changes requested):

CMakeLists.txt
marcusbritanicus: Thanks. It's fixed now.

stefonarch:
Works as it should - for me GTM

tsujan:
I'll review it after QTerminal 2.0 is released.

tsujan:
Although this code seems OK, I'm afraid not only a D-Bus-based single instance needs much less code (because D-Bus is already used), but also it's more reliable than the socket-lock method (I say that based on my experience with old versions of FeatherPad).

Here is the D-Bus-based PR: https://github.com/lxqt/qterminal/pull/1131

marcusbritanicus:
@tsujan Just tested #1131. It works great!! I'll close this.

stefonarch:
Works ok here too, but thanks to @marcusbritanicus 's code which made it already possible to have it for the last weeks ;)

Issue #1112, opened 2024-04-25
Regression in Qt6: inactive tab may have elided text


tsujan:
##### Steps to Reproduce (for bugs)

1. Unckeck Terminal Settings → Appearance → Fixed tab width
2. Open two tabs, such that the text of the active tab (in bold) is NOT elided.

Then the text of the inactive tab may be elided, while it's the same as the text of the active tab.

This happens with all Qt styles. A screenshot with Fusion:

![elided](https://github.com/lxqt/qterminal/assets/981076/a0f842a4-8fb1-4301-aa2f-22529c033479)

##### Context
I think it's about a change in Qt6.

##### System Information

Latest git version.

tsujan:
Apparently, in Qt6, `QStyle::sizeFromContents(QStyle::CT_TabBarTab, ...)` isn't called anymore for the inactive tab when there is a style sheet like "QTabBar::tab:!selected { font-weight: normal; }", as there is in our `tabbar.cpp`. Therefore, the width of the inactive tab is different from the active one, although they have the same text.

In short, our trick in the c-tor of `TabBar`  doesn't work with Qt6 (or at least with Qt ≥ 6.7, which I have here), while it worked with Qt5.

PR #1110, opened 2024-04-15
Improve qt6

luis-pereira:

PR #1116, opened 2024-04-25
Fixed inactive tab text with Qt6

tsujan:
Due to a change (or bug?) in Qt6, using a stylesheet for setting the font of active/inactive tabs doesn't work as it did with Qt5.

Here, `QStyle` is used directly, which is always the best approach because stylesheets can interfere with the widget style in bad ways.

Fixes https://github.com/lxqt/qterminal/issues/1112

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -17,19 +17,59 @@
  ***************************************************************************/
 
 #include "tabbar.h"
+#include <QPainter>
+#include <QStyleOption>
+#include <QProxyStyle>
+
+class TabTextStyle : public QProxyStyle
+{ // Give a normal font to inactive tabs.
+public:
+    using QProxyStyle::QProxyStyle;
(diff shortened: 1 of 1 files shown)


tsujan:
@yan12125 
This is needed before releasing Qt6 QTerminal.

luis-pereira:
Works as described with 6.6.2.

yan12125:
Thanks! Although I don't fully understand QStyle, I trust your expertise.

tsujan:
@yan12125,
Any estimation about when QTerminal 2.0.0 can be released? I haven't found any performance issue so far, but it may be better to silence compilation warnings before releasing.

yan12125:
I have some local changes for minor fixes and improvements, which should be finished in one or two days. Besides that, I'm waiting for https://github.com/lxqt/qterminal/pull/1110 and https://github.com/lxqt/qtermwidget/pull/540.

PR #1121, opened 2024-05-08
Obsolete

luis-pereira:


src/tabwidget.cpp
```diff
@@ -315,7 +315,7 @@ void TabWidget::removeFinished()
 {
     QObject* term = sender();
     QVariant prop = term->property(TAB_INDEX_PROPERTY);
-    if(prop.isValid() && prop.canConvert(QVariant::Int))
+    if(prop.isValid() && prop.canConvert<int>())
```
tsujan, line 318: `QMetaType::Int` ?

tsujan: Never mind, it's obsolete, and the current patch is good.

tsujan (approved):
GTM

PR #1122, opened 2024-05-12
Update build.sh

stefonarch:
Will still fail as `qtermwidget-git` doesn't exist yet.
EDIT: probably not true.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -3,7 +3,7 @@ set -ex
 source shared-ci/prepare-archlinux.sh
 
 # See *depends in https://github.com/archlinuxcn/repo/blob/master/archlinuxcn/qterminal-git/PKGBUILD
-pacman -S --noconfirm --needed git cmake lxqt-build-tools-git qt5-tools qtermwidget-git qt5-x11extras qt5-translations libcanberra
+pacman -S --noconfirm --needed git cmake lxqt-build-tools-git qt6-tools qtermwidget-git qt6-translations libcanberra
 
 cmake -B build -S .
 make -C build


yan12125:
As a record, I gave up maintaining LXQt *-git packages used in CI (https://github.com/archlinuxcn/repo/issues/3787). If you want to keep CI alive, you'll need to refresh -git packages first. See also https://github.com/lxqt/lxqt/issues/1644#issuecomment-2076343537 about concerns for the current approach.

stefonarch:
I didn't remember where to look for your comments, thanks.

tsujan:
> If you want to keep CI alive...

CI was annoying to me from the start, although I learned how to tolerate it ;) I may be wrong, but I don't remember a single case where it has been really useful.

yan12125:
> I don't remember a single case where it has been really useful.

It was useful when qtermwidget was ported to Qt 6 (https://github.com/lxqt/qtermwidget/pull/532). At that time nobody tested the port with PyQt, and CI became a good reminder.

Anyway, the current CI approach brings more issues than it solves.

stefonarch:
> but I don't remember a single case where it has been really useful

I remember that once I made a translation PR which went trough and blocked compiling, that was before weblate which prevents malformatted files. But as usually we compile all PRs locally we could do without CI.

 I admire the CI of labwc which checks compiling on I don't know how many platforms, but that's one small binary which compiles here in 15 seconds or less.

stefonarch:
Atm I disabled CI testing in QTerminal and lxqt-panel.

PR #1124, opened 2024-05-13
Silenced Qt6 compilation warnings

tsujan:
All of them were trivial.

Strangely enough, the file `qxtglobalshortcut.cpp` had the MS-Windows newline character, which was corrected by FeatherPad on saving. That's the cause of what you see on GitHub. Use a GUI diff tool (like kdiff3) to see the real difference.


stefonarch:
Looks all fine now, IMO we could release QTerminal Qt6 soon, I don't notice issues in my workflow.

tsujan:
> I don't notice issues in my workflow.

I neither. Users will tell us more after a release.

> IMO we could release QTerminal Qt6 soon

I think we should. I'll merge this and https://github.com/lxqt/qtermwidget/pull/549 tomorrow and prepare a release.

PR #1131, opened 2024-05-20
Single instance for drop-down mode

tsujan:
Based on D-Bus.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -41,6 +41,9 @@
 
 const char* const short_options = "vhw:e:dp:";
 
+static const char* serviceName = "org.lxqt.QTerminal";
+static const char* ifaceName = "org.lxqt.QTerminal.Process";
+
 const struct option long_options[] = {
     {"version", 0, nullptr, 'v'},
     {"help",    0, nullptr, 'h'},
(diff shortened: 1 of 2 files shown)


marcusbritanicus:
I can confirm that this solves both #1028 and #1059.

stefonarch:
Works fine, yes.

This has a nice side-effect: `qterminal -d` in x11 needs to be in autostart and opens the window at login, while on wayland the first hit on F12 opens it and if you actually close it by ctrl+q no issue, it reopens again at F12. So I added `X-LXQt-X11-Only=true` in it's autostart ` .desktop` file

tsujan:
@marcusbritanicus, @stefonarch, thanks for testing it! Will merge it along with https://github.com/lxqt/qterminal/pull/1113 after testing them together for a week or so.

@marcusbritanicus, as I mentioned at https://github.com/lxqt/qterminal/pull/1115, your method was also correct and worked fine. I didn't check why D-Bus was already used in the code, but I didn't find a *real* use for it by a glance before the current PR.

If needed, we could also make a single-instance app out of the windowed mode later. But I'm not sure if that's a good idea. Personally, I feel safer when my terminal instances are independent from each other. However, the drop-down mode was already used like a single-instance app.

@stefonarch,, please rebase https://github.com/lxqt/qterminal/pull/1113 to fix its conflict (only one) with the current master.

marcusbritanicus:
@tsujan No issues at all. This way is much shorter and it will be easy on the maintenance. Plus, no new files added.

luis-pereira:
Testing it.

Issue #1135, opened 2024-06-04
Crash on entering `exit` with bookmark dock shown


tsujan:
When the bookmark dock is shown, if `exit` is entered inside a single-tabbed window of QTerminal, it crashes instead of exiting.

##### Steps to Reproduce (for bugs)

1. Open QTerminal and show the bookmark dock.
2. Enter `exit` in QTerminal.

Then, QTerminal crashes here, with this backtrace:

```
#0  0x00007f214493991c in QWidgetPrivate::deepestFocusProxy() const () at /usr/lib/libQt6Widgets.so.6
#1  0x00007f21449588a3 in ??? () at /usr/lib/libQt6Widgets.so.6
#2  0x00007f21437a17b7 in ??? () at /usr/lib/libQt6Core.so.6
#3  0x00007f2144a6b242 in QDockWidget::event(QEvent*) () at /usr/lib/libQt6Widgets.so.6
#4  0x00007f21448fc55c in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#5  0x00007f2143744e08 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#6  0x00007f2144945951 in QWidgetPrivate::hideChildren(bool) () at /usr/lib/libQt6Widgets.so.6
#7  0x00007f2144945f02 in QWidgetPrivate::hide_helper() () at /usr/lib/libQt6Widgets.so.6
#8  0x00007f2144949a0a in QWidgetPrivate::setVisible(bool) () at /usr/lib/libQt6Widgets.so.6
#9  0x00007f214494678a in QWidgetPrivate::handleClose(QWidgetPrivate::CloseMode) () at /usr/lib/libQt6Widgets.so.6
#10 0x00007f2144963127 in ??? () at /usr/lib/libQt6Widgets.so.6
#11 0x00007f2143fdf300 in QWindow::event(QEvent*) () at /usr/lib/libQt6Gui.so.6
#12 0x00007f21448fc55c in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#13 0x00007f2143744e08 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
[...]

PR #1136, opened 2024-06-04
Fixed a crash on entering `exit` when bookmark dock is shown

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/1135.


tsujan:
This fix is completely safe, even if the crash doesn't happen for some users. So, I merge it.

Issue #1140, opened 2024-06-18
Height of drop-down terminal is incorrect


tsujan:
The height of the drop-down terminal isn't what is set in Preferences → Dropdown → Size.

I wonder why this issue has remained hidden for such a long time. Perhaps most users have adjusted the height visually and overlooked the discrepancy — as I did.

The problem is caused by an incorrect line of code. I'll fix it soon.                                      --->

##### System Information

* QTerminal version: 2.0.0
* Qt Version: 6.7.1

PR #1141, opened 2024-06-18
Fixed the window height in drop-down mode

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/1140


stefonarch (approved):
All fine, but it's not backwards compatible - we'll reduce the size on upgrading ;)

tsujan:
> but it's not backwards compatible - we'll reduce the size on upgrading ;)

Yes. People will notice the difference and think the fix is a bug. We might have to close some reports later ;)

PR #1144, opened 2024-06-27
Pre-release changes

tsujan:
Also set `QTERMWIDGET_MINIMUM_VERSION`, which was missing after Qt6 port.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 qterminal-2.0.1 / 2024-06-27
 =============================
  * Single instance for drop-down mode.
+ * Used layer shell for the drop-down mode under Wayland.
  * Prevented a crash on entering `exit` when bookmark dock is shown.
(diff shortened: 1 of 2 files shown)


stefonarch:
`    Set layer overlay for dropdown terminal on wayland (#1113)` is missingin the changelog.

And I think we should also allow to close the dropdown terminal windows with modal dialogs open on wayland as it doesn't look as this can be fixed on compositors side anytime soon.

tsujan:
> And I think we should also allow to close the dropdown terminal windows with modal dialogs open on wayland

I think you forget that it isn't limited to the preferences dialog and that the terminal should remain open on showing other dialogs.

> it doesn't look as this can be fixed on compositors side anytime soon.

It _should_ be fixed there. Whether they find a way soon or not is out of our control.

tsujan:
> ... is missing in the changelog.

Added now. I thought we did it for 2.0.0.

Issue #1145, opened 2024-06-29
Changed settings are overwritten again if closing another open instance


stefonarch:
If there is more than one instance open (like a normal and a dropdown window) changes made in the first are overwritten by the old settings  from the other instance when closing it.

##### Expected Behavior

Changed settings remain changed.
##### Current Behavior

They don't, which means users have to close all other  instances before changing a setting.

##### Steps to Reproduce (for bugs)

1. Open 2 instances of QTerminal
2. Change settings in the first and exit it.
3. Close the second instance.
4. Open QTerminal → settings are unchanged.

##### System Information

* QTerminal version: all

PR #1146, opened 2024-06-29
Resolved conflicts when settings are changed by multiple instances

tsujan:
Please note that changing of settings in one instance isn't reflected immediately by the windows of others, because that might result in problems.

Fixes https://github.com/lxqt/qterminal/issues/1145

### DIFF
--- a/src/propertiesdialog.cpp
+++ b/src/propertiesdialog.cpp
@@ -368,9 +368,7 @@ void PropertiesDialog::apply()
     Properties::Instance()->historyLimited = historyLimited->isChecked();
     Properties::Instance()->historyLimitedTo = historyLimitedTo->value();
 
-    saveShortcuts();
-
-    Properties::Instance()->saveSettings();
+    applyShortcuts();
 
     Properties::Instance()->dropShowOnStart = dropShowOnStartCheckBox->isChecked();
     Properties::Instance()->dropKeepOpen = dropKeepOpenCheckBox->isChecked();
@@ -379,10 +377,6 @@ void PropertiesDialog::apply()
     Properties::Instance()->dropShortCut = dropShortCutEdit->keySequence();
 
     Properties::Instance()->useBookmarks = useBookmarksCheckBox->isChecked();
-    saveBookmarksFile();
-    // NOTE: Because the path of the bookmarks file may be changed by saveBookmarksFile(),
-    // it should be saved only after that.
-    Properties::Instance()->bookmarksFile = bookmarksLineEdit->text();
 
     Properties::Instance()->terminalsPreset = terminalPresetComboBox->currentIndex();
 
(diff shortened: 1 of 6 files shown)


isf63:
Wouldn't the easiest solution be to not save settings on exit, and only upon Ok/Apply in settings?

tsujan:
> Wouldn't the easiest solution be to not...

That would result in bugs. Some settings are changed directly from the main window.

stefonarch (commented):
At first glance all works fine now, but I found that the preset (1, 2 vertical, 2 horizontal, 4 terminals) is just not working anymore. The line `TerminalsPreset=0`  is never updated, reopening the settings it always switches back to 1 terminal.

Changing it to 1,2 or 3 works, but can't be changes anymore again in the GUI.

stefonarch:
Just rechecked in master git: there it works fine.

tsujan:
Thanks for finding the problem. I think there should be 7 other settings with the same behavior and that they come from another part of the code that isn't compatible with this change, namely `PropertiesDialog::apply()`. Will look into it.

tsujan:
@stefonarch, I added a commit; please try again.

Among other issues, changes made to shortcuts by another instance would be lost if its window was closed *before* the current one. That's fixed now. Moreover, as soon as the preferences dialog of the current instance is shown, the new shortcuts will also be applied to its window; otherwise, they will be applied to new windows.

I didn't find another problem by checking the code, but you might find one in practice ;)

src/propertiesdialog.cpp
```diff
@@ -370,19 +370,13 @@ void PropertiesDialog::apply()
 
     saveShortcuts();
```
correabuscar, line 371: this causes a `Properties::Instance()->saveSettings();` [inside](https://github.com/lxqt/qterminal/pull/1146/files/ab8ccaf23aa6631e36fadef9e27221a784b3905b..8509f7e1a55f11bd194b197d6cff6fe78517431d#diff-fb415beef169a9cbd4c61db798927f7032fedfa56d3abf9174e85716f35e3d87R461) it, will it affect anything by doing that here? then a bit later below at line [397](https://github.com/lxqt/qterminal/pull/1146/files/ab8ccaf23aa6631e36fadef9e27221a784b3905b..8509f7e1a55f11bd194b197d6cff6fe78517431d#diff-fb415beef169a9cbd4c61db798927f7032fedfa56d3abf9174e85716f35e3d87R397) ? ie. a double save so quickly?

stefonarch (approved):
Didn't find any issues anymore now.

src/propertiesdialog.cpp
tsujan: Yes, I noticed it too but left it as it was. Will remove it before merging the PR.

IMO, The code structure of QTerminal is far from ideal — each class may suppose some things about another, there are redundant computations, nullity checks aren't sufficient, etc. So, in general, some sort of restructuring may be needed, but it should be done with the utmost care to prevent regressions. Perhaps a job for the future...

tsujan:
> Didn't find any issues anymore now.

I also did my best to find an issue, both in the PR and in practice, but only found a few small issues in the original code instead. Might address them later, in separate PRs.

tsujan:
Any objection to merging this? I've used it and encountered no issue. Giving it to git users means having more testers.

PR #1151, opened 2024-07-16
Added hint on dopdown shortcut under Wayland

tsujan:
… to Terminal settings → Dropdown → Shortcut.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -985,7 +985,7 @@ To remove/disable a Shortcut, at point 2 press only a modifier (like Shift)</str
        <item>
         <widget class="QLabel" name="waylandLabel">
          <property name="text">
-          <string>Note: On Wayland, you need to add the shortcut to the compositor settings for 'qterminal -d'.</string>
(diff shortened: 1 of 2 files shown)


stefonarch (commented):
IMHO the text shouldn't be personal with "you" but neutral, kind of : "Note: A shortcut for qterminal -d has to be added manually in the compositor settings". Afaik we use this on other places too.

tsujan:
(1) That "manually" can be confusing; some compositors may have GUI for it. (2) "A shortcut" is also confusing; we want to say concisely that this very shortcut won't work on Wayland unless added to the compositor. (3) Direct sentences are usually preferable. And (4) QTerminal is used by a person ;)

stefonarch:
Sure, manually is weird, but you use your hands also in a GUI ;)

immagine

tsujan:
Are you happy with "Note: On Wayland, the shortcut has to be added in the compositor settings for 'qterminal -d'."? I put `for 'qterminal -d'` at the end intentionally because the shortcut has the same role but is for that command on Wayland.

stefonarch:
I'm ok with that text, it's clear and in line with other notes.

tsujan:
Done.

stefonarch (approved):

Issue #1153, opened 2024-07-18
New main windows spawned in dropdown mode share the same process


tsujan:
##### Steps to Reproduce (for bugs)

1. Open the dropdown QTerminal.
2. Show its menubar and click File → New Window.
3. Enter `exit` inside the dropdown window.

Now you have an ordinary window, and you can't open any dropdown window because the ordinary window has a dropdown process.

##### System Information

Latest git

PR #1154, opened 2024-07-18
Prevent ordinary main windows under dropdown process

tsujan:
When a new main window is opened in the dropdown mode (menubar → File → New Window), it should have a separate process because the dropdown process is supposed to consist of a single dropdown window. Of course, the work directory and profile must be respected too.

Fixes https://github.com/lxqt/qterminal/issues/1153


stefonarch (approved):
Works as expected now.

PR #1155, opened 2024-07-19
Add CMake option to disable Wayland support code

sideeffect42:
I added a CMake option to disable Wayland support code requiring LayerShellQt.

The dependency on LayerShellQt generates quite a few extra dependencies which are not used for people running QTerminal only on X11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ set(QTERMINAL_VERSION "2.0.1")
 
 option(UPDATE_TRANSLATIONS "Update source translation translations/*.ts files" OFF)
 option(BUILD_TESTS "Builds tests" ON)
+option(ENABLE_WAYLAND "Enable additional support code for the Wayland display server" ON)
 
 if(APPLE)
     option(APPLEBUNDLE "Build as qterminal.app bundle" ON)
@@ -31,7 +32,9 @@ find_package(Qt6Core ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6Gui ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6LinguistTools ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6Widgets ${QT_MINIMUM_VERSION} REQUIRED)
-find_package(LayerShellQt ${SHELLQT_MINIMUM_VERSION} REQUIRED)
+if(ENABLE_WAYLAND)
+    find_package(LayerShellQt ${SHELLQT_MINIMUM_VERSION} REQUIRED)
+endif()
 if(UNIX)
     find_package(Qt6DBus ${QT_MINIMUM_VERSION} REQUIRED)
     find_package(Qt6 COMPONENTS Core REQUIRED)
@@ -209,8 +212,11 @@ target_link_libraries(${EXE_NAME}
     Qt6::Gui
     Qt6::Widgets
     qtermwidget6
-    LayerShellQtInterface
 )
+if(ENABLE_WAYLAND)
+    add_definitions(-DHAVE_LAYERSHELLQT)
+    target_link_libraries(${EXE_NAME} LayerShellQtInterface)
+endif()
 if(QXT_FOUND)
     target_link_libraries(${EXE_NAME} ${QXT_CORE_LIB} ${QXT_GUI_LIB})
 endif()
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -213,6 +213,13 @@ int main(int argc, char *argv[])
         app->installTranslator(&translator);
     }
 
+#ifndef HAVE_LAYERSHELLQT
+    if (QGuiApplication::platformName() == QStringLiteral("wayland"))
+    {
+        fprintf(stderr, "Running on Wayland, although Wayland support was disabled at compile-time. Expect errors.\n");
+    }
+#endif
+
     TerminalConfig initConfig = TerminalConfig(workdir, shell_command);
     app->newWindow(dropMode, initConfig);
 
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -44,8 +44,10 @@
 #include "qterminalapp.h"
 #include "dbusaddressable.h"
 
+#ifdef HAVE_LAYERSHELLQT
 #include <LayerShellQt/Shell>
 #include <LayerShellQt/Window>
+#endif
 
 typedef std::function<bool(MainWindow&, QAction *)> checkfn;
 Q_DECLARE_METATYPE(checkfn)
@@ -173,6 +175,7 @@ void MainWindow::enableDropMode()
 {
     if (QGuiApplication::platformName() == QStringLiteral("wayland"))
     {
+#ifdef HAVE_LAYERSHELLQT
         winId();
         if (QWindow *win = windowHandle())
         {
@@ -184,6 +187,7 @@ void MainWindow::enableDropMode()
                 layershell->setAnchors(anchors);
             }
         }
+#endif
     }
 
     setWindowFlags(Qt::Dialog | Qt::WindowStaysOnTopHint | Qt::CustomizeWindowHint);


tsujan:
IMO, this isn't acceptable anywhere in LXQt. You may dislike Wayland, but it's unavoidable in the long rum.

LayerShellQt is just a small dependency, whose removal doesn't justify adding conditions that could easily complicate development and maintenance.

sideeffect42:
This PR is not about whether or not I like Wayland. I'm agnostic, actually.
It's just that my window manager is X11 and so I don't see a reason to clobber the system with Wayland-only libraries which will never be executed.

I've seen other software having options to enable or disable Wayland/X11 support at compile-time (including Qt itself).
Is wrapping Wayland-specific (or X11-specific for that matter) code in an `#ifdef` really so complicated?
Currently it's a dozen of lines in total and I would go so far as to say that QTerminal (in an ideal world) shouldn't require any display server specific code at all.

While LayerShellQt may be a small dependency, it comes with a rat tail of other requirements.
On my system this "small dependency" triggers changes in at least 6 other packages:
```
[ebuild  N     ] x11-terms/qterminal-2.0.1-r1::gentoo  USE="-test" 225 KiB
[ebuild  N     ]  kde-plasma/layer-shell-qt-6.1.3:6::gentoo  USE="-debug" 36 KiB
[nomerge       ] kde-plasma/layer-shell-qt-6.1.3:6::gentoo  USE="-debug" 
[ebuild  NS    ]  kde-frameworks/kf-env-6:6::gentoo [5:5::gentoo] 0 KiB
[nomerge       ] kde-plasma/layer-shell-qt-6.1.3:6::gentoo  USE="-debug" 
[ebuild  N     ]  dev-qt/qtwayland-6.7.2-r3:6/6.7.2::gentoo  USE="vulkan -accessibility -compositor -qml -test" 1,097 KiB
[ebuild     U  ]   dev-qt/qtbase-6.7.2-r1:6/6.7.2::gentoo [6.7.2:6/6.7.2::gentoo] USE="X concurrent cups dbus gles2-only gtk gui icu libinput libproxy network nls opengl sql sqlite ssl udev vulkan wayland* widgets xml (zstd) -accessibility -brotli -eglfs -evdev -gssapi -mysql (-oci8) -odbc -postgres (-renderdoc) -sctp -test -tslib" 0 KiB
[ebuild  N     ]  dev-libs/wayland-protocols-1.36::gentoo  USE="-test" 94 KiB
[ebuild  N     ]  dev-libs/wayland-1.23.0::gentoo  USE="-doc -test" 233 KiB
[ebuild  N     ]   dev-util/wayland-scanner-1.23.0::gentoo  0 KiB
```

LXQt is about lightness, isn't it?

tsujan:
On Arch, the package of `layer-shell-qt` (33 KiB) depends on `qt6-wayland` (1.1 MiB) that depends on `wayland` (137 KiB). Their installation takes < 9 MiB of disk space.

Anyway, that's not the point.

> LXQt is about lightness, isn't it?

Yes, but it isn't a bare-bones DE either.

sideeffect42:
> On Arch, the package of `layer-shell-qt` (33 KiB) depends on `qt6-wayland` (1.1 MiB) that depends on `wayland` (137 KiB). Their installation takes < 9 MiB of disk space.
> 
> Anyway, that's not the point.

Correct, the point is not installed binary size.
On Gentoo the build dependencies also need to be installed, that's why it's a few more.

In any case: are you suggesting me to recompile Qt with Wayland support, compile the Qt Wayland module and the Wayland libraries which I won't be able to use because my window manager does not support Wayland?
Just to avoid `#ifdef`ing 5 lines of code?

Or is QTerminal just not meant to be used outside of the LXQt desktop environment and I should be looking for an alternative instead?

stefonarch:
I'm half-half on this, it's not a invasive change as there are a few lines only to disable a small block. The point is if we accept it in QTerminal why not in pcmanfm-qt or in the other components? We could accept it in apps but not in the DE?

sideeffect42:
@stefonarch pcmanfm-qt would've been my next target, indeed.

tsujan:
> I'm half-half on this

I'm against it for the reason I clearly mentioned in my first comment. Also, I don't have time for reports like this:

— It doesn't work on LabWC/Wayfire/....
— Have you compiled it with LayerShellQt?
— What's LayerShellQt?
....

tsujan:
> Or is QTerminal just not meant to be used outside of the LXQt desktop environment and I should be looking for an alternative instead?

This was irrelevant. Of course QTerminal can be used outside LXQt or without Wayland.

But if you want a bare-bones system, you'll have to install bare-bones programs.

sideeffect42:
I don't think this change will be the source of many reports.

First of all, the option is default _on_, so you have to actively disable it.
And most people won't be compiling from source. Binary distros will enable the option anyway because they likely also ship Wayland-enabled DEs and so will want it.

stefonarch:
For me for the applications only this option would be fine. If it will generate timeconsuming issues we can still revert it but I don't think so as most distros are shipping wayland.

For the DE it's a not necessary because we depend on kwindowsystem/kscreen which depends among other on layer-shell-qt anyway.

Let's hear others maybe.

tsujan:
Also see https://github.com/tsujan/Kvantum/discussions/948. It's as "valid" as this.

I think we should avoid changing our codes because someone wants to compile Qt without X11 or Wayland. Of course, they have right to do so, but they also have to accept the consequences.

Last but not least, as the new maintainer (or half-Maintainer) of QTerminal, and considering that I also maintain libfm-qt, pcmanfm-qt, lximage-qt,..., I should manage my time efficiently. That requires having an approach different from the previous maintainer. We lack manpower for these things.

EeyoreTheDonkee:
> Let's hear others maybe. - @stefonarch

So, ok, my two cents.. below.

> I think we should avoid changing our codes because someone wants to compile Qt without X11 or Wayland. Of course, they have right to do so, but they also have to accept the consequences. - @tsujan

> IMO, this isn't acceptable anywhere in LXQt. You may dislike Wayland, but it's unavoidable in the long rum. - @tsujan

This is either a hard codified agenda or speculation. But, both lead to the same conclusion. There will be no choice at all.

I think since I do like LXQt -now- and I run it with X11 exclusively. It's not hard to choose the path that tries to maintain performance and efficiency for such a platform. If you can't, you can't .. fine. I'll find a different DE.

Too bad.

stefonarch:
It would be only a small change here, sure. 

But the point is that it would be needed to be implemented anywhere (panel, notifications, lxqt-leave, powermanagement, screengrab...) and that is a no-go IMO, especially the panel. So it wouldn't change anything for LXQt users who want no Wayland dependencies.

Our choice to hide all Wayland related settings if `lxqt-wayland-session` is not installed is a good compromise I think.

sideeffect42:
> But the point is that it would be needed to be implemented anywhere (panel, notifications, lxqt-leave, powermanagement, screengrab...) and that is a no-go IMO, especially the panel. So it wouldn't change anything for LXQt users who want no Wayland dependencies.
> 
> Our choice to hide all Wayland related settings if `lxqt-wayland-session` is not installed is a good compromise I think.

It really depends on the use case.
If LXQt has decided that it will be a Wayland desktop environment in the future, then I can totally understand not wanting to add configure options to components like the panel, notifications, lxqt-leave etc.

However, there is a different use case (which I was after when I opened this pull) and it is taking some LXQt apps (which are really good) and run them on top of another window manager/desktop environment, in my case awesomewm.

But I took the response to this pull request as that it is not desired by the LXQt developers that people take their apps to run them on top of other desktop environments.

stefonarch:
>If LXQt has decided that it will be a Wayland desktop environment in the future

There is no such "decision". As things go we will support both, and for X11 it will depend on Qt and KF when this will be impossible. We won't take a side in any holy war.

IMO for QTerminal  _only_ we could accept  this configure option. 
Qps, lxqt-archiver do not have any Wayland dependencies and for screengrab there are many other screenshot tools available.

tsujan:
> We won't take a side in any holy war.

Exactly.

However, whether one likes it or not, X11 or Xlibre doesn't seem to have a bright future.

Issue #796, opened 2021-02-18
Feature request : --fullscreen switch


T0MuX:
Hi, this is not really an issue, but a feature request.
Is there a way to start qterminal in fullscreen mode ? I tried --fullscreen -fullscreen -f but these options are not recognized. To me this function could be great :)

Thank you

ballo:
Not being able to launch this with full screen is annoying as I have to reboot often. Is there a work-around?

ballo:
I think I found a work-around:
qterminal --geometry 9000x9000

stefonarch:
Some window manager (openbox, kwin) have rules for windows, you could use them.

tsujan:
However I look at it, adding a trivial action like pressing F11 to the command-line options of a terminal emulator isn't justified.

PR #1175, opened 2024-10-29
Fullscreen startup option flag

josiahtripp:
Added an option flag -f "--fullscreen" which can be used to start qterminal in fullscreen mode.

The flag sets the main window to fullscreen & checks the box for the fullscreen option in the GUI "view" menu.

When a window is opened using the flag, any additional windows created by the main window will not be opened in fullscreen mode.

Resolves issue #796
(Feature request)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -77,10 +77,11 @@ QTerminalApp * QTerminalApp::m_instance = nullptr;
     exit(code);
 }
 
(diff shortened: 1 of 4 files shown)


tsujan:
Reviewing the code aside, I'm not sure this is needed in practice, i.e., beyond personal preferences of a few users.

Issue #1104, opened 2024-02-20
New tab from Preset Option 1 Terminal opens "Start with preset"


AlphaWHH:
##### Expected Behavior
Start with preset is set to any option in the config, the 'New Tab From Preset':1Terminal should open a single terminal.

##### Current Behavior
In any configuration the 'New Tab From Preset':1Terminal will open the selected option in the config under start with preset.

##### Steps to Reproduce (for bugs)
1. Set 4 terminals preset in menu
2. new tab from preset and select 1 Terminal
3. The preset option is used instead of the 1 terminal

##### Possible Solution
Current code
Mainwindow.cpp line 363-364
presetsMenu->addAction(QIcon(), tr("1 &Terminal"),
this, SLOT(addNewTab()));

PROBLEM AREA:
Line 879 void MainWindow::addNewTab(TerminalConfig cfg)
{
    if (Properties::Instance()->terminalsPreset == 3)
        consoleTabulator->preset4Terminals();
    else if (Properties::Instance()->terminalsPreset == 2)
        consoleTabulator->preset2Vertical();
    else if (Properties::Instance()->terminalsPreset == 1)
        consoleTabulator->preset2Horizontal();
    else
        consoleTabulator->addNewTab(cfg);
    // disabled actions are updated by TabWidget::onCurrentChanged()
}
PROPOSED SOLUTION:
Mainwindow.cpp line 363-364
presetsMenu->addAction(QIcon(), tr("1 &Terminal"),
this, SLOT(consoleTabulator->addNewTab(cfg);));

tsujan:
Thanks for reporting this! It was hidden among less important reports. We lack enough manpower, and I should attend to other projects first.

PR #1176, opened 2024-10-30
Fixed new tab from preset and context menu's collapse action

tsujan:
This patch fixes two bugs when the terminal is started with a preset of N > 1 terminals:

 * Previously, the action "New tab from preset" didn't work correctly when the number of preset was less than N.
 * Also, if the terminal context menu was the first menu that was opened, the action "Collapse Subterminal" would be disabled.

I also silenced a compilation warning about `&QCheckBox::stateChanged`.

Fixes https://github.com/lxqt/qterminal/issues/1104


tsujan:
@stefonarch 
I think it's better to remove these bugs before the next release; hence bothering you with testing. I've tested it too.

stefonarch (approved):
Confirming issues are gone.

PR #1178, opened 2024-11-03
propertiesdialog: small enhancements for UI consistency

Fat-Zer:
- add a space before "px" suffix
- remove colons ':' in the form's labels

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -276,7 +276,7 @@
            <item row="12" column="0">
             <widget class="QCheckBox" name="fixedTabWidthCheckBox">
              <property name="text">
-              <string>Fixed tab width:</string>
+              <string>Fixed tab width</string>
              </property>
             </widget>
            </item>
@@ -286,7 +286,7 @@
               <bool>false</bool>
              </property>
              <property name="suffix">
-              <string>px</string>
+              <string> px</string>
              </property>
              <property name="maximum">
               <number>1000</number>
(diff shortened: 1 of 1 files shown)


isf63:
This looks good, granted translations would be wiped unfortunately, unless manually edited (which is a hassle.)

Apart from that, this has merge conflicts that need to be resolved before being able to merge.

isf63:
Actually I was confused: labels of actionable widgets (apart from QGroupBox) should have colons.

Fat-Zer:
> Actually I was confused: labels of actionable widgets (apart from QGroupBox) should have colons.

As for now most of them don't have colons, so‌ I've assumed it was the preferred style. I can update the PR adding the missing one instead if you wish.

isf63:
> I can update the PR adding the missing one instead if you wish.

Please do. You can refer to PCManFM-Qt's Preferences dialog to see the general convention.

Unfortunately this will result in many translations being invalidated (as the entire string is used in `tr(...)`), and will require manual changes. As there are many missing colons this will be quite cumbersome.

" px" is used in PCManFM-Qt and probably is the correct string.

Fat-Zer:
done

> Unfortunately this will result in many translations being invalidated (as the entire string is used in tr(...)), and will require manual changes. As there are many missing colons this will be quite cumbersome.

Well... that's expected expenses of gettext-like translations.

> " px" is used in PCManFM-Qt and probably is the correct string.

There is already yet another line edit with " px" suffix, so it should catch up automatically even without the rest of lxqt.

stefonarch:
I managed to preserve translations in `lxqt-session`, will look into it. Don't merge yet.

Issue #1188, opened 2024-11-20
An option to focus sub-terminals on mouse-over


tsujan:
This was what @stv-t had in mind at https://github.com/lxqt/qterminal/issues/1187.

PR #1189, opened 2024-11-20
Focuse the sub-terminal on mouseover

tsujan:
Closes https://github.com/lxqt/qterminal/issues/1188

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -69,7 +69,7 @@
       <enum>QFrame::StyledPanel</enum>
      </property>
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="appearancePage">
       <layout class="QVBoxLayout" name="verticalLayout_4">
@@ -301,6 +301,13 @@
             </widget>
            </item>
            <item row="14" column="0" colspan="2">
+            <widget class="QCheckBox" name="focusOnMoueOverCheckBox">
+             <property name="text">
+              <string>Focus the terminal on mouseover</string>
+             </property>
+            </widget>
+           </item>
+           <item row="15" column="0" colspan="2">
             <widget class="QCheckBox" name="closeTabButtonCheckBox">
              <property name="text">
               <string>Show close button on each tab</string>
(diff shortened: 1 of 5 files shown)


tsujan:
@stefonarch 
My other 2 PRs are also waiting for you ;)

stefonarch:
> @stefonarch My other 2 PRs are also waiting for you ;)

I didn't manage to have 2 links in Qterminal and neither the slash thing. But you didn't ask for reviews no?

stefonarch (commented):
Works fine, but the setting shouldn't be under "Appearance" but under "Behavior" IMO.

tsujan:
> But you didn't ask for reviews no?

I feared you might say no, because you were busy. That was the only reason.

> but the setting shouldn't be under "Appearance" but under "Behavior" IMO.

IMO, it should be where "Show a border..." is. Without a border, it's dangerous.

tsujan:
> Could "show border" be forced automatically when activation follow mouse?

Users have right to feel brave.

> Keyboard focus has little to do with appearance.

Several options can be interpreted both ways. Appearance and Behavior were words chosen a long time ago, and now they aren't good enough. We might want to make changes to that dialog later. Frankly, I can hardly find the option I want among that mess.

However, the 4 options about the current (= focused) terminal should be together, IMO, as anyone may expect.

tsujan:
I'll rearrange the Preferences dialog in my next PR. It's a long time I want to do so, but either I'm busy, or I forget.

tsujan:
Orderly, neat, and easy to search:




"Set the current terminal on mouseover" will be the new text, and it'll remain beside the other settings for the current terminal.

stefonarch (commented):
The "current" one is the one already focused on mouseover. I think it was better before, or just "Focus terminal on mouseover".

tsujan:
> The "current" one is the one already focused on mouseover

No. The current terminal is the one that *will be* focused if the whole window is activated. You can see that by putting the mouse cursor over a inactive terminal window and *then* activating it.

That shows the main problem of "Focus the terminal on mouseover": it might easily be mistaken for focusing the window on mouseover, as if we want to do a WM job here, because people use the word "terminal" for the window too.

tsujan:
Went a step further and made it so in the upcoming PR:



You wont find a cleaner GUI anywhere ;)

However, here isn't the place to discuss it.

stefonarch (approved):

tsujan:
@stefonarch 
While preparing the next patch (will make a PR soon), I saw that some texts weren't good. I found these:

 * Some labels like "Application transparency", that come before another widget (spin/combo box) have no ending colon (this one can't work on Wayland, BTW). We either should add those colons or remove them from other places (I think it's reasonable to use colons).
 * In "Save Position..." and "Save Size...", only the first letter should be capitilized.

I didn't touch them because it would affect trasnlations.

stefonarch:
I think both can be solved preserving translations by using sed as we did before when changing capitalized strings. Maybe change those texts in a different PR is better,

tsujan:
> I think both can be solved preserving translations by using sed

Yes. That's why I told you — you've done it before.

> Maybe change those texts in a different PR is better

Yes, after https://github.com/lxqt/qterminal/pull/1191 is decided.

PR #1191, opened 2024-11-21
Cleaner preferences dialog

tsujan:


tsujan:
With a Kvantum theme (there are screenshots with Fusion at https://github.com/lxqt/qterminal/pull/1189):

kv

tsujan:
Just a reminder, for another PR, later:

"Application transparency", "Save position when closing", and perhaps the shortcut entry are better to be disabled under Wayland because they can't do anything there.

BTW, IMO, app transparency is redundant and weird in general. We have terminal transparency. My guess is that app transparency was added when terminal transparency seemed impossible for some reason. But it's a topic for another discussion.

stefonarch (approved):

tsujan:
> Spotted a lot of missing colons

Yes, there are some. Would you please add them? Although it isn't easy, because of French and some other languages.

> application transparency could be removed

I agree, but for now (in an upcoming PR), I'll just disable it. We could make a report and wait to see if there's no objection.

tsujan:
> For the colons: I think there is no way to preserve translations

It's OK.

stefonarch:
Well "no way" doesn't exist. We would need to use a script which adds for every colon a colon in the next line like this:
```
        <source>Scrollbar position:</source>
        <translation>Posizione barra di scorrimento</translation>
```
to 
```
        <source>Scrollbar position:</source>
        <translation>Posizione barra di scorrimento:</translation>
```
While LTR needs manual intervention, I don't know cinese and other asian languages too.

tsujan:
The problem I had in mind is that some of them can't be simple colons, e.g., in French a special kind of space is needed before it.

stefonarch:
> The problem I had in mind is that some of them can't be simple colons, e.g., in French a special kind of space is needed before it.

That will be done by updating translations afaik.

isf63:
Finally I can parse the options!

PR #1194, opened 2024-11-22
Change sizeHint of spacer in Preferences->Behavior

isf63:
Allows smaller window sizes for this page to not have a scrollbar


tsujan:
Yes, `0` is better.

Issue #1142, opened 2024-06-19
Dropdown terminal on wayland always on the same monitor


stefonarch:
We could use `layershell->setScreenConfiguration(LayerShellQt::Window::ScreenConfiguration::ScreenFromCompositor);` to open it where the cursor is, the downside is that it takes a screen resolution/size from one monitor and applies always that.

##### Possible Solution

A menu entry similar to `lxqt-runner`: On wayland open on monitor: active, 1, 2, ?
##### Steps to Reproduce (for bugs)

1. Use dropdown terminal on wayland with a dual head setup with different resolutions.
2. Move cursor between monitors and press shortcut
                                   

##### System Information

* QTerminal version: git

tsujan:
> the downside is that it takes a screen resolution/size from one monitor and applies always that.

Actually, there's an effective workaround for that.

The deeper problem arises when the second (third,...) monitor is scaled. Then the window will have a wrong pixel ratio (= scale factor) inside it, and that will ruin everything. This is a bug in Qt's handling of Wayland — probably in `layer-shell-qt` — and can be seen in lxqt-runner (and other Qt apps) too, although it's tolerable there because lxqt-runner usually has a small window.

tsujan:
A more accurate explanation:

The enum `ScreenFromCompositor` opens the window on the screen with the cursor. So far, so good. However, the window is opened with the pixel ratio it had before.

For example, suppose there are two screens with the same resolution, but the second one has a scale factor of 2. You put the mouse cursor into the first screen and open the dropdown window. Then you close it, put the cursor inside the second screen, and reopen it. Although the pixel ratio of the second screen is 2.0, that of the window is still 1.0. The result is a huge window.

It may be asked, "who scales only one of two monitors that have the same resolution?!" That was only one example; a similar (but a little more complex) situation happens with different resolutions, when only one screen is scaled, or when scale factors are different: we'll end up with a wrong window size — either too big, or too small.

tsujan:
For reference: https://bugs.kde.org/show_bug.cgi?id=482733, https://bugs.kde.org/show_bug.cgi?id=485069, https://bugs.kde.org/show_bug.cgi?id=488780, and probably more.

tsujan:
> On wayland open on monitor: active, 1, 2, ?

The "active" option isn't possible — I explained the reason above.

[...]

PR #1196, opened 2024-11-28
Put dropdown window on active screen under Wayland

tsujan:
All complex problems in implementing this feature were caused by resizing the window programmatically on Wayland, as was done on X11. This patch sets 4 anchors instead, and then sets the left, right and bottom margins properly on showing the window.

Fixes https://github.com/lxqt/qterminal/issues/1142


stefonarch (approved):
Nice work, works flawlessly!

PR #1197, opened 2024-11-28
Added scope

stefonarch:
Can be useful for compositors like Hyprland for rules:
`layerrule = animation slide top, dropdown_terminal`


tsujan (approved):
No harm to it ;)

PR #1200, opened 2024-11-30
Prompt on closing with a running process

tsujan:
Previously, the closing prompt was useful only when there was a running process in the terminal, but the code didn't check that. A prompt without a running process was redundant.

This patch gives a real job to the closing prompt by making it work only when there is a running process, whether when the main window is going to be closed, or when a subterminal is going to be collapsed.

To detect a running process, the code makes use of @marcusbritanicus's commit https://github.com/lxqt/qtermwidget/commit/ccb3b2132d3c7d4b19b9a3221f9c4dcf13eee9dc by comparing `getForegroundProcessId()` with `getShellPID()`. In addition, it checks whether `TerminalConfig` has a command (e.g., when an app is executed in the terminal by the file manager).

Therefore, the config text "Ask for confirmation when closing" is also changed to "Prompt on closing with a running process".

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -825,7 +825,7 @@
               <item row="7" column="0" colspan="2">
                <widget class="QCheckBox" name="askOnExitCheckBox">
                 <property name="text">
-                 <string>Ask for confirmation when closing</string>
+                 <string>Prompt on closing with a running process</string>
                 </property>
                </widget>
               </item>
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -306,7 +306,7 @@ void MainWindow::setup_ActionsMenu_Actions()
 
     data.setValue(checkSubterminals);
 
-    setup_Action(SUB_COLLAPSE, new QAction(tr("&Collapse Subterminal"), settingOwner),
+    setup_Action(SUB_COLLAPSE, new QAction(tr("&Close Subterminal"), settingOwner),
                  nullptr, consoleTabulator, SLOT(splitCollapse()), menu_Actions, data);
 
     setup_Action(SUB_TOP, new QAction(QIcon::fromTheme(QStringLiteral("go-up")), tr("&Top Subterminal"), settingOwner),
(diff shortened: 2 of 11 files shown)


tsujan:
@marcusbritanicus
Could you please take a look at it to see whether I've missed something?

stefonarch (commented):
Same issue with the preferences dialog in the dropdown: the prompt opens under the window. Regular window is fine. IMO we should use layershell for those if on wayland and if the terminal is dropdown.

tsujan:
> Same issue with the preferences dialog in the dropdown

It's not related to this PR ;)

And please try the PR (again) after I add a commit — I missed a rare and special case of closing. Will tell you when it's ready.

tsujan:
@stefonarch 
Now it's ready for your tests. Please note that there are various cases of closing a terminal with a running process in it (window, tab, collapsing ,...).

stefonarch:
Not always the checkbox and text "Don't ask again" is shown, for example when  2 tabs are open. Looks good overall but  will test more later.

tsujan:
That check-box is shown only when closing the window. My plan is to remove it in an upcoming PR because the prompt has a real meaning now.

tsujan:
The prompts should be shown with a running process on

 * Closing the whole window, with the close button or the shorcut.
 * Closing a tab, with its close button, by middle clicking, by right clicking and choosing "Close session", or just the shortcut (if any).
 * Collapsing a (sub-) terminal.
 * Opening an executable file inside terminal in pcmanfm-qt.

tsujan:
I my last commit, I unified all prompt dialogs in one place but with appropriate, different texts and titles (for exiting, closing tab, and closing subterminal). In that way, "Don't ask again" is always shown too.

I also changed `Collapse Subterminal` to `Close Subterminal` for the sake of consistency. Sorry, it should be translated anew. If that's not acceptable to you, tell me to reverse it. My reason is that "collapse" may be mistaken for shrinking or folding, as if the subterminal is still there but hidden. Too pedantic? ;)

stefonarch:
> Opening an executable file inside terminal in pcmanfm-qt and trying to close the window.

Didn't understand that.

I'm fine with changing "collapse".

tsujan:
> I'm fine with changing "collapse".

Good!

In GUIs, "collapse" is usually used alongside "expand": when something "collapses", no widget is removed, destroyed or closed. Here we really close (=destroy) the subterminal.

> Didn't understand that.

When you click an executable binary file, pcmanfm-qt shows a prompt, and you can choose "Execute in Terminal"..

stefonarch (approved):

PR #1205, opened 2024-12-02
Better wording for current terminal section

stefonarch:


tsujan (approved):
Yes, we have a group box now. GTM.

Issue #403, opened 2018-02-12
Cursor autohide feature


ysoloviov:
Can this feature be added? It was very useful when I used vte based terminals. Konsole also has this feature but for my purposes it is a little bit bloated.

RomanBruhanov:
There is setAutoHideCursor in qtermwidget, but for that to work you will need to link to KDE. I guess it will be a lot of work to implement hiding in lxqt natively. Easier to stick with KDE or GTK terminals.

yan12125:
KWidgetsAddons is also in Tier 1, so depending on it will not be a big problem. Of course it's better to be optional to avoid issue on platforms where KF5 is not fully supported (e.g., macOS).

tsujan:
> KWidgetsAddons is also in Tier 1

Yes tier-1 K-things are nice extensions of Qt that don't depend on KDE. I think we could use them in LXQt if really needed -- as we use `KWindowSystem`.

luebking:
Why does that need some KDE stuff?
You monitor MouseMove events re/start a timer and when that actually fires QApplication::setOverrideCursor(Qt::BlankCursor); and QApplication::restoreOverrideCursor(); once the mouse starts moving.

Alternatively (just seeing this in phototonic) you could poll the mouse position (not sure whether I'm gonna replace that during the alpha cycle, but I'm pretty sure that's gonna leave there at some point ;)

tsujan:
> Why does that need some KDE stuff?

You're right: there's no need to any KDE stuff for that.

At the time this report was opened, I was rather a bystander (and a user) here. Months ago, I started to maintain qterminal and qtermwidget, went through their bug trackers to find the most important reports, and fixed them. Sadly, I can't check all valid reports, being busy with several other codes. Contributions are really appreciated.

luebking:
Preferably in qterminal or qtermwidget?
(I'm gonna check whether the latter tracks the mouse anyway and would base the decision whether to go for polling or event handling based on that - the implementation can be in either location)

tsujan:
> Preferably in qterminal or qtermwidget?

Wherever you see fit — I'm confident of your coding skills, having seen your codes and used some of them in [Kvantum](https://github.com/tsujan/Kvantum).

`qtermwidget` is the place where we put the *core* codes, i.e., those that can be used by any app based on `qtermwidget`. QTerminal is only one app that uses `qtermwidget`. So, if the feature you want to implement is for QTerminal, `qterminal` is the proper place.

[...]

PR #1222, opened 2025-02-06
Added option for mouse cursor auto-hiding

tsujan:
This patch needs https://github.com/lxqt/qtermwidget/pull/582.

The new option is in Preferences → Behavior. It can be from "No hiding" to 60 seconds. A value of zero hides the cursor immediately (as designed by @luebking).

Closes https://github.com/lxqt/qterminal/issues/403

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -772,7 +772,33 @@
                 </property>
                </widget>
               </item>
-              <item row="4" column="0" colspan="2">
+              <item row="4" column="0">
+               <widget class="QLabel" name="label_4">
+                <property name="text">
+                 <string>Mouse cursor hiding delay</string>
+                </property>
+               </widget>
+              </item>
+              <item row="4" column="1">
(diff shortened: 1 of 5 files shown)


stefonarch (commented):
Works fine, cursor remains where it should remain, but 0 should not be allowed as the cursor is never shown, not sure it anybody will like that...

stefonarch (approved):

tsujan:
> but 0 should not be allowed

I don't like it either, but who knows? Maybe someone does.

luebking:
Fwwi, I simply saw no reason to add artificial restrictions here and eg. 10ms is below one frame on 60fps and would be equivalent.
Also keep in mind that next to user preferences the widget can be used by sth. that's not a(n interactive) TE and where it would be desirable to have no mouse.

tsujan:
It's good for `qtermwidget`, and it could easily be disabled in QTerminal if users complain. So, there's no problem.

PR #1223, opened 2025-02-07
Added option for blinking cursor

tsujan:
The new option is in Preferences → Appearance. To prevent cluttering and duplicating, it isn't added to any menu.

Closes https://github.com/lxqt/qterminal/issues/924

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -175,7 +175,7 @@
              </layout>
             </widget>
            </item>
-           <item row="2" column="0">
+           <item row="1" column="0">
             <widget class="QLabel" name="label_5">
              <property name="text">
               <string>Color scheme</string>
@@ -185,10 +185,10 @@
              </property>
             </widget>
            </item>
-           <item row="2" column="1">
+           <item row="1" column="1">
             <widget class="QComboBox" name="colorSchemaCombo"/>
            </item>
-           <item row="3" column="0">
+           <item row="2" column="0">
             <widget class="QLabel" name="label_6">
              <property name="text">
               <string>Widget style</string>
@@ -198,10 +198,10 @@
              </property>
             </widget>
            </item>
-           <item row="3" column="1">
+           <item row="2" column="1">
             <widget class="QComboBox" name="styleComboBox"/>
            </item>
-           <item row="4" column="0">
+           <item row="3" column="0">
             <widget class="QLabel" name="label_7">
              <property name="text">
               <string>Scrollbar position</string>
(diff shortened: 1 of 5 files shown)


stefonarch (commented):
Blinks fast for my taste. Strange issues between dropdown and normal window. It worked first fine in dropdown, but not in normal window, now it won't work at all anymore after trying with both windows set and unset it, only a "double" cursor con be seen and only in dropdown terminal

https://github.com/user-attachments/assets/f142850a-0d9f-4e9a-b6b0-e7106c35f9f6

tsujan:
It uses the standard `QApplication::cursorFlashTime()` in `qtermwidget`. This simple patch only implements it in QTerminal.

stefonarch:
Started to work again somehow, but again only in the dropdown window. The cursor is drawn twice here too:

https://github.com/user-attachments/assets/f355e648-987b-4de5-b695-0962a177817b

tsujan:
> now it won't work at all anymore

I can't reproduce that. Please describes the steps for reproducing it.

tsujan:
Something's wrong in your both screencasts. First, the blinking seems too fast (but that may be a side effect of screencasting); second, there's a rectangle with a different color; third, the cursor is thicker than what it should be at the start.

How do you do that?

Please also note that the code is not in QTerminal, but in `qtermwidget`. So, if there's a bug, it should be fixed in the latter.

stefonarch:
Ok, done some round with compositors. This issue (doubled cursor and not working in normal windows) is 
a) only in Niri 
b) only with this shape- block and underline are fine, but the speed is always twice as normal.
No issues in sway or hyprland or labwc.

tsujan:
Niri has given us a headache recently ;) Such observations are signs of deep problems in it.

tsujan:
A shot in the dark:

Qt says that `QApplication::cursorFlashTime()` may be negative, e.g., when cursor blinking is disabled. I don't know if a Wayland compositor can do that, but we could add a safeguard to `qtermwidget` — IMO, we *should* do so, regardless of Niri.

stefonarch:
I just noticed that "Hide cursor delay" was set to 0, creating all those issues on Niri, except the blinking speed. We shouldn't allow it probably.

tsujan:
> We shouldn't allow it probably.

You didn't add that it's OK in all other Wayland compositors ;) Niri's problems are Niri's.

tsujan:
Added a safeguard against negative and too short values of `QApplication::cursorFlashTime()` in https://github.com/lxqt/qtermwidget/pull/583.

stefonarch:
> > We shouldn't allow it probably.
> 
> You didn't add that it's OK in all other Wayland compositors ;) Niri's problems are Niri's.

I didn't test _all_ seven.... sway and hyprland were fine.

https://github.com/lxqt/qtermwidget/pull/583 works for the blinking time on Niri.

stefonarch (approved):

tsujan:
> https://github.com/lxqt/qtermwidget/pull/583 works for the blinking time on Niri.

Good :) Until the next time, when Niri will show another bizarre behavior ;)

PR #1225, opened 2025-02-08
Don't allow zero timeout for mouse cursor auto-hiding

tsujan:
Although it is (and should be) allowed in `qtermwidget`, it has no use in QTerminal and can be confusing to users.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -795,13 +795,13 @@
                  <string> sec</string>
                 </property>
                 <property name="minimum">
-                 <number>-1</number>
+                 <number>0</number>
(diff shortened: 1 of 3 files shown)


tsujan:
@stefonarch 
Just for you ;)

stefonarch:
> @stefonarch Just for you ;)

Sometimes you open to reasons ;)

tsujan:
> Sometimes you open to reasons ;)

It depends on the dreams I had the night before.

stefonarch (approved):
Hoping your sleep is always good ;)

tsujan:
Good? You were threatening me in it; hence the PR.

Issue #1227, opened 2025-02-09
Font and color settings overlaid after recent changes


stefonarch:
Noticed only now:

![Image](https://github.com/user-attachments/assets/13e8da4f-ccaf-4168-8fc7-47d10bcf034c)

![Image](https://github.com/user-attachments/assets/f3a59919-59df-45df-ab92-1e8f613145dc)

##### Possible Solution

Fix propertiesdialog.ui
##### Steps to Reproduce (for bugs)

1. Open preferences in qterminal 

* QTerminal version: git

tsujan:
My mistake. Will fix it soon. Thanks!

PR #1228, opened 2025-02-09
Fixed a regression in preferences GUI

tsujan:
Fixes https://github.com/lxqt/qterminal/issues/1227


tsujan:
@stefonarch 
I remembered why I made that mistake, but please double check it.

stefonarch (approved):
Looks good now.

PR #1229, opened 2025-02-09
Removed a wrong style sheet from dropdown mode

tsujan:
Style sheets should never be applied in that way because, otherwise, they'll badly interfere with `QStyle`.

Only by chance, I saw the problem in combo-boxes of the dropdown preferences dialog with Kvantum; styles like Fusion and Breeze don't reveal such issues clearly (but, as the old code comment said, macOS was also unhappy about it).


tsujan:
The issue was a blasphemous ;) Merging …

PR #1230, opened 2025-02-09
Set font dialog in front of Preferences dialog on Wayland

tsujan:
This is rather a small hack than a clean code.

Also, silenced a compilation warning about "gridLayout".

### DIFF
--- a/src/fontdialog.cpp
+++ b/src/fontdialog.cpp
@@ -18,8 +18,8 @@
 
 #include "fontdialog.h"
 
-FontDialog::FontDialog(const QFont &f)
-    : QDialog(nullptr)
+FontDialog::FontDialog(const QFont &f, QWidget *parent)
+    : QDialog(parent)
 {
     setupUi(this);
 
--- a/src/fontdialog.h
+++ b/src/fontdialog.h
@@ -28,7 +28,7 @@ class FontDialog : public QDialog, public Ui::FontDialog
 {
     Q_OBJECT
 public:
-    FontDialog(const QFont &f);
+    FontDialog(const QFont &f, QWidget *parent = nullptr);
     QFont getFont();
 
 private slots:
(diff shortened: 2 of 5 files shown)


tsujan:
@stefonarch 
It works under labwc. If it doesn't work under Niri, uninstall Niri immediately ;)

stefonarch (approved):
Works fine, but testing X11 we have this issue there as well, also in master: font window is behind.

immagine

isf63:
> font window is behind

I believe the X11 WMs that have that bug are Xfwm4 (confirmed), and Openbox.

tsujan:
I don't touch X11 codes in patches for Wayland. It should be enough to change
```c++
    FontDialog dia(fontSampleLabel->font(),
                   QGuiApplication::platformName() == QStringLiteral("wayland") ? this : nullptr);
```
to
```c++
    FontDialog dia(fontSampleLabel->font(), this);
```

stefonarch:
> > font window is behind
> 
> I believe the X11 WMs that have that bug are Xfwm4 (confirmed, [EDIT: only in dropdown mode]), and Openbox.

Good catch, in fact this was xfwm4 in my test. Kwin_x11 is fine.

tsujan:
> could a (grey/palette) border be added in those circumstances?

The "outer" border is a part of decoration and drawn by WM. Wayland WMs don't draw any border in this case.

As for an "inner" border, a stylesheet can't be used because it'll interfere with `QStyle` (see https://github.com/lxqt/qterminal/commit/b6feac83f5f106919c76717cadb86e280fe9bd97), but it may be possible to draw it consistently in the code. However, the amount of the required code may not be worth the effort. I might experiment with it if I find the time.

> in fact this was xfwm4 in my test. Kwin_x11 is fine.

You could try my suggestion, [above](https://github.com/lxqt/qterminal/pull/1230#issuecomment-2646421877), and make a PR if it works under X11.

isf63:
--> https://github.com/lxqt/qterminal/pull/1231

PR #1231, opened 2025-02-09
Make font dialog in front of Preferences window on certain X11 WMs

isf63:
WMs:
- Xfwm4
- Openbox

Works for me; needs testing


stefonarch (approved):
Works in all 3 main stacking WMs

stefonarch:
I thought there was an issue about, couldn't find it.

tsujan:
@isf63 
Thanks for taking care of this!

`FontDialog` didn't accept a parent widget before, and that was a design flaw which was fixed in a recent commit about another thing.

PR #1234, opened 2025-02-11
Added namespaces

stefonarch:


tsujan (approved):
It's the layer shell scope. `namespace` is a totally different thing.

stefonarch:
> It's the layer shell scope. `namespace` is a totally different thing.

ops, yes :)

tsujan:
It's OK. I told that because, in other places, it might complicate writing of changelogs. But here, a single line is enough for multiple commits (about the font dialog).

Issue #1066, opened 2023-08-15
QTerminal dropdown occasionally opens in a minimized state


isf63:
##### Expected Behavior
Activating dropdown QTerminal with a hotkey should never open it minimized.

##### Current Behavior
Pressing a hotkey to open QTerminal opens it in a minimized state once or twice randomly throughout the day.

##### Steps to Reproduce (for bugs)
1. Press the hotkey to open QTerminal, then minimize it from the taskbar. Upon pressing the hotkey to open it opens in a minimized state.
2. Press the hotkey to open QTerminal, then press the Show Desktop panel button. 50% of the time upon pressing the hotkey to reopen it opens in a minimized state.

##### System Information

.config/qterminal.org/qterminal.ini DropMode section:
```
[DropMode]
Height=45
KeepOpen=false
ShortCut=F12
ShowOnStart=false
Width=70
```
* QTerminal version: qterminal-git
* Distribution & Version: Arch Linux
* Kernel: 6.4.10
* Qt Version: 5.15.10+kde+r147

tsujan:
Not reproducible with KWin. What's your WM?

isf63:
My WM is Xfwm4; tested on Openbox and it's the same behavior as Xfwm4.

stefonarch:
>Press the hotkey to open QTerminal, then minimize it from the taskbar.

The dropdown terminal shouldn't be show in the taskbar no?  Maybe I'm biased now by wayland where it isn't shown, but there is no need to have it in the taskbar.

isf63:
@stefonarch, it also happens when you press Show Desktop sometimes. In that case it's helpful that you can reopen it with the taskbar, but I have no preference for it being in the taskbar.

tsujan:
> it's helpful that you can refocus it by clicking the task item

Oh, if that's so, Xfwm4 has a serious problem, because the dropdown terminal should *hide* on losing its focus.

isf63:
Unfortunately this was not fixed in the recent 4.20 release. XFCE has a slow release cycle for minor versions (~2 years), although they sometimes do point releases.

For sake of completeness, these are all the issues I've noticed with Xfwm4 (X11):
- Xfwm4 shows the task button for QTerminal dropdown _always_ when active (normally or minimized.) (But not when correctly hidden.)
- If you minimize the QTerminal task button, reopening QTerminal opens it in a minimized state (can get the window open by clicking the task button.)
[...]

PR #1235, opened 2025-02-16
Add workaround for bug in reactivating dropdown in certain X11 WMs

isf63:
Adds a workaround for WMs who sometimes activate QTerminal dropdown in a minimized state

"Certain X11 WMs" being namely Xfwm4 and Openbox

To reproduce: https://github.com/lxqt/qterminal/issues/1066

----

Adding `raise();` did not suffice. This patch explicitly sets window state to not minimized when shown.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -831,6 +831,7 @@ void MainWindow::showHide()
         }
         realign();
         show();
+        setWindowState(windowState() & ~Qt::WindowMinimized);
         activateWindow();
     }
 }


tsujan:
I'm afraid, this is out of context. The dropdown window shouldn't be minimizable in the first place — we never minimize it in the code, and WMs shouldn't do it.

As I may have mentioned elsewhere, it isn't good to add workarounds for specific apps. The problem of Xfwm4 is too serious, and Openbox is dead.

isf63:
The problem with minimization on Xfwm4 (and Openbox) is that the dropdown mode appears in the taskbar.

In my estimation, this fixes half of the problem there.
>not good to add workarounds for specific apps

_If_ the code is harmless, why not?

tsujan:
> the dropdown mode appears in the taskbar.

As I said, that's a serious issue of Xfwm4 and should be fixed in it.

> In my estimation, this fixes half of the problem there.

At least for three reasons:

 * If "harmless" codes were allowed, they would result in a messy source.
 * There's no reason to sweep others' bugs under the carpet (especially when you say, "this fixes half of the problem there" — actually, it isn't a fix at all). The more visible they are, the sooner they might be fixed by their devs.
 * Each team of devs is responsible for their own apps.

The situation is quite different in the case of bugs in an *underlying library* (like Qt). In that case, workarounds are necessary, if feasible.

isf63:
It's irritating because Xfwm4/XFCE moves so slowly, and also Wayland is taking more of their manpower now. I experience this bug daily. It does look like they make point releases for release versions however, hopefully I will find time to make a bug report.

tsujan:
I completely understand your problem. You might want to use it as a personal patch. But it isn't fair to sacrifice our code integrity for an issue in a totally different app.

> hopefully I will find time to make a bug report.

I thought a report has already been made. How they could know about the issue without a report?!

If I were you, I'd also test yakuake. If it doesn't have the same problem, perhaps we could consult its code.

isf63:
Will look into yakuake. Maybe using KF6 KWindowSystem methods is more robust than Qt ones.

tsujan:
KWindowSystem is Qt+X11. It might have been started in the hope of providing parallel X11 and Wayland methods, but that idea was proven to be futile a long time ago. Depending on KF6 is not an option.

> Will look into yakuake.

I suggested that to see if I should read its source. Reading alien codes needs motivation.

isf63:
>I suggested that to see if I should read its source.

Oh, okay. That's good because my WM/compositor knowledge is ad-hoc.

When I install it fails with a QDBus error:
```
Couldn't start kglobalaccel from org.kde.kglobalaccel.service: QDBusError("org.freedesktop.DBus.Error.ServiceUnknown", "The name is not activatable")
Segmentation fault (core dumped)
```

Too bad because I suspect it doesn't have the problem.

stefonarch:
Thought it was possible like that but no:

```
 /usr/lib/kglobalacceld &                                                          
[1] 858176                                                                                          
stef@archlinux:~$ yakuake                                                                           
Couldn't start kglobalaccel from org.kde.kglobalaccel.service: QDBusError("org.freedesktop.DBus.Error.ServiceUnknown", "The name is not activatable")                                                   
KCrash: Application 'yakuake' crashing... crashRecursionCounter = 2                                 
[1]+  Completato              /usr/lib/kglobalacceld                                                
Errore di segmentazione (core dump creato)
```

tsujan:
> it properly hides in the taskbar

What do you mean? Is its button shown on the task-bar?

tsujan:
I mean, when it isn't hidden, does it have a task-bar button under Xfwm4?

isf63:
On Xfwm4 it never has the taskbar button. On Openbox (I know, unsupported, ...) it's a weird situation.

tsujan:
Thanks for testing! If I find the time, I might take a look at its code to see if it has added a Qt attribute to its window that we haven't — there will be nothing to do beyond that, because we neither want KF6 nor direct X11 codes.

tsujan:
I took a look at Yakuake's code. The only relevant differences I found was that Yakuake set the window flag to `Qt::Tool`, while we set it to `Qt::Dialog` and also set `Qt::WindowStaysOnTopHint`.

Since QTerminal has worked without problem outside Xfwm4, changing window flags doesn't seem wise to me, especially because we couldn't check all other X11 WM anew. However, you could experiment with them in `mainwindow.cpp` → `MainWindow::enableDropMode`.

Again, IMO, the bug should be fixed Xfwm4. Not only it shouldn't minimize the window (because we have no code for that), but also it shouldn't show the task button.

isf63:
Good news :)
Adding `Qt::Tool` makes dropdown behave exactly as Yakuake - never shows up in the taskbar, always opens when hitting the keybind. Except for triggering "Show Desktop" where here dropdown correctly hides.

I reverted by previous commit and simply added the flag to the list. Openbox behavior is also identical to Yakuake (and to what it was before).

tsujan:
Years ago, the dropdown window was the child of a hidden tool window — an odd combination I've seen elsewhere (perhaps, by the same coder). But that made the app exit after closing sub-dialogs of the properties dialog, which was a bad bug. So, the parent tool window was removed.

Please check the font dialog and also LXQt file dialog when called from the properties dialog of the dropdown window.

isf63:
Again, good news!

This combination of flags and inheritance doesn't do that. Tested Xfwm4 and Openbox. I will have to pause KWin testing as I don't want my filesystem being blown up currently.

tsujan:
KWin is important, but I can't test with it either, because KWin 6.3.0 has a nasty bug that causes a total freeze, whether under X11 or under Wayland.

I'll make a PR, without merging it soon.

tsujan:
More tests and discussions could be done here: https://github.com/lxqt/qterminal/pull/1236

PR #1236, opened 2025-02-18
Made the dropdown window a tool window under X11

tsujan:
… by adding `Qt::Popup` to the `Qt::Dialog` flag. The reason is that some X11 WMs may need it to hide the task button.

NOTE: This needs thorough tests under well-known X11 WMs that are being developed, especially by showing the properties dialog its sub-dialogs like the font dialog and LXQt file dialog, as well as the pasting and exiting prompt dialogs.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -190,6 +190,10 @@ void MainWindow::enableDropMode()
     }
 
     setWindowFlags(Qt::Dialog | Qt::WindowStaysOnTopHint | Qt::CustomizeWindowHint);
+    if (QGuiApplication::platformName() != QStringLiteral("wayland"))
+    { // some X11 WMs may need this
+        setWindowFlags(windowFlags() | Qt::Popup);
+    }
 
     m_dropLockButton = new QToolButton(this);
     m_dropLockButton->setToolTip(tr("Keep window open when it loses focus"));


tsujan:
These are also important for testing:

 * Context menus, inside the terminal or a sub-terminal, on tabs, and inside the search entry.
 * The menu-bar and its menus.

stefonarch:
Started some testing, xfwm4 is fine, but kwin_x11 shows it in the taskbar. We could exclude it now with the settings, but I didn't investigate if it's class is different from a normal qterminal window.

tsujan:
> as it always did

Really? I don't remember that. Does it get minimized on kwin_x11. I mean, does it have on kwin the problems @isf63 reported on Xfwm4?

Also, context menus are very important. Qt6 didn't show them in Fancy Menu.

isf63:
Context menus are shown well under Xfwm4. Testing KWin now...

isf63:
Can confirm Kwin shows a task button in the taskbar, but the original "minimize/restore" bug isn't there...

isf63:
To my understanding we support 2.5 X11 stacking WMs - (Xfwm4, Kwin, and hesitantly Openbox.)

The original issue is completely fixed in Xfwm4. In Kwin (and Openbox, same behavior currently) the dropdown window never activates in a minimized state, however the taskbar button is always shown. Yakuake's behavior on Kwin and Openbox is that the taskbar button is mostly not shown, however it does randomly appear on activation, moreso on Openbox.

tsujan:
What happens when the Preferences dialog is shown on X11?

isf63:
On all stacking WMs the Preferences dialog works correctly and doesn't show a task button.

stefonarch:
I remember that I always needed a window rule for kwin on X11 to exclude it from the taskbar.

tsujan:
And I remember that I didn't use the dropdown mode until I came to Wayland ;)

isf63:
I forgot to mention that also Openbox and Kwin on X11 cannot maximize the dropdown window, with either F11 or the menu. Xfwm4 works here too however.

tsujan:
> Openbox and Kwin on X11 cannot maximize the dropdown window

You mean making it fullscreen, which reminds me of the fact that fullscreening of the dropdown window doesn't and can't work on Wayland directly, because of anchors and margins. A workaround was possible, but I didn't add it, because it was ugly, and a fullscreen dropdown window wasn't needed.

PR #1240, opened 2025-02-27
Set the dropdown minimum size to 25%

tsujan:
This is especially needed on Wayland because of a Qt bug that lets Qt windows become tiny. But it's good for X11 too.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -1032,6 +1032,12 @@ To remove/disable a Shortcut, at point 2 press only a modifier (like Shift)</str
               <property name="suffix">
                <string>%</string>
               </property>
+              <property name="minimum">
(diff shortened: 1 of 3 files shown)


stefonarch (approved):
Manually inserting was weird, I see the spinbox blocks at 25% 
 but I used 10% and it  jumps back to the previous value (90%). Ideally when inserting  values  <25% it should set it to 25% and not to the previous value, no idea if that's possible.

isf63:
`25%` should suffice for even ultrawide monitors. On X11 I can set width to `1%` (1920x1200) where it's completely unusable, but Qt intelligently increases the width to a non-bugged state. I _believe_ the most common ultrawide resolution is 2560x1080; `25%` of that is 640px which is still incredibly narrow for modern standards.

PR #1241, opened 2025-03-02
Give better title to drop-down desktop file

isf63:
Change title, "QTerminal drop down" → "QTerminal Drop-down".

Translations are again questionable. I'm guessing many languages will already be correct as they have their own word for "dropdown".

### DIFF
--- a/src/translations/qterminal-drop.desktop.yaml
+++ b/src/translations/qterminal-drop.desktop.yaml
@@ -1,3 +1,3 @@
-Desktop Entry/Name: "QTerminal drop down"
+Desktop Entry/Name: "QTerminal Drop-down"
 Desktop Entry/GenericName: "Drop-down Terminal"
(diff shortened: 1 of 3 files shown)


stefonarch:
There should be no issues as those are `.yaml` files, which then will show up on weblate only as "Needs editing", which I can batch remove with a plugin. Only in `*.ts` files the translations are removed on source changes and `lxqt2-transupdate`.

stefonarch:
This should be changed too, no?
https://github.com/lxqt/qterminal/blob/4095f621fc484f89ad66ef2d35abb06e02ed0592/src/translations/qterminal.desktop.yaml#L4

isf63:
It's unclear whether those actions should be title-case. BTW, I only today discovered them in Fancy Menu.

Might as well, though, because many desktop files such as Firefox and Libreoffice do use title-case.

Issue #1247, opened 2025-03-10
Shorcuts cannot be set with a profile


tsujan:
##### Steps to Reproduce (for bugs)

1. Load a profile with `qterminal -p PROFILE_NAME`.
2. Change a shortcut, like that of "Bottom Subterminal".
3. Press OK in Preferences dialog.

Then the new shortcut won't work. Actually, if you open Preferences dialog, you'll see the old shortcut.

##### Solution

The reason is that shortcuts are read only from the main config file. I'll make a PR to fix that.

##### System Information

* QTerminal version: Any
* Distribution & Version: Any
* Kernel: Any
* Qt Version: Any

PR #1249, opened 2025-03-10
Fixed shortcut setting with a profile

tsujan:
Previously, shortcuts that were set with a profile had no effect.

Fixes https://github.com/lxqt/qterminal/issues/1247


tsujan:
@stefonarch 
This is the last one in the series, I hope.

stefonarch (approved):
Other settings are fine I suppose, at least the ones I used in my "small" terminal worked.

tsujan:
Yes, other settings are OK. Shortcuts were an exception because of the way they were read. The patch reads them like the other settings, i.e., by considering the profile, if any.

PR #1250, opened 2025-03-16
Rename dropdown for desktop files

isf63:
Not sure why I chose `Drop-down` previously.

`Dropdown` is a standard word in English for user interfaces.


stefonarch:
> Merriam-Webster lists “drop-down” and “dropdown” as equal variants for the noun form (the adjective form is always hyphenated)

I think drop-down is ok everywhere, especially as adjective here.

tsujan:
> I think drop-down is ok everywhere

I agree.

isf63:
Alright, they're essentially the same.

PR #1255, opened 2025-04-09
Bumped copyright to 2025

stefonarch:


tsujan (approved):
GTM, but there's really no need to such changes every year. The second date does *not* mean the end of the copyright year but just a new start.

stefonarch:
It does just look old if clicked, most people won't.

PR #1263, opened 2025-05-10
Add option for saving window state

dernett:
Adds a checkbox option in Preferences -> Window. When unchecked, the window's state will not be saved to the settings file.

This is useful when the settings file is managed with version control.


tsujan (approved):
Thanks! GTM.

While reading your patch, I saw an old problem in the original code. Will fix it in a small PR after merging this.

PR #1265, opened 2025-05-11
Respect geometry and state options on closing with prompt

tsujan:
Previously, the options for not saving position (on X11), size or state weren't respected when the terminal was closed after showing the prompt.

Also, the maximized state should be remembered only when the size is remembered.


tsujan:
Straightforward. Merging…

Issue #1274, opened 2025-06-27
Bookmarks top branch is visible when bookmarks are shown by right clicking menubar


isf63:
These are the bookmarks on my install:

![](https://i.postimg.cc/3RqQxCj1/qterminal.png)

The first column seemingly is not resizable, and is too small. Resizing the bookmarks pane has no effect.

It is the same for dropdown and windowed.

##### Expected Behavior
Bigger default width for the first column and able to be resized

##### Current Behavior
May be specific to Qt 6.8 (?)

##### Steps to Reproduce (for bugs)
1. Have `.config/qterminal.org/qterminal_bookmarks.xml` unchanged from the preset.
2. Open Bookmarks
3. Resize width of Bookmarks pane

##### System Information
* QTerminal version: git
* Distribution & Version: Ubuntu 25.04
* Qt Version: 6.8.3+dfsg-0ubuntu2

isf63:
I investigated some, the reason is it not resizable is because the header is hidden.

With the header shown:

![](https://i.postimg.cc/nhXRHbRy/qterminal.png)

tsujan:
It doesn't happen here; there's no eliding at all.

Are you sure you've compiled QTerminal and qtermwidget from git yourself and without any extra (Ubuntu) patch? If yes, more info is needed.

tsujan:
Strange! Your installed QTerminal should be very old. See https://github.com/lxqt/qterminal/pull/893.

isf63:
Then the only thing I can think of is Qt=6.8.x (?)

Unless something is very wrong with my scripts, I'm on
```
Package: qterminal-git
Version: 2.2.1.r7.g5f09ad3-1
```

tsujan:
That "Synchronized Bookmarks" was made hidden in 2022 (→ https://github.com/lxqt/qterminal/pull/893). So, I really doubt that you have 2.2.1.

isf63:
Help -> About says "2.2.1"

For qtermwidget
```
Package: qtermwidget-git
Version: 2.2.0.r1.g1fea00c-1
```

tsujan:
That was the missing info :)

PR #1275, opened 2025-06-28
Added bookmarks action to view context menu

tsujan:
… and removed the menubar context menu.

The reason is that the menubar context menu had a single item that showed bookmarks *regardless of Preferences*. It was possible to control it, but that would require extra codes for something useless when toggling bookmarks can be added to the view context menu.

Closes https://github.com/lxqt/qterminal/issues/1274

### DIFF
--- a/src/forms/qterminal.ui
+++ b/src/forms/qterminal.ui
@@ -54,6 +54,9 @@
      <height>26</height>
     </rect>
    </property>
+   <property name="contextMenuPolicy">
+    <enum>Qt::PreventContextMenu</enum>
+   </property>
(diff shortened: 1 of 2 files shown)


tsujan:
That context menu is created in the code of `QMenuBar`. When we can easily set `Qt::ContextMenuPolicy`, there's no reason to do more.

tsujan:
@isf63 
From your comment I assumed you approved this PR, but for approving a PR completely, you could go to the "Files changed" tab in the browser, Press "Review changes", check "Approve", and then press "Submit review" — if needed, after adding a comment.

PR #1289, opened 2025-10-03
Modify custom tab name behaviour

chrcuray:
Modify custom tab name behavior:
    use available session name during session edit (was empty)
    restoring to terminal title when custom name is empty (was not possible)

After comments on the https://github.com/lxqt/qterminal/pull/678 , the change was rewritten and the described use-case has been considered.


tsujan:
I found time to have a quick look (without testing it):

 * For whatever reason, the original code removes the tab icon only when the new text is set. Please keep that behavior in your PR.
 * Now that you're at it, please exclude the trailing and leading spaces from `text`. Particularly, we don't want a name like `   `; do we?

PR #1328, opened 2026-03-14
Silenced layershell and QFile warnings

stefonarch:


isf63:
For the layer-shell warnings, `setWantsToBeOnActiveScreen` doesn't exist with my layer-shell-qt 6.4.5, needing ≥ 6.5.0 or 6.6.0. Updating the minimum version would exclude Debian Stable users. Also https://github.com/lxqt/lxqt-notificationd/pull/457

stefonarch:
There are more, lxqt-leave and and lxqt-runner. But LXQt 2.4 won't be on debian stable ever anyway. But yes, we need to make sure that everything compiles on debian forky/testing. No idea if LXQt 2. 4 will make it into it though.

src/mainwindow.cpp
```diff
@@ -186,7 +186,7 @@ void MainWindow::enableDropMode()
                                                          | LayerShellQt::Window::AnchorRight};
                 m_layerWindow->setAnchors(anchors);
                 m_layerWindow->setScope(QStringLiteral("dropdown_terminal"));
-                m_layerWindow->setScreenConfiguration(LayerShellQt::Window::ScreenConfiguration::ScreenFromCompositor);
+                m_layerWindow->setWantsToBeOnActiveScreen(true);
```
luis-pereira, line 189: `setWantsToBeOnActiveScreen` was introduced [here](https://invent.kde.org/plasma/layer-shell-qt/-/commit/430ad3630fc648a9216a45a9f68511d0a7981281).
It was not released yet!

luis-pereira (changes requested):

src/mainwindow.cpp
stefonarch: >  It was not released yet!

What do you mean? I get the deprecated warnings for quite some time now, it's in 6.6
https://repology.org/project/layer-shell-qt/versions

luis-pereira: `git tag --contains 430ad3630fc648a9216a45a9f68511d0a7981281`  returns none.
But.. it was cherry-picked to landing at `60b7d27ae0890e3cf60f6237fdb7c1ac5e8ac5f3` . That's why I don't like cherry-picks.

``` 
git tag --contains 60b7d27ae0890e3cf60f6237fdb7c1ac5e8ac5f3
v6.5.91
v6.6.0
v6.6.1
v6.6.2
v6.6.3
```

I'm running `v6.5.6` released on 2026-03-10 and get a FTBFS.
`v6.5.91` was released on  2026-01-27 has it!

stefonarch: We could skip it for LXQt 2.4, it's just a warning.

luis-pereira: Agreed.

src/main.cpp
```diff
@@ -180,9 +180,12 @@ int main(int argc, char *argv[])
...
         QFile style(customStyle.canonicalFilePath());
-        style.open(QFile::ReadOnly);
-        QString styleString = QLatin1String(style.readAll());
-        app->setStyleSheet(styleString);
+        if (style.open(QFile::ReadOnly)) {
+            QString styleString = QLatin1String(style.readAll());
+            app->setStyleSheet(styleString);
+        } else {
+            qWarning() << "Failed to open stylesheet";
+        }
```
tsujan, line 186: Let's not bother the user with this `qWarning()`, especially because a style sheet may be inapplicable for other reasons too, like a bad CSS.

Also, please bump `SHELLQT_MINIMUM_VERSION` to 6.6.0 in the main cmake file.

stefonarch: >Also, please bump SHELLQT_MINIMUM_VERSION to 6.6.0 in the main cmake file.

This is done here: https://github.com/lxqt/qterminal/pull/1334

tsujan: > This is done here: https://github.com/lxqt/qterminal/pull/1334

Oh, I hadn't seen it. It's OK then. Just remove that `qWarning()`.

stefonarch: Done.

tsujan: > This is done here: https://github.com/lxqt/qterminal/pull/1334

Am I missing something? I don't see see it in https://github.com/lxqt/qterminal/pull/1334, neither `SHELLQT_MINIMUM_VERSION` nor `QT_MINIMUM_VERSION`.

stefonarch: I made them visible now :)

PR #1345, opened 2026-08-04
tr: Use plain names for cursor shapes

isf63:
In the GUI:
- "BlockCursor" → "Block"
- "UnderlineCursor" → "Underline"
- "IBeamCursor" → "I-Beam"

This will reset these translations but I think it's gainful to not use identifier-type strings.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -532,9 +532,9 @@ void MainWindow::setup_ViewMenu_Actions()
     /* Keyboard cursor shape */
     if (keyboardCursorShape == nullptr) {
         keyboardCursorShape = new QActionGroup(this);
-        QAction *block = new QAction(tr("&BlockCursor"), this);
-        QAction *underline = new QAction(tr("&UnderlineCursor"), this);
-        QAction *ibeam = new QAction(tr("&IBeamCursor"), this);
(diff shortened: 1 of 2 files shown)


stefonarch:
Not sure if it is worth it, many translations are without "cursor":
https://translate.lxqt-project.org/search/lxqt-desktop/qterminal/?offset=&q=beam&sort_by=-priority%2Cposition&checksum=

isf63:
>Not sure if it is worth it, many translations are without "cursor"

Scanning that page, it's about 50% or more that translate "Cursor", plus translating literally (same capitalization, no space ...)

tsujan:
IMO, this change makes perfect sense.

As for translations, there may be a compromise: @stefonarch could preserve all of them in another PR — if that's not hard to do, of course. I don't think that will be a UX problem, because users who don't use English will see no change.

stefonarch:
Ok, maybe better not preserving translations as some are ok and some not.
But it will remove all translations so it needs retranslating.

isf63:
Go ahead with merging?

tsujan:
> Go ahead with merging?

Only if @stefonarch agrees. Translations are important parts of GUI.

stefonarch:
I  gave already the ok, go.

PR #1347, opened 2026-08-06
Updated *.ts files

stefonarch:


stefonarch:
It looks like there are 2 locations where the translation is used, didn't manage to find the other one. @tsujan ?

tsujan:
Yes, in `mainwindow.cpp` with ampersand, and in `propertiesdialog.cpp` without ampersand. I see both of them in this PR.

stefonarch:
I wondered if it's ok in the GUI, found it:

immagine

PR #1351, opened 2026-08-11
tabwidget: connect currentChanged->currentTitleChanged once in ctor, fix #357

### LINKED ISSUE #357: Performance issues after opening and closing a lot of tabs
This has been bugging me forever and I'm only now bothering to file an issue about it.

##### Steps to Reproduce

1. Open a new QTerminal window
2. Open 3 tabs
3. Hold shift-rightarrow to quickly switch between the tabs, verify that it's snappy.
4. Open, then immediately close, 50-100 more tabs
5. Do the same procedure as earlier: Hold shift-rightarrow to quickly switch between the tabs, performance now sucks.

KrisTHL181:
Fixes #357

The forward connection was created inside addNewTab() on every new tab and never disconnected, so after N tabs each currentChanged() fired N currentTitleChanged() emissions, each running MainWindow::onCurrentTitleChanged (a ~50ms setWindowIcon on X11). Tab switching and creation became O(total tabs ever created); closing tabs did not help because the connections stayed on the TabWidget itself.

Move the signal-to-signal forward into the constructor so it is connected exactly once for the widget's lifetime.


tsujan:
Indeed the connection was so misplaced. The problem is very old but its effect wasn't noticed on relatively fast machines. Thanks a lot for finding and fixing it! I'll merge the PR after some checks…

Issue #1123, opened 2024-05-12
[Feature request] Open Qterminal with custom tab, title, cmd


Angell1993:
I've been trying to get Qterminal to open with custom specific tabs, titles and commands, but this is as close as I'm able to get with python and dbus, it would be lovely if this could be implemented..

`import subprocess

from pydbus import SessionBus
from gi.repository.GLib import Variant
from pprint import pprint

proc = subprocess.Popen(['qterminal'])
dest = 'org.lxqt.QTerminal-' + str(proc.pid)

subprocess.check_call(['gdbus', 'wait', '--session', dest])

bus = SessionBus()
process = bus.get(dest, '/')['org.lxqt.QTerminal.Process']
current_window = process.getWindows()[0]

window = bus.get(dest, current_window)['org.lxqt.QTerminal.Window']
#window.title = "Terminal000"; #Missing Property

#Will add the first tab on launch then use window.newTab() for as many as you need
window.newTab({'shell': Variant('s', '/bin/bash')}) #Only default shell doesn't use input
window.newTab({'shell': Variant('s', '/bin/fish')}) #Only default shell doesn't use input
#window.newTab({'shell': Variant('s', '/bin/bash')})

print({window}) #Will print Window object
pprint(dir(window)) #Will print Window object available properties

tab000 = window.getTabs()[0]
print({''})
print({tab000}) #Will print Tab/Terminal object
pprint(dir(tab000)) #Will print Tab/Terminal object available properties
#tab000.title = "Steam" #Read Only
#tab000.cmd = "/usr/bin/steam-runtime %U -silent" #Missing Property - Should run command after shell..

tab001 = window.getTabs()[1]
#tab001.title = "Scream-VM" #Read Only
#tab001.cmd = "scream -m /dev/shm/scream-ivshmem" #Missing Property - Should run command after shell..

tab002 = window.getTabs()[2]
#tab002.title = "Scream" #Read Only
#tab002.cmd = "scream" #Missing Property - Should run command after shell..

proc.wait()`

florent4014:
I agree, it would be very nice to have something like this.

Issue #980, opened 2022-08-01
pssfix the name in the tabs


mc36:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe eventual alternatives you've considered**

Issue #812, opened 2021-04-07
[QUESTION] Is it possible to create a few instances of qterm with some tabs in each?


Axisar:
Hello there.
Is it possible to create a few instances of qterm with some tabs in each instance with executed command in each with changing location before execute using bash script of bash oneliner
like: qterminal -tab0 -w /somelocation -e somecommand -tab1 -w /somelocation1 -e somecommand1 and etc?

yan12125:
It is possible, but quite difficult. DBus APIs are needed. Here is an example in Python:

```Python
import subprocess

from pydbus import SessionBus
from gi.repository.GLib import Variant

proc = subprocess.Popen(['qterminal'])
dest = 'org.lxqt.QTerminal-' + str(proc.pid)

subprocess.check_call(['gdbus', 'wait', '--session', dest])

bus = SessionBus()
process = bus.get(dest, '/')['org.lxqt.QTerminal.Process']
current_window = process.getWindows()[0]

window = bus.get(dest, current_window)['org.lxqt.QTerminal.Window']
current_tab = window.getTabs()[0]
window.newTab({'shell': Variant('s', '/usr/bin/python')})
window.newTab({'shell': Variant('s', '/bin/bash')})
tab = bus.get(dest, current_tab)['org.lxqt.QTerminal.Tab']
tab.closeTab()

proc.wait()
```

To make things worse, there are currently no documents about DBus APIs. One needs to read `src/org.lxqt.QTerminal.*.xml` files to see what APIs are available - and maybe even C++ sources to understand actual functionality of those APIs. Many things can be improved, so I keep this issue open as a reminder.

Issue #531, opened 2018-12-30
A few settings commonly found in other terminal emulators


Karl-D-Asmussen:
QTerminal looks good, but lacks a couple of otherwise ubiquitous settings.

Almost every other TE available allows the user to:
- Completely disable all scrollback (popular with purists/masochists)
- Specify initial size in terms of character cells

(Using 0.8.0, from Ubuntu upstream)

yan12125:
As a side note, the second feature is discussed in #492 too.

cxw42:
I just recently started using lxqt and am enjoying it!  I would like to request the feature that XTerm has to switch between the regular screen and the alternate screen ([ref](https://invisible-island.net/xterm/xterm.faq.html#xterm_tite)).  I have looked in the manual, issues, Google results, and preferences, and have not been able to find this function.  Thank you for considering this request!

yan12125:
@cxw42 Alternate screens are supported and enabled unconditionally in qtermwidget/qterminal. Mind to share more details for your use cases?

cxw42:
@yan12125 Thanks for asking!  Yes, the alternate screen works fine.  In xterm, I have appreciated the ability to switch back and forth without disrupting application state.  E.g.,

- Normal screen: shell prompt, and the output from a previous `git log` command
- Alternate screen: `less`, showing results of an `ag` search over a large codebase.

By switching between normal and alternate screens, I can refer to the log results in deciding how to use my search results.

Alternatives:

- Ctl-z the `less`, then `fg` back.  However, this scrolls the normal screen.  Therefore, if I switch back and forth several times, information I need on the normal screen may scroll off the top of the window.
- Multiple terminals or tabs.  However, the former requires more screen space, and both require keeping two shell sessions in sync (e.g., same working directory, sometimes same custom env cars).

Please let me know whether or not this example is helpful.  If I am not explaining clearly enough, I can try a different way of conveying the information.  Thanks!

yan12125:
Hmm, the xterm page you've mentioned talks about enable/disable the alternate screen. I guess what you want is somewhat different - switching between normal and alternate screens without Ctrl-Z/fg hacks?

cxw42:
Correct- sorry for the confusion.  I should have quoted from that link:

> xterm allows you ... to switch the display back to the alternate screen to select text from it, to paste into the normal screen

And it works the other way as well (switch to normal then paste into alternate).

yan12125:
Got it! Thanks for the confirmation.

[...]

PR #1353, opened 2026-08-16
Ipc

luebking:
QTerminal has a dbus interface but it's of extremely limited use.

This series of patches adds some **methods to control the terminal look and feel via dbus** (there's already a sendText function, so this should be no security concern), methods to **create new windows and tabs that are usable via qdbus** (which does not support QVariantMap) and allows **scriptable access** to the dbus address, either by **predetermining service and object path** resp. also by exporting those into the **clients environment**.

This allows to control special terminals (running mutt, top or your qnd shell) to be accessed via keyboard shortcuts, create new windows and tabs and distinguish them by color, font or background image, change between bright and dark schemes on the fly (for daytime awareness) and stuff that I haven't figured yet (but because sendText is unfiltered for newlines you should theoretically be able to remote control mutt for mailto: links - now that you know which terminal runs mutt ;)

~There's likely **requirement for fix-up**, because the present **coding style** is all over the place and I forgot what the preferred one is :)
CONTRIBUTING.md might benefit from addressing that *cough*~

Also it's not strictly required but you might prefer to hide API and command switch behind the preprocessor token.

Also
https://github.com/lxqt/qterminal/issues/1123
https://github.com/lxqt/qtermwidget/issues/507
https://github.com/lxqt/qterminal/issues/812
https://github.com/lxqt/qterminal/issues/980
https://github.com/lxqt/qterminal/issues/301 (but that's technically unrelated and a QTermWidget deficiency)

Todo:
~set tab (and window?) titles (`printf "\033]0;%s\007" "foobar"`usually doesn't work because the shell will reset it)
make splitting more accessible (qdbus compatible and immediate proportion control)
size parameter for new instance (might also make sense for newWindow to allow informed window placement), https://github.com/lxqt/qterminal/issues/531~

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,7 +39,7 @@
 
 #define out
 
-const char* const short_options = "vhw:e:dp:";
+const char* const short_options = "vhw:e:dp:i:";
 
 static const char* serviceName = "org.lxqt.QTerminal";
 static const char* ifaceName = "org.lxqt.QTerminal.Process";
@@ -51,6 +51,7 @@ const struct option long_options[] = {
     {"execute", 1, nullptr, 'e'},
     {"drop",    0, nullptr, 'd'},
     {"profile", 1, nullptr, 'p'},
+    {"dbus_id", 1, nullptr, 'i'},
     {nullptr,   0, nullptr,  0}
 };
 
@@ -63,6 +64,7 @@ QTerminalApp * QTerminalApp::m_instance = nullptr;
     puts("  -d,  --drop               Start in \"dropdown mode\" (like Yakuake or Tilda)");
     puts("  -e,  --execute <command>  Execute command instead of shell");
     puts("  -h,  --help               Print this help");
+    puts("  -i,  --dbus_id <name>     Register with predetermined dbus interface, org.lxqt.QTerminal-<id>");
     puts("  -p,  --profile <name>     Load profile from ~/.config/<name>.conf");
     puts("  -v,  --version            Prints application version and exits");
     puts("  -w,  --workdir <dir>      Start session with specified work directory");
@@ -77,7 +79,7 @@ QTerminalApp * QTerminalApp::m_instance = nullptr;
     exit(code);
 }
 
-void parse_args(int argc, char* argv[], QString& workdir, QStringList & shell_command, out bool& dropMode)
+void parse_args(int argc, char* argv[], QString& workdir, QStringList & shell_command, out bool& dropMode, QString &dbus_id)
 {
     int next_option = 0;
     dropMode = false;
@@ -106,6 +108,9 @@ void parse_args(int argc, char* argv[], QString& workdir, QStringList & shell_co
                 break;
             case 'p':
                 Properties::Instance(QString::fromLocal8Bit(optarg));
+                break;
+            case 'i':
+                dbus_id = QString::fromLocal8Bit(optarg);
                 break;
             case '?':
                 print_usage_and_exit(1);
@@ -153,10 +158,11 @@ int main(int argc, char *argv[])
     QString workdir;
     QStringList shell_command;
     bool dropMode = false;
-    parse_args(argc, argv, workdir, shell_command, dropMode);
+    QString dbus_id;
+    parse_args(argc, argv, workdir, shell_command, dropMode, dbus_id);
 
     #ifdef HAVE_QDBUS
-        app->registerOnDbus(dropMode);
+        app->registerOnDbus(dropMode, dbus_id);
     #endif
 
     if (!app->isPrimaryInstance())
@@ -302,7 +308,7 @@ QList<MainWindow *> QTerminalApp::getWindowList()
 }
 
 #ifdef HAVE_QDBUS
-void QTerminalApp::registerOnDbus(bool dropDown)
+void QTerminalApp::registerOnDbus(bool dropDown, QString dbus_id)
 {
     if (!QDBusConnection::sessionBus().isConnected())
     {
@@ -324,8 +330,8 @@ void QTerminalApp::registerOnDbus(bool dropDown)
     }
     else
     {
-        if (!QDBusConnection::sessionBus().registerService(QLatin1String(serviceName)
-                                                           + QStringLiteral("-%1").arg(getpid())))
+        const QString suffix = dbus_id.isEmpty() ? QStringLiteral("-%1").arg(getpid()) : QStringLiteral("-%1").arg(dbus_id);
+        if (!QDBusConnection::sessionBus().registerService(QLatin1String(serviceName) + suffix))
         {
             fprintf(stderr, "%s\n", qPrintable(QDBusConnection::sessionBus().lastError().message()));
             return;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -36,10 +36,11 @@
 #include "qterminalapp.h"
 #include "qterminalutils.h"
 #include "terminalconfig.h"
+#include "termwidget.h"
 
 #define out
 
-const char* const short_options = "vhw:e:dp:i:";
+const char* const short_options = "vhw:e:dp:i:s:";
 
 static const char* serviceName = "org.lxqt.QTerminal";
 static const char* ifaceName = "org.lxqt.QTerminal.Process";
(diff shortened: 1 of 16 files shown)


tsujan:
It looks promising; thanks! Thorough tests will be needed of course.

> the present coding style is all over the place and I forgot what the preferred one is :)

Yes, I've had that headache too. Recently we agreed on this for new codes:

```c++
if (a)
{
    b;
}
else
{
    c;
    d;
}
```

But changing the style of the whole code will be a nightmare, to say nothing of reviewing. So, inconsistencies might persist indefinitely.

luebking:
"QString& foo" or "QString &foo" (or we don't care because we don't have a major OCD :laughing: )?
"dbus_id" or "dbusId" ? (That's a stroker => CamelCase except when it looks weird?)

tsujan:
> "QString& foo" or "QString &foo" (or we don't care because we don't have a major OCD 😆 )?

Yes, it's a jungle :D  I usually check what's often used in the source file I work on, but inconsistencies exist everywhere in QTerminal/QTermWidget.

Only the original creator of LXQt had a fully consistent coding style in libfm-qt, pcmanfm-qt, lximage-qt, … (but QTerminal wasn't his area). Although I tried to imitate that style in those apps, I made "mistakes". Worse than that, I have three different coding styles in my own apps, but at least they're self-consistent ;)

luebking:
Looking around in the bugs: https://github.com/lxqt/qtermwidget/issues/472 :rofl: 
And https://github.com/lxqt/qterminal/issues/311

src/org.lxqt.QTerminal.Terminal.xml
```diff
@@ -19,6 +19,15 @@
         <arg name="text" type="s" direction="in"/>
     </method>
     <method name="closeTerminal"/>
+    <method name="setColorScheme">
+      <arg name="text" type="s" direction="in"/>
+    </method>
+    <method name="setBackgroundImage">
+      <arg name="image" type="s" direction="in"/>
+    </method>
+    <method name="setBackgroundMode">
```
luebking: I'm thinking about merging the two setBackground* methods

pros:
- less overhead on dbus and terminal update
- you're unlikely to change the mode w/o the image
con:
- the mode is a magic number and altering it might be irrelevant if only using this change a backdrop logo

Solution?
Treat mode < 0 as "keep the current one"?

src/termwidget.cpp
```diff
@@ -442,4 +442,21 @@ void TermWidget::setFont(const QString& font, const int pointSize)
...
 }
 
+void TermWidget::setSize(const int columns, const int lines)
+{
+    if (impl())
+    {
+        impl()->setSize(QSize(columns, lines));
+        QSize sh = impl()->sizeHint();
+        // https://github.com/lxqt/qtermwidget/issues/656
+        for (const QWidget *kid : impl()->findChildren<QWidget*>()) {
```
luebking: This is of course a baghira memorial hack, but w/o changes to QTermWidget also a necessity to make this happen.
🤷

tsujan:
Please don't forget to signal to us by removing the draft state when the PR is complete.

tsujan:
I did some preliminary tests by using QDbusViewer. Most things worked as expected, but I found a problem:

When a new window is opened from a drop-down window, it should have a separate process, and it does when it's opened from the GUI (see https://github.com/lxqt/qterminal/pull/1154). But when I do it through QDbusViewer, it shares the same process and causes problems.

tsujan:
Oh, this issue isn't new — it's just doubled here — I hadn't noticed it because I hadn't tested qterminal with dbus thoroughly. So we might want to fix it in another PR later. Please ignore my previous comment.

luebking:
I've a patch but would be relieved if I don't have to merge that into the series (because of the later size control addition) and provide it as "fix newWindow dbus method for dropdown mode" on top of the stack :)

tsujan:
> I've a patch

IMHO, it may be better to remove it from this PR and make another PR based on it after merging this one. That seems much cleaner to me. What do you think?

luebking:
I really have no opinion on how patches get merged - you call the shots.
If you'd prefer a dedicated PR, I'll remove it with the other amendments. Or as only action if there's really nothing else.

tsujan:
I don't have a strong opinion; just thought we could find it more easily later if it was a separate PR. Do it as you see fit.

luebking:
Your first instinct was to isolate it, I'll do that and remove the patch from the PR when pushing the next round of amendments.

src/main.cpp
```diff
@@ -63,7 +66,9 @@ QTerminalApp * QTerminalApp::m_instance = nullptr;
     puts("  -d,  --drop               Start in \"dropdown mode\" (like Yakuake or Tilda)");
     puts("  -e,  --execute <command>  Execute command instead of shell");
     puts("  -h,  --help               Print this help");
+    puts("  -i,  --dbus_id <name>     Register with predetermined dbus interface, org.lxqt.QTerminal-<id>");
     puts("  -p,  --profile <name>     Load profile from ~/.config/<name>.conf");
+    puts("  -s,  --size <name>     Load profile from ~/.config/<name>.conf");
```
tsujan: Please fix this line.

src/termwidget.cpp
```diff
@@ -397,4 +426,60 @@ void TermWidget::sendText(const QString& text)
...
+
+void TermWidget::setFont(const QString& font, const int pointSize)
+{
+    if (impl())
+    {
+        impl()->setTerminalFont(QFont(font, pointSize < 0 ? impl()->getTerminalFont().pointSize() : pointSize));
+    }
+}
+
+void TermWidget::setSize(int columns, int lines)
```
tsujan, line 457: It doesn't seem to work as expected. For example, `qterminal -s 20x20` gives me this:

Image

luebking: Figured, It's because of the QMainWindow / central widget mechanics (the window size before the show event is disjunct from the central widget and the terminal inside)

Looking into it…

stefonarch:
Playing I can manage to make the dropdown window a ghost by closing last tab or closing it's window. 

immagine

No way to reactivate it with F12 (Wayland, just sending `qterminal -d` by shortcut), have to kill the process.

src/termwidget.cpp
tsujan: My 2 cents: Isn't it better to get the size in pixels, as in Terminal settings → window?

tsujan:
> I can manage to make the dropdown window a ghost by closing last tab or closing it's window.

Confirming. I hadn't checked it with the dropdown window before.

src/termwidget.cpp
luebking: Controlling the terminal size in cells is normal for TE and kinda important because clients frequently have requirements about the minimal or specific dimensions in characters.

If this was about the window interface, "qterminal -geometry 400x400" would already work.
I'lll just fight qmainwindow to it ;)

luebking:
Closing how?
If I alt+f4 or ctrl+d the terminal/window, the dropdown process quits.
(I'm not at hand sure what patch could cause this - the dbus interface to close tabs  existed before)

stefonarch:
>Closing how?

Closing with qdbus. Closing by ctrl+d or tabbutton or whatever quits the process too.

luebking:
I suppose that will then become part of the series to fix up dbus for the dropdown mode (together w/ the newWindow patch dicussed before) - I didn't touch the close method at all :)

stefonarch:
Yes, I checked, it's not your baby, it's an old baby ;)

src/termwidget.cpp
tsujan: It makes sense.

tsujan:
> it's an old baby ;)

Good to see that old problems are revealed by testing this PR.

The code is good. So I suggest we merge it after @luebking fixes the size issue.

src/termwidget.cpp
luebking: Good news is that I fixed the glitch.
Problem turned out to be QTabWidget which is lazy on setting its "layout" - simple solution after lots of brain-wrecking was to move the resize into the MainWindow's showEvent :smile:

luebking:
The update should have removed the dropmode patch (I'll look into the window that PR later - I'm already late) and fix the size handling (and add two blanks)

tsujan:
The size issue is fixed here. Thanks!

Please correct this line too:
```c++
puts("  -s,  --size <name>        Load profile from ~/.config/<name>.conf");
```
` <name>` and the description are copied from `--profile`.

luebking:
Done, sorry for being blind.

tsujan (approved):
GTM. Thank you so much!

Let's give it to git users …

PR #1355, opened 2026-08-18
Polish "Handle history…" feature

### LINKED ISSUE #1272: [Feature request] save history
Is there a way to save the history or select it all and copy it (eg. select the top, press shift and use the scroll bar to the bottom to select it all).

luebking:
* Re-label "Handle history" to less vague "View history in editor",
Clever users can figure that other actions can be taken on the file
Everybody else just has a menu entry that does something ominous and apparently nothing
* Use a temporary file location and auto-delete the file.
The scrollback buffer may hold sensitive data and the user can save it from the editor if they want
* Default the command to $VISUAL
* Warn the user if no editor command is set

fixes #1272

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -352,7 +352,7 @@ void MainWindow::setup_ActionsMenu_Actions()
     setup_Action(FIND, new QAction(QIcon::fromTheme(QStringLiteral("edit-find")), tr("&Find..."), settingOwner),
                  FIND_SHORTCUT, this, SLOT(find()), menu_Actions);
 
-    setup_Action(HANDLE_HISTORY, new QAction(QIcon::fromTheme(QStringLiteral("handle-history")), tr("Handle history..."), settingOwner),
+    setup_Action(HANDLE_HISTORY, new QAction(QIcon::fromTheme(QStringLiteral("handle-history")), tr("View history in editor"), settingOwner),
                  NULL, this, SLOT(handleHistory()), menu_Actions);
 
 #if 0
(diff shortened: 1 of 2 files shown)


stefonarch:
Looks working fine, I actually never understood what this was for ;)
GTM

src/mainwindow.cpp
```diff
@@ -851,7 +851,20 @@ void MainWindow::find()
...
+    {
+        QMessageBox::information(this, tr("No editor available"),
+                        QStringLiteral("<h1>:(</h1>")
+                        + tr("Please either set the $VISUAL environment or configure an editor command in the preferences."));
+        return;
+    }
+
+    QString dir = QStandardPaths::writableLocation(QStandardPaths::RuntimeLocation);
+    if (dir.isEmpty())
+        dir = QStandardPaths::writableLocation(QStandardPaths::TempLocation);
```
tsujan, line 867: I fully agree on not using `QStandardPaths::CacheLocation`, but why `QStandardPaths::RuntimeLocation`? Why not using `QStandardPaths::TempLocation` directly?

luebking: The runtime location is private to the user, /tmp is an open playground and most likely 022 (ie. your file will be created world readable)
Unfortunately the RuntimeLocation isn't universally available so we need to failsafe to TempLocation

Both are nowadays typically temporary filesystems in RAM

tsujan: Makes sense. Thanks.

tsujan (approved):

PR #1357, opened 2026-08-18
Drop woes

luebking:
Fixes the newWindow and closeWindow dbus methods for the dropdown mode.

In case of closeWindow the "fix" is to not do that but just hide the window (commented in code)
Changing this would require to redesign the dropmode code.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -371,8 +371,28 @@ QList<QDBusObjectPath> QTerminalApp::getWindows()
     return windows;
 }
 
+static QDBusObjectPath spawnNewProcess(const QString &dbus_id, const QString &shell_command, const QString& workdir, int columns, int lines)
+{
+    QStringList args;
+    args <<  QStringLiteral("-i") << dbus_id;
+    if (columns > 0 && lines > 0)
+        args <<  QStringLiteral("-s") << QStringLiteral("%1x%2").arg(columns).arg(lines);
+    args <<  QStringLiteral("-w") << workdir;
+    QString profile = Properties::Instance()->profile();
+    if (!profile.isEmpty())
+        args << QStringLiteral("-p") << profile;
+    args <<  QStringLiteral("-e") << shell_command;
+    QProcess::startDetached(QCoreApplication::applicationFilePath(), args);
+    return QDBusObjectPath();
+}
+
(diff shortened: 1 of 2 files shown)


stefonarch:
For the window it's ok, but closing last tab makes it still going to ghost.

luebking:
The process terminates for me?
(No matter whether I close the tab using dbus or the tab button)

tsujan:
I confirm @stefonarch's observation: closing the last tab of the dropdown terminal makes it unresponsive.

luebking:
Ah, I foolishly tried to do this from within the dropdown - it has to be hidden for this to work (you're losing the entire dbus interface)

tsujan:
When the dropdown terminal is locked and shown, it quits correctly on closing its last tab through dbus. The issue happens when it's hidden.

luebking:
Yes, see above.
This btw affects tristate-hiden windows as well.

The dbus interface is actually still intact - you could create a new window (that just needs to account for the dropMode)

So… votes on how to handle this:
1. Closing the last tab/window always kills the process?
2. Closing the last tab/window is ignored for hidden windows?
3. Closing the last tab/window closes the (hidden) window, but keeps the process - the user can create new windows
4. Special casing the dropdown: closing the last tab implicitly creates a new tab (preventing the window from being closed, preserves)

To me (1) is the most consistent, (2) the most buggy  (program doesn't do what I say), 3+4 maybe the most interesting one (though I don't yet know what to do with that - you could replace the only window snafu running foo with the only window snafu running bar :/)

stefonarch:
My vote: 1)

luebking:
I guess @tsujan  already voted 4 years ago, though some users beg to differ :grin: 
https://github.com/lxqt/qterminal/issues/154#issuecomment-1206407606

So (1) - tomo… later.

tsujan:
> I guess @tsujan already voted 4 years ago

Oh, sorry, I thought it was obvious. Yes, (1) is the correct approach.

stefonarch:
GTM, looks fine now closing window, tab or terminal with dropdown.

tsujan:
The last change makes perfect sense. Let's use it for a day or two before merging.

tsujan (approved):
GTM

Issue #480, opened 2018-09-13
Feature Request: Show notification when process completes


softwaresale:
##### Expected Behavior

This is a feature to display a desktop notification that notifies the user that a command
is completed when a terminal window is minimized. I'd be happy to work on this.

tim78245:
+1, this is the coolest feature of Konsole.

PR #1359, opened 2026-08-19
silence notification

luebking:
Got curious about https://github.com/lxqt/qterminal/issues/480
Preliminary patch, I guess it should be configurable.

```for ((i=0;i<10;++i)); do echo $i; sleep 1; done``` and hide the terminal
You should™ get a notification after ~20 total (10s for the script, 10 for the silence detection)
Activating the terminal either way (conveniently through the notification if your daemon supports actions) should™ implicitly hide the notification.

Do we want this at all?

Considerations:
1. config option (should be obvious but I spent way too much time on this anyway :smile: )
2. after showing the notification, monitor for activity and hide it again (not sure whether this can run into conflicts w/ the timeouts) in case some long term process stalls > 10s and then continues.

### DIFF
--- a/src/termwidget.cpp
+++ b/src/termwidget.cpp
@@ -26,6 +26,8 @@
 #include <cassert>
 
 #ifdef HAVE_QDBUS
+    #include <QTabWidget>
+    #include <QWindow>
     #include <QtDBus/QtDBus>
     #include "termwidgetholder.h"
     #include "terminaladaptor.h"
@@ -300,6 +302,10 @@ bool TermWidget::eventFilter(QObject * /*obj*/, QEvent * ev)
     {
         impl()->setFocus();
     }
+    else if (ev->type() == QEvent::KeyRelease)
+    {
+        m_activityClock.start();
+    }
     return false;
 }
 
@@ -309,6 +315,7 @@ TermWidget::TermWidget(TerminalConfig &cfg, const QString &dbus_id, QWidget *par
     , m_term(new TermWidgetImpl(cfg, this))
     , m_layout(new QVBoxLayout)
     , m_border(palette().color(QPalette::Window))
+    , m_silenceNotification(0)
 {
 
     #ifdef HAVE_QDBUS
@@ -347,11 +354,35 @@ void TermWidget::propertiesChanged()
     else
         m_layout->setContentsMargins(0, 0, 0, 0);
 
+    m_term->setMonitorActivity(Properties::Instance()->notifySilence);
+    if (Properties::Instance()->notifySilence)
+    {
+        connect(m_term, &QTermWidget::silence, this, &TermWidget::notifySilence, Qt::UniqueConnection);
+        // this is to limit the notification to events after prolonged activity and
+        // withdraw the notification if the terminal re-activates by itself
+        connect(m_term, &QTermWidget::activity, this, &TermWidget::notifyActiviy, Qt::UniqueConnection);
+    }
+    else
+    {
+        disconnect(m_term, &QTermWidget::silence, this, &TermWidget::notifySilence);
+        disconnect(m_term, &QTermWidget::activity, this, &TermWidget::notifyActiviy);
+    }
+
     m_term->propertiesChanged();
 }
 
+#define NOTIFICATION_INTERFACE QStringLiteral("org.freedesktop.Notifications"), QStringLiteral("/org/freedesktop/Notifications"), QStringLiteral("org.freedesktop.Notifications")
+
 void TermWidget::term_termGetFocus()
 {
+    if (m_silenceNotification)
+    {
+#ifdef HAVE_QDBUS
+        QDBusInterface notifications( NOTIFICATION_INTERFACE );
+        notifications.callWithArgumentList(QDBus::NoBlock, QStringLiteral("CloseNotification"), QList<QVariant>() << m_silenceNotification);
+#endif
+        m_silenceNotification = 0;
+    }
     m_border = palette().color(QPalette::Highlight);
     emit termGetFocus(this);
     update();
@@ -376,7 +407,101 @@ void TermWidget::paintEvent (QPaintEvent *)
     }
 }
 
-#if HAVE_QDBUS
+void TermWidget::notifySilence()
+{
+    if (m_activityClock.elapsed() < 45 * 1000)
+        return; // don't care about short term jobs
+    if (isActiveWindow())
+        return; //the user probably got that alright
+    bool visible = isVisible();
+    if (visible && window()->windowHandle())
+        visible = window()->windowHandle()->isExposed(); // for wayland and minimized windows
+
+    if (visible)
+    {
+        QApplication::alert(this);
+        return; // only notify for hidden windows
+    }
+
+#ifndef HAVE_QDBUS
+    // at least alert
+    QApplication::alert(this);
+#else
+    QList<QVariant> vl;
+    QVariantMap hints;
+    hints[QStringLiteral("transient")] = true; // don't log this notification
+    hints[QStringLiteral("urgency")] = 0; // low urgency
+
+    QImage screenshot(qMin(512,width()), qMin(256,height()), QImage::Format_RGB888);
+    QRect corner = screenshot.rect();
+    corner.moveBottomLeft(rect().bottomLeft());
+    render(&screenshot, QPoint(), corner);
+    QDBusArgument iiibiiay;
+    iiibiiay.beginStructure();
+    iiibiiay << screenshot.width();
+    iiibiiay << screenshot.height();
+    iiibiiay << screenshot.bytesPerLine();
+    iiibiiay << screenshot.hasAlphaChannel();
+    iiibiiay << 8; // bits per sample
+    iiibiiay << 3; // channels - RGB
+    iiibiiay << QByteArray(screenshot.constBits());
+    iiibiiay.endStructure();
+    hints[QStringLiteral("image-data")] = QVariant::fromValue(iiibiiay); // screenshot
+
+    QString body = window()->windowTitle();
+    if (TermWidgetHolder *holder = findParent<TermWidgetHolder>(this))
+        body = holder->label();
+    QStringList actions;
+    if (qApp->platformName() != QStringLiteral("wayland")) // wayland sucks™, https://github.com/lxqt/qterminal/pull/1359
+        actions << QStringLiteral("show") << QStringLiteral("Show Me!");
+    // vl <<  << summary << body << actions << hints << timeout;
+    vl  << QStringLiteral("QTerminal") << m_silenceNotification << QStringLiteral("qterminal") // appName << id << appIcon
+        << tr("Output ended") << body << actions << hints << 0 /*forever, until we close it*/;
+    QDBusInterface notifications( NOTIFICATION_INTERFACE );
+    QDBusReply<uint> reply = notifications.callWithArgumentList(QDBus::Block, QStringLiteral("Notify"), vl);
+    if (reply.isValid())
+    {
+        m_silenceNotification = reply.value();
+        QDBusConnection::sessionBus().connect(NOTIFICATION_INTERFACE, QStringLiteral("ActionInvoked"), this, SLOT(handleNotificationAction(uint, QString)));
+    }
+}
+
+void TermWidget::handleNotificationAction(uint id, QString action)
+{
+    if (id != m_silenceNotification)
+        return;
+    if (action == QStringLiteral("show"))
+    {
+        emit termGetFocus(this);
+        if (TermWidgetHolder *holder = findParent<TermWidgetHolder>(this))
+            if (QTabWidget *tabs = findParent<QTabWidget>(this))
+                tabs->setCurrentWidget(holder);
+        QWidget *win = window();
+        if (QWindow *handle = win->windowHandle())
+            handle->setWindowStates(handle->windowStates() & ~Qt::WindowMinimized);
+        win->show();
+        win->activateWindow();
+        win->raise();
+    }
+}
+
+#endif
+
+void TermWidget::notifyActiviy()
+{
+    m_activityClock.start();
+    if (m_silenceNotification)
+    {
+#ifdef HAVE_QDBUS
+        QDBusInterface notifications( NOTIFICATION_INTERFACE );
+        notifications.callWithArgumentList(QDBus::NoBlock, QStringLiteral("CloseNotification"), QList<QVariant>() << m_silenceNotification);
+#endif
+        m_silenceNotification = 0;
+    }
+    m_term->setMonitorSilence(true);
+}
+
+#ifdef HAVE_QDBUS
 
 QDBusObjectPath TermWidget::splitHorizontal(const QHash<QString,QVariant> &termArgs)
 {
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -855,13 +855,20 @@
                </widget>
               </item>
               <item row="12" column="0" colspan="2">
+               <widget class="QCheckBox" name="notifySilenceCheckBox">
+                <property name="text">
+                 <string>Notify when output stops in minimized windows</string>
+                </property>
+               </widget>
+              </item>
+              <item row="13" column="0" colspan="2">
                <widget class="QCheckBox" name="useCwdCheckBox">
                 <property name="text">
                  <string>Open new terminals in current working directory</string>
                 </property>
                </widget>
               </item>
-              <item row="13" column="0" colspan="2">
+              <item row="14" column="0" colspan="2">
                <widget class="QCheckBox" name="openNewTabRightToActiveTabCheckBox">
                 <property name="toolTip">
                  <string>If unchecked the new tab will be opened as the rightmost tab</string>
@@ -871,21 +878,21 @@
                 </property>
                </widget>
               </item>
-              <item row="14" column="0">
+              <item row="15" column="0">
                <widget class="QCheckBox" name="audibleBellCheckBox">
                 <property name="text">
                  <string>Audible bell</string>
                 </property>
                </widget>
               </item>
-              <item row="15" column="0">
+              <item row="16" column="0">
                <widget class="QLabel" name="label_14">
                 <property name="text">
                  <string>Default $TERM</string>
                 </property>
                </widget>
               </item>
-              <item row="15" column="1">
+              <item row="16" column="1">
                <widget class="QComboBox" name="termComboBox">
                 <property name="sizePolicy">
                  <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
@@ -911,10 +918,10 @@
                 </item>
                </widget>
               </item>
-              <item row="16" column="1">
+              <item row="17" column="1">
                <widget class="QLineEdit" name="handleHistoryLineEdit"/>
               </item>
-              <item row="16" column="0">
+              <item row="17" column="0">
                <widget class="QLabel" name="label_17">
                 <property name="toolTip">
                  <string>This command will be run with an argument containing the file name of a tempfile containing the scrollback history
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -152,6 +152,7 @@ void Properties::loadSettings()
     useCWD = m_settings->value(QLatin1String("UseCWD"), true).toBool();
     m_openNewTabRightToActiveTab = m_settings->value(QLatin1String("OpenNewTabRightToActiveTab"), false).toBool();
     audibleBell = m_settings->value(QLatin1String("AudibleBell"), false).toBool();
+    notifySilence = m_settings->value(QLatin1String("NotifySilence"), false).toBool();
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by stefonarch** [comment]
I use `notify-send blabla` in my scripts, but yes it would be nice. BUT: using tiling/scrolling WMs like niri there is no minimized state but only unfocused.

**Comment by stefonarch** [comment]
Ok, tested it under X11 and fine, under niri nothing happens (where no minimize exists) and labwc (where it does). Probably there is no way under Wayland despite its name to know if it's focused).

Maybe use urgency hints? Under niri on my setup it would activate itself, by default it gets a red border, kwin_wayland can display it in the taskbar/workspace buttons.

File: src/termwidget.cpp [hunk 1]
```diff
@@ -374,6 +390,55 @@ void TermWidget::paintEvent (QPaintEvent *)
 
 #if HAVE_QDBUS
 
+void TermWidget::notifySilence()
+{
+    bool visible = isVisible();
```
stefonarch, line 414: Checking also/or
`bool isWindowActive = (QGuiApplication::focusWindow() == win);`
works here on Wayland, both niri and labwc.

maybe it could also notify if not minimized but on another workspace or covered (=inactive) only? Anyway, I think it's a real useful feature to have. It works fine for compilations. Could maybe include the cmd used in the notification?

stefonarch:
There is a bug: the notification still pops up repeatedly after 10s when another cmd is executed, e.g. `ls` in the same terminal and it's unfocused after. Not sure if this is due to my hack though.

stefonarch:
There is a bug: the notification still pops up repeatedly after 10s when another cmd is executed, e.g. `ls` in the same terminal and it's unfocused after or also when returning from suspend. Not sure if this is due to my hack though but I don't think so.

luebking:
> Not sure if this is due to my hack though.

Probably - the notifications are not supposed to show for visible windows, otherwise you get silence signals all the time (output -> no output for 10 seconds -> fire)

I'll still have to check whether monitoring for activity to auto-hide the notification might be a good idea,

I'll try something else for wayland… (which let's remember: wayland a blessing for all of us. You cannot control where windows show up nor whether they're visible and you need three protocol extensions to control the window layer and decoraction)
I assume there's not gonna be a surface anymore when the window is hidden in any capacity (incl. when minimized) - niri would become an unrelated problem.

Speaking of which: the dropdown mode should work w/ the original patch because it's internally hidden.

About the practical sense: I also fire notifications from scripts all the time - as you mentioned it's more geared towards the casual user/long running job.
Afaict the terminal only tells you that the output stopped, but it might be possible to leverage the history buffer (last line there…)

luebking:
Ok, the updated patch should™ work on wayland (including niri if /it/ does it right) - windows get to know whether they're "exposed"
It will also withdraw notifications when the output reactivates by itself.

About the history: meh.
QTermWidget allows to save it but no direct access, this this would require to **copy** the entire scrollback buffer into a https://doc.qt.io/qt-6/qbuffer.html to get access to its contents.
Doable but not sure how @tsujan  feels about that.
It would however allow the notification to show where the output (compiler job…) ended.

stefonarch:
Quick test:
* labwc and wayfire minimized: notification is shown, raising on it's button doesn't work anymore (I _think_ it did before). Executing `ls` and minimize→triggers again the notification.
* niri: Nothing happens.Changing `isExposed` to `isActive()`  it works again on niri, including switch/focus to the window from the notification button. But also here any other command afterwards restarts the notification timer. 
Niri does everything right as it's screens aren't confined to the monitor edges but infinite, as long as there are windows.
* Hyprland, switching to another workspace: `isExposed` works, click notification to focus too.

tsujan:
If I were you, I wouldn't waste my time by trying to activate, minimize, un-minimize, check the exposed/activated state, … under Wayland. Some tricks may work with some compositors, but that happens only by chance.

luebking:
@stefonarch  I didn't change the activation code, from a  quick test I ran earlier I seem to recall that the window got shown at the bottom of the stack.
The "ls + immediate minimize"  will fire because ls causes output, 10s later you get a silence signal and then the window is minimized.
Only listening when the terminal gets hidden (or the window unexposed) will trigger the bogus initial signal, so the trade-off would be skip that and only fire for jobs that run at least 25 seconds.

@tsujan 
Self-activation is also frequently inhibited by focus protection rules in X11 WMs.
As much as wayland annoys me, it seems possible to check whether the window is exposed (to the user) at least on WLR-adjacent compositors and gnome is probably no target audience.
Either way, the implication is that there's no appetite for the patch?

About niri: not sure that it does everything right if the exposure state doesn't depend on whether the window is currently visible to the user since that state is meant for clients to optimize drawing (ie. stop expensive actions the user won't see anyway)
But the concept doesn't seem to be part of https://wayland.freedesktop.org/docs/html/apa.html so might just be WLR specific (I'd have to look up the Qt code to figure how this is determined)

tsujan:
> As much as wayland annoys me

It annoyed me before, but now I'm reluctant to log into an X11 session ;) 

Public Qt methods aren't enough for dealing with Wayland's peculiarities consistently. Hopefully that will change in some future version of Qt, but until then, adding Wayland codes by using Qt's private headers is justified only in a few apps (like lxqt-panel). So, we have to accept that features related to positioning, activating, (un-) minimizing, etc. don't work on Wayland.

stefonarch:
I 'm quite sure niri does it right (it does really right everything usually) because on "normal" tiling WMs a window is or always visible but tiled (if not minimized on X11) or not (because on another workspace), while a screen on niri can look like this:
immagine. So all windows on all workspaces are exposed, the screen cuts the view only.

>The "ls + immediate minimize" will fire because ls causes output.

Oh, it looks like that also in a new terminal the message is shown. So to be usable it should act only on longer running cmd, otherwise it's useless as you may switch a moment to your terminal to get some info and switch away after. Ofc minimized is different then switching away to another workspace.

tsujan:
> it does really right everything usually

I agree. Niri and Labwc can be trusted most of the time. That's not the case with other Wayland compositors.

luebking:
>  the screen cuts the view only
And does that result in windows being completely off screen so you can't see them but they still show up as exposed?

> can be trusted most of the time
I don't trust any of them :P
Forcing myself to use one, wtf is my mmb paste?
And why can't I have scrollwheel emulation AND scroll wheel run in the correct direction (one's "natural" and the other isn't)
Toggling multiple outputs on and off also works less than great 

Either way
> to be usable it should act only on longer running cmd
Pretty much that - and it might still not cut it.
After more synthetic tests, running some actual terminals with that quickly became very annoying because tmux (clock), mutt (status updates), ncmpcpp (tells you  that playback has now finished), fzf (throbber) this becomes very annoying very quickly.

I'll keep playing with it and keep you posted if there's principal interest in the feature (provided we can make it behave)

luebking:
Set the minimum job time to ~30-35 seconds (~10s go into the actual silence detection) - this should be much calmer.
Window activation only works on wayland (for me) if the window is hidden, but not if it's minimized or at the bottom of the stack, but oh well…

Edit:
This is still not gonna be flawless - occurred to me (and tested) that you could violently use the terminal for a minute (frequently run ls) and then hide it.
Because the terminal cannot/does not distinguish what processes cause the activity, it will figure that you ran a long term job that's no longer producing any output - maybe we can intercept input to reset the activity clock…

Edit #2:
Yes, we can.

tsujan:
> I don't trust any of them :P Forcing myself to use one, wtf is my mmb paste? …

@luebking, it seems that you hate Wayland ;)  Don't take me wrong; we need a Wayland hater among us to focus on X11. But in case you want to test something on Wayland without hassles, you could try an LXQt Wayland session by using LXQt Session Settings → Wayland Settings with labwc.

luebking:
I'll usually test wayland in a nested labwc session.

It's not hate and I'd happily switch to a leaner display server, but I also expect some basic things to work.
I need to check whether https://github.com/labwc/labwc/pull/2832 is just a dumb default or there's some bigger bug and I want the wheel emulation and actual wheel to operate differently (natural and classic) … I'll spare you the rest of the list of annoyances.

As for activation, I intend to "fix" qiq by ironically flickering it into the active state, since activation of new windows works, just not of present ones - what some genius probably deemed mOrEsEcUrE…
That might also be a vector for the qterminal activation pattern that currently fails on wayland.

Back to the regularly scheduled worries: do we want this patch in qterminal?
(assuming that the latest iteration provides sane behavior)
Should I go for adding a config option (and maybe a blacklist for client processes where this would at best produce noise, like ncmpcpp)?

tsujan:
> do we want this patch in qterminal?

If no one wanted it, https://github.com/lxqt/qterminal/issues/480 would be closed. On the other hand, it doesn't seem very popular.

I'm impartial and will accept any decision that might be made by others. @stefonarch, @yan12125?

luebking:
Ftr
```
diff --git a/src/termwidget.cpp b/src/termwidget.cpp
index b5b299d..b62946b 100644
--- a/src/termwidget.cpp
+++ b/src/termwidget.cpp
@@ -483,9 +483,11 @@ void TermWidget::handleNotificationAction(uint id, QString action)
             if (QTabWidget *tabs = findParent<QTabWidget>(this))
                 tabs->setCurrentWidget(holder);
         QWidget *win = window();
-        win->show();
         if (QWindow *handle = win->windowHandle())
             handle->setWindowStates(handle->windowStates() & ~Qt::WindowMinimized);
+        if (win->isVisible() && qApp->platformName() == QStringLiteral("wayland"))
+            win->hide(); // this is ridiculous…
+        win->show();
         win->activateWindow();
         win->raise();
     }
```
This actually works…

tsujan:
> This actually works…

You couldn't see many details by testing Wayland only in a nested session of one compositor. Hiding and showing a window that doesn't have a fixed position are a very bad way of activating it.

The most you could do is calling `QApplication::alert`. Some compositors see it, some don't.

luebking:
Was actually a bare metal session.
I'm aware that it's impossible to generalize anything with "wayland" , https://doc.qt.io/qt-6/qapplication.html#alert doesn't even try to address that platform. Could still call it along the notification.

But it's an interesting enough result an d if it works in the wlr sphere (@stefonarch  may want to try niri)  it's maybe "good enough"?

tsujan:
> maybe "good enough"?

Far from that. First, the position of the window might change. Second, the window might not be shown at all. Third, it might not be activated.

Years of dealing with Wayland have shown me that each Wayland compositor may behave differently. Even testing under full sessions of LXQt + labwc/kwin_wayland/wayfire/niri (the compositors I have here) isn't enough.

In short: activating and setting the position of a normal window are among taboos under Wayland. Layer windows — like those of lxqt-panel, LXQt notification and drop-down QTerminal — have a different story.

EDIT: Of course, I mean it inside the app and without using *direct* Wayland codes plus Qt's private headers. The latter is a taboo in LXQt ;) We don't use *direct* X11/Wayland codes. There are exceptions, but QTerminal isn't one of them.

stefonarch:
@luebking my way of testing is: tty2, a clean or minimal configured user (terminal, session settings on the desktop or panel), switch to it and run `startlxqtwayland`. Using /tmp to pass files or screenshots. Inside my home I've too many hacks I forgot about ;)

Will test the patch later.

tsujan:
> Will test the patch later.

Although I'm impartial toward adding this feature (it could be done by `notify-send` after the command too), I'm not so regarding codes that may work randomly on this or that Wayland compositor ;)

yan12125:
> do we want this patch in qterminal?

It's okay to have this. `notify-send` or special escape sequences (`ESC ] ... BEL`, OSC 777, OSC 99, ...) can provide notifications when commands are completed, and this PR seems to also provide notifications when commands get stuck.

Regarding Wayland issues, I don't have a good idea, as I never use Wayland as daily driver, and during the last time I study Wayland (years ago), it's too messy to understand.

luebking:
@tsujan, sounds like somebody's hating wayland :stuck_out_tongue: 
(No offense, it /is/ very frustrating to develop for that platform)

But what I meant is, if the window activation works on labwc, niri, sway, river, hyprland, KDE… how many gnome users use qterminal?
Do they justify denying everybody else a feature just because (assuming) gnome won't play ball?

That being said: it'd be no big deal to whitelist the "show me" button for known-to-work environments.

@stefonarch, make sure to apply the patch I posted on top (didn't update the PR yet)

@yan12125, I think the original scenario is actually for non-premeditated runs or casual users, not for frequent tasks.
But yes, it will (ideally) detect if something produced output for > 30s (w/o user input) and then stalls for 10s
If the task resumes by itself, the notification will be withdrawn.
It will also be automatically withdrawn when the terminal (window) gets focused and the wayland discussion is about a button in the notification (iff the notification daemon supports that itfp) that allows to conveniently activate the terminal from there.

---
If everyone who gets to vote is ok with moving forward w/ the principle feature (details tbd) I'll add a config option (assuming that's implied)


One aspect not addressed is @stefonarch's request to have the notification show the tail of the history (which I guess is nice because you can see whether the compilation failed w/o having to open the terminal) and whether that justifies copying the entire history into a QBuffer (what will be required w/ the current API) to get to the tail.

stefonarch:
I've no idea what happened but I can't get any notification anymore now, neither on niri with `isActive()` nor on labwc or x11. Just running `sleep 60` as before. Latest Qt update broke the desktop, maybe it did some more damage or I made some stupid mistake but don't think so.

luebking:
```for ((i=0;i<60;++i)); do echo $i; sleep 1; done``` and minimize the window.
Just running "sleep 60" won't work because it will only produce a single update that doesn't match the "this is a long running job and not some random update or the user typing" pattern.

stefonarch:
I see, forgot about this change.
 The patch is working/raising/focusing the window if minimized or on the other monitor with kwin_wayland and labwc, and probably also with wayfire. But still with `isActive()` for niri ;)

luebking:
Does it actually activate the window on niri (regardless of the problems w/ the exposure status on that shell)?

https://codebrowser.dev/qt6/qtbase/src/plugins/platforms/wayland/qwaylandwindow.cpp.html#1605
https://codebrowser.dev/qt6/qtbase/src/plugins/platforms/wayland/qwaylandshellsurface_p.h.html#_ZNK15QtWaylandClient20QWaylandShellSurface9isExposedEv

::isExposed() actually just tests whether a wayland surface exists, ie. niri doesn't destroy those even if a window is out of sight what sounds like a massive performance bug (using CPU and GPU time and RAM to render a surface that is the fully clipped out - you could probably not even run E1M1 on modern hardware if doom was implemented that way…)

stefonarch:
On niri (and I think also Hyprland and niri) it always activated the window also before.
I'm quite confident in the niri dev, for the rest, if he does talks like [this](https://mastodon.social/@YaLTeR@mastodon.online/117054437416998470). Btw I didn't test if it's  raising if below the stack on labwc.

Maybe showing the cmd could be an option, if this is a performance issue.

tsujan:
> sounds like somebody's hating wayland

Oh, Wayland has tortured me on the coding level. But I like its performance, especially when I have 3 reliable Wayland sessions.

It's just that I dislike dirty workarounds. Some Wayland compositors (like kwin and labwc) have smart and random window positioning, and even if there's no other issue (my gut feeling tells me there will be), making a window disappear and reappear somewhere else is bad.

Please ignore activation on Wayland.

stefonarch:
>making a window disappear and reappear somewhere else is bad.

That's impossible, just disappear and reappear can be done, but not somewhere else  afaik, the compositor will place it where it was on stacking  ones.

I tested Hyprland with `isExposed()`, it will not trigger the notifications if it's visible, even on the "inactive" monitor, but only when it's on a invisible workspace.

luebking:
I wanted to try something else for niri first (got curious what @stefonarch  is simping so hard about here :P) but
https://codebrowser.dev/qt6/qtbase/src/plugins/platforms/wayland/qwaylandwindow.cpp.html#650
hide() only detaches the buffer from the wl_surface, it doesn't destroy it.
So while I'm certainly not gonna die on that hill I wonder whether there's at least anecdotal evidence of a compositor taking that as opportunity to randomly warp around a window… or your gut feeling is only driven by your irrational hate at wayland ;)

Off topic on niri performance, there might be ulterior performance motives, https://github.com/niri-wm/niri/issues/1150 and the surface withdraw might contradict the entire concept, https://github.com/niri-wm/niri/issues/539 (interestingly the low refresh rate itself could be leveraged to detect the condition - but that would be for Qt to do) but keeping all surfaces around all the time will be to the detriment of your RAM - which thanks to the clankers is no longer anywhere near free :)

tsujan:
> That's impossible

It's quite possible. There's more than one compositor and one configuration for each.

And I've already experimented with ways beyond a simple hiding/showing in the case of FeatherPad. I couldn't put them into a comment even if I was motivated or had time. One of them worked with all compositors I had except for Niri (by using a tiny and totally transparent "dummy" widget on the top layer), but it wasn't Niri's fault.

luebking:
niri actually moves the window to the right of the active one for at least a prolonged activateOrHide cycle (away from a tile), so probably dumps the surface at some point after the buffer is detached (I've not tried the flickering, ie short-term hide/show - the other way to figure whether we're currently on-screen failed as well :( )

Of course the approach would also not be necessary on niri since it apparently allows direct activation… but the feature doesn't work there itfp. because we cannot figure whether the window is visible to the user.

Wayland Sucks™

tsujan:
> Wayland Sucks™

If you're a coder who has to deal with several Wayland compositors by using Qt, it's a nightmare. That's 100% guaranteed. But if you're a user who has found his favorite compositor(s) with the best setup, it's very nice :P

stefonarch:
Just tested what happens if it's below another window on labwc: won't go at the top...
Had a chat with the niri dev, it looks like other compositor use "suspended state" which Qt checks for among other with `isExposed()`.

luebking:
> if it's below another window on labwc
Does that even fire the notification then?

I assume one can forget about *only* raising the window because the relevant bit is likely that https://codebrowser.dev/qt6/qtbase/src/plugins/platforms/wayland/qwaylandwindow.cpp.html#644 Qt asks to be activated when showing.
Or is the window on a layer lower than the window covering it?
You're never getting some random window over a fullscreen one, not even on (most) X11 WMs

Another thing (also because of the scrollback buffer) - we can technically send a (partial/scaled) screenshot with the notification.
Whether and how the notification daemon will show it is a different matter.
(Though, caveat, this would likely suck for very wide windows except if we're just cutting out the lower left cornder w/ a static size - 512x256 or so)

luebking:
Lastest push 
* adds a config option (disabled by default)
* deactivates the show action on wayland and
*  posts a screenshot of the lower left corner - nb. not every notification daemon will show that.

luebking:
Latest push:
- alerts visible windows (doesn't do anything on either labwc or niri, I assume one needs a taskbar for that)
- enables the feature w/p dbus (alerting exclusively)

labwc btw. gets bonus points because it apparently suspends completely occluded windows (you could check that w/ Xlib but QWindow::isExposed() doesn't, ie. this won't work on X11 but on labwc the window can be "visible" behind a video and you still get a notification)

tsujan:
> doesn't do anything on either labwc or niri, I assume one needs a taskbar for that

They both rely on lxqt-panel's wlroots backend (although Niri isn't based on wlroots), which doesn't support alerting yet. The kwin_wayland backend does — it has the most complete support, which is also the most Qt-friendly one. After it comes Wayfire's backend, but I don't remember if it supports alerting — maybe not.

luebking:
Latest compare looks horrific but is actually a simplification of the propertyform.ui patch (I noticed that the label for the default $TERM was gone, probably to the last merge horror, and decided to just manually patch the form to avoid qt-designers antics)

tsujan:
@luebking 
Ready to merge?

luebking:
From here, yes.
@stefonarch  might want to comment on the approach of posting screenshots of the TE fragment (and whether that's useful for him at all)

tsujan:
It's mostly for X11, which most people use. Even with `QApplication::alert()`, it won't do anything under most Wayland compositors (except for kwin_wayland perhaps).

tsujan:
Let's go to a testing state (or as it's called in Debian, a "freeze" state) for a month, before releasing LXQt 2.5.0 in November.

luebking:
Errr… the notification will be sent when the window is not exposed what worked exceptionally well on labwc (which at least understands that a window is not exposed when covered by a fullscreen window - something Qt on X11 does not)

tsujan:
Yes, it works on Labwc with a minimized window and also when the window is on another workspace. You have hidden love for Wayland ;)

Whether or how it may work on other Wayland compositors is out of our control.

luebking:
> Let's go to a testing state (or as it's called in Debian, a "freeze" state) for a month, before releasing LXQt 2.5.0 in November.

https://github.com/lxqt/qtermwidget/pull/677 (if you want to break ABI I also have a patch to set the size, the alternative would be to introduce a signal that the implementation can hook up with)
https://github.com/lxqt/qtermwidget/pull/681

tsujan:
> if you want to break ABI

That's never a problem in itself. ABI is there to be broken when really needed.

tsujan:
Please tell it in the main comment of any PR that may break ABI (not inside the code). That's just for the git users to recompile qterminal afterward.

Issue #1356, opened 2026-08-18
Visual bell, please


luebking:
##### Relevant Problems
Nobody wants their terminal to make funny noises, but there should™ be a response to \a
This is also an a11y problem, because there are people who cannot hear the funny noises their computer makes.

##### The Solution You'd Like
The feature is currently stashed in TerminalDisplay and the mode unconditionally set when creating QTermWidget
There two ways to address this
1. expose the TerminalDisplay setting via the QTermWidget API
2. https://doc.qt.io/qt-6/qgraphicseffect.html - which doesn't require qtermwidget patches and allows all sorts of fancy stuff (animated inversions, flashes, slaps, blurring, fish-eye… whatever you can come up with) but is kinda monkey-patching the feature on top.

This is basically an inquiry for the preferable approach.

tsujan:
> Nobody wants their terminal to make funny noises

I like the current sound  :P 

> and allows all sorts of fancy stuff

In general, LXQt avoids fancy stuff in its codes. As for animations, there's only an optional one in lxqt-panel, because some users might find it unnatural or even annoying if their auto-hiding panel was shown or hidden suddenly.

luebking:
I'm perfectly fine w/ a 150ms inversion flash - I just brought that up to point out the pros and cons and options.
Do you have a preference on how to enable the visual bell?

tsujan:
I just told what I knew about LXQt; otherwise I like an optional animation :)

> I'm perfectly fine w/ a 150ms inversion flash

If it doesn't clutter the code, make a PR please.

luebking:
I meant about either exposing the API in qtermwidget and using that or ignoring the terminaldisplay feature and implement a home-brew effect leveraging QGraphicsEffect

Not so much the style of preferred flashiness ;)

PR #1361, opened 2026-08-19
implement visual bell

luebking:
https://github.com/lxqt/qterminal/issues/1356

So here's a POC for the approach that does not require an (aligned) API update of qtermwidget.
Since the graphicseffect approach also means I get to do whatever I want, I want a bloodshot (like in FP shooters :))

The final implementation would probably use the red of the color scheme and also be configurable (if one prefers only an audible or no bell), for now this is RFC on the topic.

### DIFF
--- a/src/forms/propertiesdialog.ui
+++ b/src/forms/propertiesdialog.ui
@@ -872,20 +872,27 @@
                </widget>
               </item>
               <item row="14" column="0">
+               <widget class="QCheckBox" name="visualBellCheckBox">
+                <property name="text">
+                 <string>Visual bell</string>
+                </property>
+               </widget>
+              </item>
+              <item row="15" column="0">
                <widget class="QCheckBox" name="audibleBellCheckBox">
                 <property name="text">
                  <string>Audible bell</string>
                 </property>
                </widget>
               </item>
-              <item row="15" column="0">
+              <item row="16" column="0">
                <widget class="QLabel" name="label_14">
                 <property name="text">
                  <string>Default $TERM</string>
                 </property>
                </widget>
               </item>
-              <item row="15" column="1">
+              <item row="16" column="1">
                <widget class="QComboBox" name="termComboBox">
                 <property name="sizePolicy">
                  <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
@@ -911,10 +918,10 @@
                 </item>
                </widget>
               </item>
-              <item row="16" column="1">
+              <item row="17" column="1">
                <widget class="QLineEdit" name="handleHistoryLineEdit"/>
               </item>
-              <item row="16" column="0">
+              <item row="17" column="0">
                <widget class="QLabel" name="label_17">
                 <property name="toolTip">
                  <string>This command will be run with an argument containing the file name of a tempfile containing the scrollback history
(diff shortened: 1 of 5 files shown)


luebking:
Withdrawing the draft to get some comments (not gonna add a config as long as the silence notification is pending or this one gets a clear go! ;))

1. Do we add a visual bell at all?
2. Is it preferable to implement it this way (in the client) or to extend the qtermwidget API to allow enabling the visual bell?
(though iirc r/n that's mutually exclusive w/ the notification, ie. the user can have either a visual or an audible bell, not both like with the presented approach)
3. What's it gonna look like?
The bloodshot is actually just semi-a-joke and ret-conned; I arrived at it because:
   * we are at liberty to implement the VB anyway we want
   * the inversion at night flashes a bright light into a dark room, so maybe something more muted
   * vignette/inverted vignette are too subtle/aggressively bright
   * what's a muted but alerting color that's not too bright and probably no sane beings terminal background?
   * obviously red!
   * at this point I realized that I had just re-invented the bloodshot and figured I'm just gonna own that.
   * in a final version I'd probably explore the themes red, rather than hardcoding one (mine)
I would however also understand objections to both the effect as well the class name.

tsujan:
My answers are my opinions; they aren't meant to be a guide.

 1. If possible with a minimal code, why not?
 2. In this way (as implicit in 1).
 3. No opinion yet. I'll tell you later.

That being said, if it's completed, I think we should merge it only *after* the next release, in November.

tsujan:
I tried your effect now and liked it :)  Minimal and effective. As far as I'm concerned, it only needs to be made optional.

More tests/opinions are needed. @stefonarch, @ahsand97, … ?

stefonarch:
As I never used the bell I've no idea when it will be triggered, but yes, it's nice to have IMO, so GTM


https://github.com/user-attachments/assets/b277cfb0-d1d1-4946-9dc2-b3d5357c4938

tsujan:
> As I never used the bell

The audible bell has been useful to me since it was implemented, but this visual bell is even better.

tsujan:
Contrary to what I said earlier. now I think we could merge it when it's made optional and added to the settings dialog.

IMO, there's no need to fancy effects; this is already good and practical  (my answer to 3).

ahsand97:
> but this visual bell is even better.

I agree, tested and works nice.

luebking:
Well well well… seems really *nobody* likes the funny noises their terminal makes after all ;)

@stefonarch
It's kind of an antique "dingg" and used by various clients to indicate a soft-error (eg. by "less" or vim whenever you reach either end and try to go on - means "you've reached the end of the document") to get the users attention.
Historically the idea was probably to make terminal A make a sound after finishing a job while you were 10m away at terminal B to fix the tape device - today the bell is not super-important but you get kinda used to it happening.

Also people quickly quickly figured that the noise is just annoying when you're actually in front of the system :)

Chiitoo:
> As I never used the bell I've no idea when it will be triggered, but yes, it's nice to have IMO, so GTM

On Konsole, for example, it seems to be triggered by, for example, hitting backspace when there is nothing to back space.

Have not tested this change yet though.

It can also "monitor for activity" and maybe alarm you for that.

I think Ubuntu not so long ago still had an audible bell default on.

I saw that at work, that I have not had in some years now, and it was kind of funny when someone wrote a guide for setting up the Linux environment, and turning off said bell was one of the steps.  :]

tsujan:
> On Konsole, for example, it seems to be triggered by, for example, hitting backspace when there is nothing to back space.

The same in QTerminal. Also hitting left/right arrow keys when the text cursor is at the left/right-most position.

luebking:
Technically it's neither konsole nor qterminal but bash that's doing that :)

tsujan:
Giving it to git users to enjoy and test it …

PR #1362, opened 2026-08-20
add feature to maximize the current tile

### LINKED ISSUE #1041: [Feature request] "maximise" current sub-term
**Is your feature request related to a problem? Please describe.**
It's possible to split qterm screen into smaller sub-terminals, but sometimes, one wants to have a sub-terminal bigger temporarily, hiding other sub-terminals during this time.

**Describe the solution you'd like**
An action (with shortcut) to toggle "big" state of a sub-terminal that hides other sub-terminals in the same tab.
This feature exists for exemple in terminator, another screen splitting terminal app.

##### Context

I use terminator frequently with sub-terminals like in qterminal, but sometimes I want to have a subterminal taking the whole window, temporarily to do some stuff like an ncurses app or viewing a lot of data. There's a shorcut in terminator to toggle that. Then when I quit the task in the subterminal, I press the shortcut again to toggle back the splitting state and see other subterminals as before.
This doesn't exist in qterminal and prevents me from using sub-terminals.

luebking:
fixes #1041

I guess this makes sense since users will see tiling as nested… tiling window manager

This turned out to have an interesting obstacle - there's extensive explanation in the code comments that probably need to stay for future readers.

### DIFF
--- a/src/termwidgetholder.cpp
+++ b/src/termwidgetholder.cpp
@@ -266,6 +266,72 @@ void TermWidgetHolder::propertiesChanged()
         w->propertiesChanged();
 }
 
+class EventKiller : public QObject
+{
+    protected:
+        bool eventFilter(QObject *, QEvent *) { return true; }
+};
+
+void TermWidgetHolder::maximize(TermWidget *term)
+{
+    QList<TermWidget*> all = findChildren<TermWidget*>();
+    const bool reset = !all.contains(term); // nullptr is ok, "maximize nothing"
+
+    // We need to pull a little magic trick here and we do that with a
+    static EventKiller eventKiller;
+    // Hiding the TermWidget will make the splitter collapse it, meaning resize it 0x0
+    // It will be properly resized with the show event BUT the Konsole::TerminalDisplay
+    // does not respond well to that - it's update optimizations somehow™ respond to the
+    // shrinking but not the growing event, ie. it will look mostly empty until the
+    // content updates, in doubt because of a focus event
+    // It therefore must not notice any of this…
+
+    QWidgetList hibernates;
+    for (TermWidget *other : all)
+    {
+        bool willBeVisible = reset || other == term;
+        if (willBeVisible == other->isVisibleTo(this))
+            continue;
+
+        // … find the "Konsole::TerminalDisplay" - one more time and TermWidget needs a function for that ;)
+        QWidget *terminalDisplay = nullptr;
+        for (QWidget *kid : other->findChildren<QWidget*>())
+        {
+            if (kid->inherits("Konsole::TerminalDisplay"))
+            {
+                terminalDisplay = kid;
+                break;
+            }
+        }
+        // If we're hiding this, we put it in hibernation first
+        if (terminalDisplay && !willBeVisible)
+            terminalDisplay->installEventFilter(&eventKiller);
+        // Without the need for magic, this is the actual business
+        other->setVisible(willBeVisible);
+        // If we're showing this, enlist it to wake up from hibernation - LATER!
+        if (terminalDisplay && willBeVisible)
+            hibernates << terminalDisplay;
+    }
+    // After the layout has been updated, immediately refresh the splitters, ie. resize all widgets
+    for (QSplitter *splitter : findChildren<QSplitter*>())
+        splitter->refresh();
+    // …and now we wake up the displays out of hibernation.
+    for (QWidget *terminalDisplay : hibernates)
+        terminalDisplay->removeEventFilter(&eventKiller);
+    // TADAA :)
+}
+
+void TermWidgetHolder::toggleMaximized(TermWidget *term)
+{
+    assert(term);
+    if (!term->isVisibleTo(this))
+        return maximize(term);
+    for (TermWidget *other : findChildren<TermWidget*>())
+        if (other != term && other->isVisibleTo(this))
+            return maximize(term); // we're currently not maximized
+    maximize(nullptr); // we're maximized, back to regular tiles
+}
+
 void TermWidgetHolder::splitHorizontal(TermWidget * term)
 {
     TerminalConfig defaultConfig;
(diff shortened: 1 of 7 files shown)


src/termwidgetholder.cpp
```diff
@@ -266,6 +266,72 @@ void TermWidgetHolder::propertiesChanged()
...
+void TermWidgetHolder::maximize(TermWidget *term)
+{
+    QList<TermWidget*> all = findChildren<TermWidget*>();
+    const bool reset = !all.contains(term); // nullptr is ok, "maximize nothing"
+
+    // We need to pull a little magic trick here and we do that with a
+    static EventKiller eventKiller;
+    // Hiding the TermWidget will make the splitter collapse it, meaning resize it 0x0
+    // It will be properly resized with the show event BUT the Konsole::TerminalDisplay
+    // does not respond well to that - it's update optimizations somehow™ respond to the
```
luebking, line 284: "its", I do speak English… somewhat :)

tsujan:
Bug:

If a maximized split view is closed (e.g., by `exit`), there's no way to see the other view(s). And if the toggle-max action of the menu-bar is used in this case, a crash will happen:

```
#0  0x00007f356fd5b4ce in QWidget::isVisibleTo(QWidget const*) const () at /usr/lib/libQt6Widgets.so.6
#1  0x0000559004451935 in TermWidgetHolder::toggleMaximized(TermWidget*) ()
#2  0x00007f356ebef236 in ??? () at /usr/lib/libQt6Core.so.6
#3  0x00007f356f85522e in QAction::activate(QAction::ActionEvent) () at /usr/lib/libQt6Gui.so.6
#4  0x00007f356ff2b0d0 in ??? () at /usr/lib/libQt6Widgets.so.6
#5  0x00007f356ff2fbbf in ??? () at /usr/lib/libQt6Widgets.so.6
#6  0x00007f356fd6407e in QWidget::event(QEvent*) () at /usr/lib/libQt6Widgets.so.6
#7  0x00007f356fd0325f in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#8  0x00007f356fd04c2c in QApplication::notify(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#9  0x00007f356eb822b8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#10 0x00007f356fcfae06 in QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) () at /usr/lib/libQt6Widgets.so.6                                                                                                              
#11 0x00007f356fd80a60 in ??? () at /usr/lib/libQt6Widgets.so.6
#12 0x00007f356fd814bb in ??? () at /usr/lib/libQt6Widgets.so.6
#13 0x00007f356fd0325f in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#14 0x00007f356eb822b8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#15 0x00007f356f3b70c0 in QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) () at /usr/lib/libQt6Gui.so.6
#16 0x00007f356f430d3b in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt6Gui.so.6
#17 0x00007f356f97be74 in ??? () at /usr/lib/libQt6Gui.so.6
#18 0x00007f356e0fec76 in ??? () at /usr/lib/libglib-2.0.so.0
#19 0x00007f356e100ed7 in ??? () at /usr/lib/libglib-2.0.so.0
#20 0x00007f356e101055 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#21 0x00007f356ee96f32 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt6Core.so.6
#22 0x00007f356eb8f664 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt6Core.so.6
#23 0x00007f356eb8685a in QCoreApplication::exec() () at /usr/lib/libQt6Core.so.6
#24 0x000055900442ce7e in main ()
```

src/termwidgetholder.cpp
```diff
@@ -399,6 +465,8 @@ void TermWidgetHolder::handle_finished()
     {
         qFatal("TermWidgetHolder::handle_finished: Unknown object to handle");
     }
+    if (w->isVisibleTo(this))
```
luebking: The compare is abhorrent for the interim commits, this fixes behavior (and implicitly crash) by unconditionally unmaximizing everything if the terminal to be removed is visible to the holder (the call is in doubt inert)

tsujan: I don't think that's the solution. It's still prone to crash, because the user can end up with a view-less window by clicking "Close Subterminal", in which case a crash is imminent.

IMO, a view-less window should be made impossible, as it is with git. Apart from being a potential cause of crashes, it's confusing to the user. (I vaguely remember I made it impossible years ago, but don't count on my memory).

EDIT: https://github.com/lxqt/qterminal/commit/8b1385e4299ca259c5759b6641cd435e0ee93f1a

src/termwidgetholder.cpp
```diff
@@ -280,6 +346,8 @@ void TermWidgetHolder::splitVertical(TermWidget * term)
 
 void TermWidgetHolder::splitCollapse(TermWidget * term)
 {
+    if (term->isVisibleTo(this))
```
luebking, line 349: Ah, that's called directly.
So let's just move the bail there.

tsujan: Yes, this should cover all cases. Will test when the PR is ready.

There are still 2 small UX issues.

 1. The action should be disabled with a single view (like the action "Close Subterminal").
 2. A text like "Toggle Subterminal Maximized" (capitalized) is more consistent with "Close Subterminal" (below it). We need to take care of these things before any translation is added.

luebking: There's even a nice infrastructure - and I got to fix my english :)

tsujan (approved):

Issue #899, opened 2022-02-02
Greeting text from .bashrc is overwritten


z0rti:
##### Expected Behavior
Text should show completely

##### Current Behavior
Text is overwritten by prompt

##### Steps to Reproduce (for bugs)
put this in .bashrc
```
echo "--- Greetings, $USER"
echo "--- Your terminal size is: ${COLUMNS}x${LINES}"
```
(nothing else)
start qterminal, then run the echo command again to see the difference
```
--- Greetings, root
root@indri:~# nal size is: 12x2
root@indri:~# echo "--- Your terminal size is: ${COLUMNS}x${LINES}"
--- Your terminal size is: 80x27                                                
root@indri:~#                                                                   
```

It works in gnome-terminal:
```
--- Greetings, root
--- Your terminal size is: 80x24
root@indri:~# 

```

##### System Information
* QTerminal version: master
* Distribution & Version: Ubuntu indri

stefonarch:
![schermata-02-02-21-32](https://user-images.githubusercontent.com/10681413/152232770-d141f233-166e-4187-839f-2bccd3fd4386.png

Cannot reproduce this neither in 1.0.0 neither in 0.17 (impish)

z0rti:
It just happens when i start a new qterminal, not with source
I tested it with 0.14.1 0.17.0 1.0.0 on impish,
0.14.1 on focal (can't compile master there due to g++ too old, and qt too)
What I do is start a qterminal, then start a new one by typing 'qterminal'
Now I noticed that when i use the menu to start a new window/tab, it does show all lines
Weird that it shows 12x2 first and then the right value with the echo command

stefonarch:
Ok, I get

```
 --- Greetings, stef
--- Your terminal size is: x
bash-5.1$ 
``` 
but only on the first tab, all the next ones are ok.

tsujan:
> Ok, I get... but only on the first tab, all the next ones are ok.

Then we have a bug, probably in `qtermwidget`.

z0rti:
I found out that it has to do with the font (Using [Liberation Mono](https://github.com/liberationfonts/liberation-fonts/releases)) size:
With   8pt i get `--- Your terminal size is: 16x2` the line is overwritten
With 10pt i get `--- Your terminal size is: 14x2` the line is overwritten
With 11pt i get `--- Your terminal size is: 12x2`
With 12pt i get `--- Your terminal size is: x` and nothing is overwritten
[...]

PR #1363, opened 2026-08-20
delay startShellProgram()

luebking:
Unfortunately not Q_INVOCABLE so singleShot(0) hack over re-implementing the function to only call its super

relates to #899
It fixes the missing ${COLUMNS}x${LINES} but I could not reproduce the "text overwritten" part, so idk whether that's fixed

indirectly relates to https://github.com/lxqt/qtermwidget/issues/480

For reference also https://bugs.kde.org/show_bug.cgi?id=516491

---------------

This isn't purely cosmetic since some clients require specific (minimum) sizes and will either falsely refuse to start or falsely start and misbehave if they check fast enough™ (this is inherently a race condition between the shell process and qterminal resizing the terminal display)
I've not test this w/ splitters which might introduce further delay for the splitter lazily resizing the terminal.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -82,6 +82,9 @@ MainWindow::MainWindow(TerminalConfig &cfg,
     setAttribute(Qt::WA_NoSystemBackground, false);
     setAttribute(Qt::WA_DeleteOnClose);
 
+    // see ::showEvent()
+    setProperty("terminal_size_pending", true);
+
     setupUi(this);
 
     // Allow insane small sizes - reason:
@@ -960,6 +963,10 @@ void MainWindow::showEvent(QShowEvent* event)
             twl.at(0)->setSize(m_initialSize.width(), m_initialSize.height());
         m_initialSize = QSize();
     }
+
+    // unconditionally here. QTabWidget lays out lazily and the (initial) terminal
+    // must know when this is done - ideally after the user specified size is set above
+    setProperty("terminal_size_pending", false);
 
     QMainWindow::showEvent(event);
 }
--- a/src/termwidget.cpp
+++ b/src/termwidget.cpp
@@ -47,6 +47,7 @@ static int TermWidgetCount = 0;
 
 TermWidgetImpl::TermWidgetImpl(TerminalConfig &cfg, QWidget * parent)
     : QTermWidget(0, parent)
+    , scheduledShellProgramStart(false)
 #ifdef HAVE_LIBCANBERRA
     , libcanberra_context(nullptr)
 #endif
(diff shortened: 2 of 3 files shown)


luebking:
This is still not enough?
I ran into a 0/0 size with a script, we'll have to figure when that state is actually fixed and how we can get notified of that instead of kicking race conditions down the road…

luebking:
It's probably the lazy ~~splitter~~ tabwidget again… this is for `qterminal -s 128x32`
```
QTermWidget(0x55ea0e5b51c0) resize terminal display
QTermWidget(0x55ea0e5b51c0) … done
QTermWidget(0x55ea0e5b51c0) add view
added view Konsole::TerminalDisplay(0x55ea0e5a89a0)
QTermWidget(0x55ea0e5b51c0) … done
Konsole::TerminalDisplay(0x55ea0e5a89a0) update image Konsole::ScreenWindow(0x55ea0e648740)
updateImageSize 1 1 9 1
updateImageSize 9 1 9 1
Konsole::TerminalDisplay(0x55ea0e5a89a0) update image Konsole::ScreenWindow(0x55ea0e648740)
session: RUN! # this is already the scheduled start
updateImageSize 9 1 1 7
updateImageSize 1 7 128 19 # default window size
updateImageSize 128 19 128 32 # target size
Konsole::TerminalDisplay(0x55ea0e5a89a0) update image Konsole::ScreenWindow(0x55ea0e648740)
```

src/mainwindow.cpp
```diff
@@ -82,6 +82,9 @@ MainWindow::MainWindow(TerminalConfig &cfg,
     setAttribute(Qt::WA_NoSystemBackground, false);
     setAttribute(Qt::WA_DeleteOnClose);
 
+    // see ::showEvent()
+    setProperty("terminal_size_pending", true);
```
luebking, line 86: I don't really like relying on this property, but the alternative is to change TermWidgetImpl to no longer auto-start the shell but completely rely on MainWindow doing that - what I like even less.

We can also not rely on the visibility of the tabwidget because then shells would not be started in minimized windows and we cannot rely on visibleTo because that's not relevant to QTabWidget getting its ass up…

Meh.

nb. the logics are set so that by default (no property) the shell will always be started early.

luebking:
This is gonna overlap and merge conflict w/ https://github.com/lxqt/qterminal/issues/1374
=> Any comments?

tsujan:
Conflicts aren't unusual. If a PR is merged, another might need to be rebased.

luebking:
:)

What I meant to say is "I'd like to avoid having to rebase resolving merge conflicts this back and forth, so if you can project which one you'd like to look at first that would probably be helpful."

tsujan:
> if you can project which one you'd like to look at first

https://github.com/lxqt/qterminal/issues/1374 doesn't have a PR :) I didn't know either if this was finished, because you showed doubt in your comments.

Apart from that, sometimes I play the role of a simple maintainer here and need the help of other contributors to know if a PR should be merged. For example, I need your help with https://github.com/lxqt/qtermwidget/pull/648, although it has conflicts too.

luebking:
There's a simple PR in qtermwidget :P
However that's the point - I'm wondering whether to hold back the PR or shift it before this patch.

I meant to point out that I'm aware there may be objections to using the object property, but all alternatives kinda suck, too.
(One would be to make that a function on the MainWindow class but that requires TermWidgetImpl to include that header and test the window() object) - I've no major concerns reg. the taken approach but being hack-adjacent you might other preferences.

tsujan:
Understood. Before your comment about conflicts, I was going to ask you if this was ready. Let me do some tests first. I'll merge it soon if everything goes well — perhaps tomorrow.

And thanks a lot for your comments at https://github.com/lxqt/qtermwidget/pull/648. I'll need your expertise more than you may think.

tsujan:
> perhaps tomorrow.

My bad; a glance at the code shows that it doesn't need to wait. Merging …

PR #1369, opened 2026-08-30
Update CONTRIBUTING.md to clarify code style

lq259:
Currently, QTerminal's codebase has multiple intermixed code styles.
If developers do not know what the intended code style is, this problem **will become worse over time**. This PR clarifies QTerminal's intended code style, which is Allman.


tsujan (approved):
A short good change.

PR #1370, opened 2026-09-04
Pick ups

luebking:
1. unminimize windows before trying to activate them (at least fluxbox doesn't allow to activate minimized windows)
2. shell clients/scripts might be interested in activating existing tabs (ie. themselves)
3. expose the window state via dbus to allow clients to query whether this is the active window w/o relying on xdotool or proprietary wayland compositor channels (swaymsg/niri msg/…) and be better at it
4. copypasta clean-up (functionally irrelevant but qdbus and the auto-completers show those and the false labels are irritating)
5. make the exit dialog (warning about a running process) window modal to not block all other windows in the same process

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1129,6 +1129,8 @@ void MainWindow::activateOrHide()
     }
     else
     {
+        if (QWindow *win = windowHandle())
+            win->setWindowStates(win->windowStates() & ~Qt::WindowMinimized);
         show();
         activateWindow();
         raise();
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -632,6 +632,7 @@ bool MainWindow::closePrompt(const QString &title, const QString &text)
     QDialog * dia = new QDialog(this);
     dia->setObjectName(QStringLiteral("exitDialog"));
     dia->setWindowTitle(title);
+    dia->setWindowModality(Qt::WindowModal);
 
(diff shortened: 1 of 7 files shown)


luebking:
Incidentally there seems a recursive connection between termGetFocus and term_termGetFocus ?
(Since 2010)

tsujan (approved):
GTM.

Sooner or later you'll migrate to Wayland and forget about window activating/unminimizing  :P

isf63:
I think this change fixed the dropdown showing a taskbar button for kwin_wayland, but I may be mis-remembering.

I vaguely remember we added a workaround for X11 / Xfwm4 to fix the taskbar button, and then reverted it because it broke kwin_x11. Perhaps the minimized states was the silver bullet.

tsujan:
> I think this change fixed the dropdown showing a taskbar button for kwin_wayland

No task-bar button should be shown for any overlay window on Wayland. As far as I remember, lxqt-panel never showed it for the drop-down mode — I used kwin_wayland for a long time.

luebking:
Should be easy enough to test by reverting the patch though the only way I see running into this would be if QMainWindow::restoreState() had restored a minimized state on the dropdown for the config (what might now be gone for good) and then some activateOrHide call (which afaict isn't used for the dropdown window) would have withdrawn that.

nb. that this would not be a property of the WM/compositor but the taskbar (though I suspect on wayland its only way to get to the window would be through the compositor exposing it) - at least on X11 a taskbar *could* pick up an override_redirect window and interpret it despite that state (preferring the NETWM one)

Overall I however don't think it's very likely that the particular patch did anything here.

PR #1372, opened 2026-09-08
Enabled QTermWidget's new background image modes

tsujan:
→ https://github.com/lxqt/qtermwidget/commit/6f527680653e6b95aca7f5eeb00bc490a7780c29


src/propertiesdialog.cpp
```diff
@@ -359,7 +359,7 @@ void PropertiesDialog::apply()
     Properties::Instance()->focusOnMoueOver = focusOnMoueOverCheckBox->isChecked();
     Properties::Instance()->showTerminalSizeHint = showTerminalSizeHintCheckBox->isChecked();
     Properties::Instance()->backgroundImage = backgroundImageLineEdit->text();
-    Properties::Instance()->backgroundMode = qBound(0, backgroundModecomboBox->currentIndex(), 5);
+    Properties::Instance()->backgroundMode = qBound(0, backgroundModecomboBox->currentIndex(), 10);
```
luebking, line 362: For regular use (bash) topright is probably a better spot (because the prompt will quickly end up on the bottom line on standard setups) though that's also personal preference (just saying because I mentioned that corner but that'S for a different kind of setup)

tsujan: It's up to the user to choose a bg image that doesn't interfere with reading — if you mean that.

luebking: No what I meant is that I pointed out that the [bottom right corner is a nifty space for logos](https://github.com/lxqt/qtermwidget/pull/658#issue-5269509436) and you changed the default there, but my comment was w/ a setup in mind that's probably different from the standard terminal use where the top right corner is gonna be the most irrelevant/silent one (readability issues aside, just to see the logo)

tsujan: > you changed the default there

I didn't. The default is still "None" — that's just `qBound()`. You can check it with something like `qterminal -p zzz`.

luebking: 'key  -  I clearly need to take more time for this ;)

There's an unrelated change to a header include that probably should be an isolated commit?
And, in my shallow defense, I'm not quite sure why you're applying bounds from a GUI element under your control anyway - but it's also not gonna hurt.

Other than that there's not much to comment if @stefonarch is ok with the strings.

tsujan: > There's an unrelated change to a header include

Yes. `<qkeysequence.h>` should have been added recently. It hurt my eyes and I changed it here silently. Not a good practice ;)

> I'm not quite sure why you're applying bounds

Here I just kept them. But yes, I like to add them to make each class bulletproof, although in this case they may be an overkill.

stefonarch (approved):
Works fine. I just noticed that adding a logo 48px it get's partially covered by the scrollbar:

Image

tsujan:
> it get's partially covered by the scrollbar:

Yes, I saw that too but thought it was expected with Kvantum's *transient* scroll-bars. However, your screenshot is taken with Fusion.

Anyway. the fix should be made in qtermwidget, and `QStyle::SH_ScrollBar_Transient` should be considered (as it is on some other occasions). This PR is only an elementary change.

tsujan:
With fusion and most other styles (problematic):

fusion

With Kvantum + transient scroll-bar (OK):

kvantum

PR #1379, opened 2026-09-14
Non-modal file dialog with dropdown mode on Wayland

### LINKED ISSUE #1378: Bad situation with file dialog and dropdown mode on Wayland
If the file dialog is opened from the properties dialog of the dropdown mode on Wayland (two cases), it might go behind the dropdown window(s) completely. Then, with some compositors, the user might end up in a situation where he couldn't interact with QTerminal at all.

The only solution I've found is open the file dialog non-modal in this case, letting the user close the properties dialog and do the job in the normal mode later. This isn't ideal but is much better than getting trapped in the above situation.

Will make a PR soon.

tsujan:
To avoid getting trapped in a bad situation, where the file dialog is completely behind the drop-down window and the user can't do anything. The goal is that the user should be able to cancel the properties dialog and do the job in the normal mode later.

Closes https://github.com/lxqt/qterminal/issues/1378.


luebking:
```dlg->setWindowFlag(Qt::Dialog, false)``` but oc that would limit the dialogs to the terminal size (so you might temporarily adjust that)

Though a modal dialog (forced) behind its (ultimate) leader is a clear-cut WM bug that *needs* to be fixed there anyway.
Did you try to file a bug against wayland™ (ie. labwc or niri or which ever has that bug)?

(It's not all that uncommon, I'm pretty sure to once have fixed that in kwin and fullscreen windows almost two decades ago :))

**Comment by stefonarch** [comment]
It's an issue "by design" as the dropdown terminal is not a window but a layer (overlay, which is rendered above everything). What would be nice is just closing the dropdown terminal when the settings window and filedialog are open.

**Comment by luebking** [comment]
So are fullscreen windows (where the kwin bug was) - doesn't matter.
Modals block their leader, they *have* to be on top of their leaders because of that, inheriting the ancestors layer.

Incidentally, minimizing/hiding a client would usually™ implicitly minimize/hide their modals (I think anothe kwin bug itr was that it hid all transient, modal or not) so that approach would have to be tested across environments at all.

The thing that technically does not need to be modal (I think) would be the preferences dialog (eg. if you've multiple terminal windows the preferences would affect all(?) but only modal for one, so it's partially non-modal anyway)

**Comment by tsujan** [comment]
> Did you try to file a bug against wayland

In this case, QTerminal and some of its dialogs are on the "overlay" layer, while the file dialog cannot be put there because it's native.

As for bugs, they should be reported to Wayland *compositors*. I think we reported this case to some. Making codes work with most Wayland compositors has been a nightmare. The original creators of Wayland were too nearsighted — I'm saying this as someone who uses only Wayland and can't tolerate X11 anymore ;)

> What would be nice is just closing the dropdown terminal

@luebking replied well to this.

Anyway, here the goal is avoiding a nasty situation, as mentioned in the PR's comment, as well in the original report: the ability to (focus and) close the properties dialog and get out of the trap on all Wayland compositors.

**Comment by luebking** [comment]
> cannot be put there because it's native.
??
So this depends on some xdg-portal or is it the built-in Qt file dialog that's affected here?
(You can tell Qt to prefer that over system-native dialog)

Either way and though just tested, the preferences are truly application modal, I'd rather look into making the preferences non-modal.
While it might not matter for some file dialogs, the approach probably doesn't scale - usually file dialogs™ are blocking modals because you want to do sth. w/ the result next and being able to interact w/ the underlying widget may run you into issues down the road.
(didn't check the patch but you could then create a de-facto modal that disables the preferences dialog until the file dialog returned)

If you care and tell me which compositor I need to try I can also check what's going on this evening.

**Comment by tsujan** [comment]
> I'd rather look into making the preferences non-modal.

No, that and other possibilities were already considered and proved to be bad. Moreover, we have other dialogs that should be put over it, and all of them are as overlays (see `PropertiesDialog::event`). The file dialog is an exception.

> usually file dialogs™ are blocking modals because you want to do sth

That's already taken into account by the patch: only one file dialog can be opened.

> If you care and tell me which compositor

(Almost?) all ;)

**Comment by tsujan** [comment]
@luebking 
You test Wayland under X11. Believe me or not, you can't see all details in that way. You need separate sessions with at least 3 totally different Wayland compositors to encounter details. By "totally different" I mean different libraries, like wlroots, smithay and kwin's.

**Review by stefonarch — APPROVED** [review]
not perfect  but better as nothing

**Comment by tsujan** [comment]
> not perfect but better as nothing

As I told in the report ;)

**Comment by luebking** [comment]
> You test Wayland under X11. 
No, I don't.
As for this specifically I hadn't tested at all but now have and there's no problem w/ either niri nor labwc

So I continue to suspect that this is because of the xdg-desktop-portal dialog and https://doc.qt.io/qt-6/qfiledialog.html#Option-enum QFileDialog::DontUseNativeDialog would likely cause proper behavior (at the expense of the native dialog) ?

(Incidentally this also means that there's no bug in any WM, but apparently either another nearsightedness in xdg-desktop-portal or a bug in the specific implementation)

**Comment by tsujan** [comment]
> So I continue to suspect that this is because of the xdg-desktop-portal dialog 

Only partially. In a Qt-based environment, it's about `platformthemes`.

> Incidentally this also means that there's no bug in any WM

Not WMs, but the original Wayland.

**Comment by luebking** [comment]
> Only partially. In a Qt-based environment, it's about platformthemes.
If it's from a plugin (ie. same process) that should™ be a bug in the plugin then as apparently the internal dialog operates correctly?
(Have you tried? I don't typically use the quake mode so if the problem hinges on a specific config or behavior I might easily have missed that)

> Not WMs, but the original Wayland.
"Compositors" - the original wayland would be weston, not sure whether that runs into wlr-layer-shell related issues.
https://wayland.app/protocols/wlr-layer-shell-unstable-v1#compositor-support

**Comment by tsujan** [comment]
> that should™ be a bug in the plugin

"should™"? You shouldn't be so sure ;) No, there's no bug.

> "Compositors" - the original wayland would be weston

The original creators of Wayland sent us back to the stone age. Years later, that was compensated little by little.

**Comment by luebking** [comment]
Does the default dialog (eg qt6ct QPA) show the same misbehavior for you?

**Comment by tsujan** [comment]
> Does the default dialog (eg qt6ct QPA) show the same misbehavior for you?

I investigated it when implementing the Wayland support. It has a long story; this is a short version:

 * Qt's default file dialog was put on the overlay layer correctly — I remember that I checked it by setting `QFileDialog::DontUseNativeDialog` in `PropertiesDialog`.
 * Also, LXQt's file dialog was set on the same layer in `PropertiesDialog::event` — `LayerShellQt::Window::layer()` returned `LayerOverlay` — *but only programmatically*, not actually! I suspected some problem in `layer-shell-qt` (and I still do) and tried to find out why it didn't work in this case.
 * At last I found that it was somehow related to the delayed showing of LXQt file dialog: https://github.com/lxqt/libfm-qt/blob/21582910892cb390375b9d6ef713066ac91ae3ad/src/filedialoghelper.cpp#L73-L80. That delay was (and is) necessary. Without it, the dialog would be blocked.
 * I vaguely remember that I thought of some changes to `FileDialogHelper` such that the problem of `layer-shell-qt` would be circumvented, but I really didn't want to run the risk of a blocked UI just because of this special case. Showing Qt's default file dialog didn't seem like a good idea either.
 * Finally, I used some swear words and left it as it was. This was only a small example of my Wayland nightmare.

A long time has passed since then, and I'm ready to take back those swear words. So I might attack the problem again later.

**Comment by tsujan** [comment]
Now I remember clearly:

Actually, the layer was set for the internal `QFileDialog`, not for `Fm::FileDialog`, because it was the former that blocked `PropertiesDialog` first. That was the reason why `QFileDialog::DontUseNativeDialog` worked.

A single-shot timer inside `PropertiesDialog::event` would set the layer for `Fm::FileDialog`, but then it would be too late because the layer should be set for a window **before** its `showEvent` returns: *this is the problem I talked about.* I was able to deal with it everywhere except for here.

**Comment by luebking** [comment]
Have you tried to filter the QEvent::show of the dialog(s) and set the layer there before allowing the event to pass?
https://doc.qt.io/qt-6/qchildevent.html should™ allow you to intercept "unpredicted" child objects (your QPA dialog)

**Comment by tsujan** [comment]
Using `showEvent` is the solution I've used everywhere. It can't be done here — whether directly or by using `QEvent::show` — because of the situation I explained above.

Anyway, I took a nap and saw a solution in a dream. It was about circumventing Qt's old problem of showing a native dialog in a way other than delaying its showing in libfm-qt. I might disappear for a while to fully bring it to the wake state …

**Comment by tsujan** [comment]
OK, dreams never lie, but it may take a while to extract the simple truth from them:

fm_file_dialog

The above screenshot is taken *without* the current patch.

What I did was as simple as adding the following two lines below https://github.com/lxqt/libfm-qt/blob/21582910892cb390375b9d6ef713066ac91ae3ad/src/filedialoghelper.cpp#L73-L80:

```c++
    dlg_->show();
    dlg_->hide();
```

However, I still don't want to take the risk. Our file dialog works fine, and I'm not sure that a seemingly innocent change like this is safe.

stefonarch:
I'll take the risk and run that for a while ;)

immagine

```
  Top layer:
    Surface:
      Namespace: "dialog"
      Keyboard interactivity: on-demand

    Surface:
      Namespace: "dropdown_terminal"
      Keyboard interactivity: on-demand


  Overlay layer:
    Surface:
      Namespace: "dialog"
      Keyboard interactivity: on-demand

**Comment by tsujan** [comment]
> I'll take the risk and run that for a while ;)

I'm doing it too :)  When I said I didn't want to take the risk, I meant adding those two lines to git.

Don't forget to revert this patch for testing.

**Comment by stefonarch** [comment]
Only thing (as always) is that focus doesn't turn to the underlaying surface. Same with the confirm dialog to  exit or not.

**Comment by tsujan** [comment]
Yes, but it's not related to this. I remember that it didn't happen in KWin.

**Comment by tsujan** [comment]
I'm making a libfm-qt PR for more tests. In short, the delayed showing is necessary, but the dialog is shown and hidden for having access to it immediately when needed — like when its layer is set without this patch.

EDIT: → https://github.com/lxqt/libfm-qt/pull/1118

**Comment by luebking** [comment]
So we're now flickering things into existence … I think I've seen that approach before ;)

This is all nuts.
Ignoring the idea that the QPlatformDialog design is so ridiculous that you cannot use it because it'll jam another modal dialog in your way…

If the problem is a race where you need to set the layer early but can't because there's no handle until it's too late - force one into existence: https://github.com/luebking/qiq/blob/main/qiq.cpp#L494
You still won't be able to use the static function but should™ be able to set the layer well before any of that dance-show or show-dance is happening?

(If the two of you are trying to trick me into installing a desktop environment, I call dips - because **I knew it** :P)

**Comment by tsujan** [comment]
> but should™ be able to set the layer well before any of that dance-show or show-dance is happening?

The layer should be set when the widget isn't shown. It can also be set in `showEvent` before returning.

> This is all nuts.

Of course it is ;) Wayland is full of such things. I've got used to them. Try to cover multi-screen setups and you'll have to deal with much more issues, especially with Qt but not only with it. Fortunately, there have always been good solutions — or almost always.

My plan: test the libfm-qt PR to make sure it doesn't have any side effect. If it's OK, merge it and revert this PR :) Then we'll have no problem with LXQt file dialog. Those who may have problem with their native dialogs should find their own solutions :P

**Comment by luebking** [comment]
The flicker in libfm-qt is only required to apply the layershell layer and for all other cases an artifact?
Did you ensure that calling winId() isn't the sufficient part of the patch for an early handle creation and layer success?
(Because I'm not quite sure what the idempotent show would do otherwise)

**Comment by tsujan** [comment]
> The flicker in libfm-qt is only required to apply the layershell layer and for all other cases an artifact?

For now, yes. But who know? It may be useful in some other cases too. That's why I didn't mention Wayland in the comments. We should be very cautious though.

> Did you ensure that calling winId() isn't the sufficient part of the patch for an early handle creation and layer success?

Yes. I tested several other things based on various hypotheses too, until the obvious solution came to my mind. I used `winId()` only because it's more secure than setting `Qt::WA_NativeWindow`.

**Comment by luebking** [comment]
I've meantime allowed myself into being tricked into installing a DE and looked at the situation a bit.
The main issue is that the layer-switch is attempted w/ QEvent::WindowBlocked (interestingly, though expectably, only one dialog comes along, checked ChildAdded) which is apparently too late (and the flicker does at this point actually not help either)

From the Qt code (and the events coming in) the helper/native dialog is however a hidden detail that you don't get to operate on (you'adjusting the layer of the mock QDialog) - the viable path would be to get the delayed show() under control.

https://codebrowser.dev/qt6/qtbase/src/widgets/dialogs/qdialog.cpp.html#766
The code looks unreachable but do you set DialogIsQtWindow anywhere?

**Comment by tsujan** [comment]
> do you set DialogIsQtWindow anywhere?

No, we don't.

This isn't limited to dialogs. There are other cases that hiding and showing of a layered window are required. For example, an already visible layered window is not shown on a new screen under Wayland; it should be made hidden and then shown by using `QTimer::singleShot` (as with the multi-screen support of lxqt-panel on Wayland).

I vaguely remember that some GTK-using devs encountered similar problems. They may not be specific to Qt. Add to it the problems of this or that Wayland compositor, which make the process of narrowing down harder …

EDIT: Personally, and based on experience, I rely on the sanity of Labwc, Niri, and to some extent, KWin. Kwin hides some Wayland issues in Qt apps. In contrast, Wayfire has bad issues regarding Qt apps; etc.

**Comment by luebking** [comment]
> No, we don't.

Despite the code superficially looking unreachable in codebrowsers checkout, I'd try - as it seems directed at the exact problem you're working around w/ the delay and then the flicker. Always remember the epicycles…

> There are other cases that hiding and showing of a layered window are required.

No comment :P

> by using QTimer::singleShot

"QMetaObject::invokeMethod(. , Qt::QueuedConnection)"?

> similar problems. […] the problems of this or that Wayland compositor, which make the process […] harder

Are you trying to convince me that wayland is a lousy protocol?

**Comment by tsujan** [comment]
> Are you trying to convince me that wayland is a lousy protocol?

Wayland itself, no. But Qt's handling of it is still poor.

**Comment by tsujan** [comment]
> Despite the code superficially looking unreachable in codebrowsers checkout …

I have offline bookmarks of the latest stable Qt source in FeatherPad, search keywords in it with pcmanfm-qt, and have used `Qt::WA_DontShowOnScreen` where needed. IMO, `QTimer::singleShot` circumvents this serious issue easily. It was the choice of the original author of libfm-qt; I kept it there because of its efficiency.

The question is: how many DEs have native Qt dialogs? I know of only LXQt and KDE. Qt's default file dialog is used in other places. This may be the reason that such Qt issues persist for a very long time: we circumvent them, and Qt devs forget them or even don't notice them.

A similar situation exists in the case of `QStyle`. Most style plugins copy Fusion and change it in some places. Only those that use the full potential of `QStyle` encounter its problems. Reporting them to Qt has been futile to me most of the time; hence making workarounds after workarounds.

**Comment by luebking** [comment]
I actually came to suspect that Qt::WA_DontShowOnScreen will be used by the built-in dialog on a second show run (for itself), thus reaching the code path trying to avoid the internal block.

> QTimer::singleShot circumvents this serious issue easily

And then apparently causes a slew of subsequent issues :\

**Comment by tsujan** [comment]
> And then apparently causes a slew of subsequent issues :\

No issue whatsoever in the long time we've used it. The situation in QTerminal's Wayland support was exceptional. I don't know how GTK-using devs deal with similar situations, or if it's GTK itself that handles them.

**Comment by luebking** [comment]
```
diff --git a/src/filedialoghelper.cpp b/src/filedialoghelper.cpp
index 4540ca0..8405a09 100644
--- a/src/filedialoghelper.cpp
+++ b/src/filedialoghelper.cpp
@@ -77,7 +77,8 @@ bool FileDialogHelper::show(Qt::WindowFlags windowFlags, Qt::WindowModality wind
     // This includes our own dialog. After the return of this show() method, QFileDialog creates its own window and
     // then make it modal, which blocks our UI. The timer schedule a delayed popup of our file dialog, so we can
     // show again after QFileDialog and override the modal state. Then our UI can be unblocked.
-    QTimer::singleShot(0, dlg_.get(), &QDialog::show);
+//    QTimer::singleShot(0, dlg_.get(), &QDialog::show);
+    dlg_->show();
     dlg_->setFocus();
     return true;
 }
diff --git a/src/filedialoghelper.h b/src/filedialoghelper.h
index d238240..58c1cd6 100644
--- a/src/filedialoghelper.h
+++ b/src/filedialoghelper.h
@@ -36,6 +36,10 @@ public:
 
     bool isSupportedUrl(const QUrl &url) const override;
 
+    QVariant styleHint(StyleHint hint) const override {
+        return hint == DialogIsQtWindow ? true : QPlatformFileDialogHelper::styleHint(hint);
+    }
+
 protected:
     inline Fm::FileDialog & dialog() { return *dlg_; }
 ```

On a quick test, the dialog isn't blocked.
~~Sadly doesn't seem to help w/ the overlay window situation, though.~~

~~But you might want to consider moving the layer control into the plugin because there might be other instances where the lxqt QPA is used w/ a special window on a  defective display server - which will break there while the native dialog works.~~

EDIT, CORRECTION: this fixes your problem.
(I had forgotten that I was running a qterminal build where I had experimented w/ the behavior and deliberately broken it)

There's nothing wrong w/ Qt, you copied a lousy hack from the clowns that are running KDE these days instead of using the proper API.

tsujan:
Thanks a lot! It should work because KDE does exactly the same thing now; I checked their new codes after seeing your PR. Will test it soon.

> you copied a lousy hack from the clowns that are running KDE these days

Not these days, but in old days ;) That comment about KDE is very old.

That being said, there are still lots of lousy codes in KDE (or QMLDE?) and their number grows constantly, especially when it comes to `QStyle`. So I share your annoyance ;)

tsujan:
Gathered some info:

`DialogIsQtWindow` was added to Qt 5.13, only 2 years after LXQt made its file dialog (2017). KDE started to use it instead of its workaround in 2020 (https://invent.kde.org/plasma/plasma-integration/-/merge_requests/1).

PR #1364, opened 2026-08-22
Improve consistency of Allman code style

lq259:
Tried to not fix the entire project's code style on this one, but I may have gone a little bit overboard.

Good luck reviewing :) Will certainly understand if this pull request is closed.

PR #1381, opened 2026-09-19
Apply code style consensus: src/main.cpp

isf63:
Inspired by "Allman style" discussion [here](https://github.com/lxqt/qterminal/pull/1364)

I think this is a better approach ⟶ changes should be limited to at most 2 or 3 files (depending on the diff), and be as thorough as possible.


isf63:
Also, only files with a large amount of possible cleanup should be targeted, so to not confuse reviewers.

lq259:
👍

